PR #2, opened 2026-02-23
Support file path as an input

dmitriyk-code:


jrob5756 (commented):
## Review: Support file path as an input

Thanks for the contribution — this is a useful feature. The `@filepath` syntax is a clean, established convention (used by `curl`, `gh`, etc.) and the implementation is well-structured.

A few issues to address before this is ready to merge. See inline comments for details, and a summary below:

| Priority | Issue | Action |
|----------|-------|--------|
| 🔴 | 2MB fixture file committed to git | Generate dynamically in test via `tmp_path` |
| 🔴 | `latin-1` in encoding fallback makes `cp1252` unreachable | Fix encoding fallback order |
| 🟡 | No file size limit | Add a reasonable cap to prevent OOM |
| 🟡 | Bare `@` with no path gives confusing error | Add explicit check |
| 🟡 | Test relies on CWD being project root | Use `tmp_path` or `monkeypatch.chdir()` |
| ℹ️ | Fixture files missing trailing newline | Minor nit |
| ℹ️ | CLA not yet signed | Must be resolved before merge |

PR #90, opened 2026-04-11
fix: use UTF-8 encoding for subprocess output on Windows

### LINKED ISSUE #89: Windows: Unicode characters in agent output are garbled (charmap encoding)
## Bug
 
 On Windows, when agents produce output containing emoji or non-ASCII characters 
 (—, ✅, ⚠️, 🔍, 🟡, etc.), the characters are mangled when passed to downstream 
 agents or when posted via tools (e.g., gh api). 
 
 Example: `✅` becomes `Γ£à`, `—` becomes `ΓÇö`, `🟡` becomes `≡ƒƒí`
 
 ## Reproduction
 
 1. Create a workflow where agent A produces emoji in its output
 2. Agent B receives agent A's output and posts it via `gh pr comment`
 3. The posted comment contains garbled characters
 
 ## Environment
 - Windows 11, PowerShell 7
 - Conductor v0.1.1
 - Provider: copilot
 - `$env:PYTHONIOENCODING = "utf-8"` is set but doesn't fix inter-agent output passing
 
 ## Root Cause (suspected)
 
 The agent output capture pipeline uses Python's default console encoding 
 (`charmap`/`cp1252` on Windows) instead of UTF-8 when collecting agent responses 
 and passing them through the workflow context. The `PYTHONIOENCODING=utf-8` env var 
 fixes Python's own stdout but doesn't affect how the Copilot SDK or subprocess 
 pipes handle encoding internally.
 
 ## Expected Behavior
 
 Agent outputs should preserve Unicode characters end-to-end on all platforms.

taipoweredpm:
## Summary

Fixes #89 — Unicode characters (emoji, em-dash, etc.) are garbled on Windows when passed between agents or posted via tools.

## Root Cause

Python defaults to cp1252 (charmap) encoding on Windows for subprocess pipes and text=True mode. This mangles non-ASCII characters in script output and subprocess calls.

## Changes

- **executor/script.py**: Set PYTHONUTF8=1 in subprocess env so child Python processes use UTF-8
- **cli/update.py**: Added encoding="utf-8" to subprocess.run()
- **mcp_auth.py**: Added encoding="utf-8" to subprocess.run()

## Testing

All 75 existing tests pass. Lint clean.

### DIFF
--- a/src/conductor/executor/script.py
+++ b/src/conductor/executor/script.py
@@ -92,7 +92,11 @@ class ScriptExecutor:
         # Build environment (merge os.environ + agent.env)
         # Note: ${VAR:-default} patterns in agent.env are already resolved
         # by the config loader during YAML parsing.
-        env = {**os.environ, **agent.env} if agent.env else None
+        # Always set PYTHONUTF8=1 so child Python processes use UTF-8 encoding
+        # instead of the system default (cp1252 on Windows), preventing garbled
+        # Unicode characters in script output.
+        base_env = {**os.environ, "PYTHONUTF8": "1"}
+        env = {**base_env, **agent.env} if agent.env else base_env
 
         _verbose_log(f"  Script: {rendered_command} {' '.join(rendered_args)}")
 
--- a/src/conductor/cli/update.py
+++ b/src/conductor/cli/update.py
@@ -17,6 +17,7 @@ import contextlib
 import hashlib
 import json
 import logging
+import os
 import shutil
 import subprocess
 import sys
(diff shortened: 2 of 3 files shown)


taipoweredpm:
@microsoft-github-policy-service agree company="Microsoft"

src/conductor/executor/script.py
```diff
@@ -92,7 +92,11 @@ async def execute(
         # Build environment (merge os.environ + agent.env)
         # Note: ${VAR:-default} patterns in agent.env are already resolved
         # by the config loader during YAML parsing.
-        env = {**os.environ, **agent.env} if agent.env else None
+        # Always set PYTHONUTF8=1 so child Python processes use UTF-8 encoding
+        # instead of the system default (cp1252 on Windows), preventing garbled
+        # Unicode characters in script output.
+        base_env = {**os.environ, "PYTHONUTF8": "1"}
+        env = {**base_env, **agent.env} if agent.env else base_env
```
jrob5756, line 99: `script.py` fixes the child side (env var), while `update.py` and `mcp_auth.py only` fix the parent side (encoding= param). For Python-based child CLIs, both sides should be addressed.

jrob5756 (commented):
One small comment.

src/conductor/executor/script.py
taipoweredpm: Good catch - addressed in e17ac2f. Both `update.py` and `mcp_auth.py` now set `PYTHONUTF8=1` in the subprocess env alongside the existing `encoding="utf-8"` param, matching the approach in `script.py`. All three call sites are now consistent with both parent-side and child-side encoding fixes.

jrob5756 (approved):
lgtm. Thanks!

PR #94, opened 2026-04-16
fix: add random suffix to log filenames to prevent collision

PolyphonyRequiem:
## Problem

When multiple workflow runs start in the same second (common when orchestrating parallel runs), \	ime.strftime('%Y%m%d-%H%M%S')\ produces identical timestamps, causing all runs to write to the same event log file. This corrupts event logs, checkpoint files, and CLI log files by interleaving events from different runs.

## Solution

Append a random 8-character hex suffix (via \secrets.token_hex(4)\) to filenames across all three affected locations:

- \EventLogSubscriber\ (\vent_log.py\)
- \CheckpointManager.save_checkpoint\ (\checkpoint.py\)
- \generate_log_path\ (\cli/run.py\)

### Before
\\\
conductor-workflow-20260416-014816.events.jsonl
\\\

### After
\\\
conductor-workflow-20260416-014816-a3b7c9f1.events.jsonl
\\\

## Backward Compatibility

Fully backward compatible — existing tools that glob \*.events.jsonl\, \*.json\, or \*.log\ continue to work. Filenames remain human-readable with the timestamp prefix; the suffix only prevents collisions.

## Testing

- Added \	est_filenames_unique_for_simultaneous_starts\ — creates 3 subscribers in rapid succession, asserts all paths are unique
- Added \	est_filename_contains_random_suffix\ — validates the 8-char hex suffix format in filenames

## Note

One pre-existing test (\	est_handles_non_serializable_data\) fails on Windows due to a path separator assertion (\\\\\some\\\\path\ vs \/some/path\). This is unrelated to this change.

### DIFF
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -52,7 +52,13 @@ def generate_log_path(workflow_name: str) -> Path:
     Returns:
         Path to the auto-generated log file.
     """
+    import secrets
+
     timestamp = time.strftime("%Y%m%d-%H%M%S")
+    # Append random suffix to avoid filename collisions
+    # when multiple runs start in the same second
+    suffix = secrets.token_hex(4)
+    timestamp = f"{timestamp}-{suffix}"
     path = Path(tempfile.gettempdir()) / "conductor" / f"conductor-{workflow_name}-{timestamp}.log"
     path.parent.mkdir(parents=True, exist_ok=True)
     return path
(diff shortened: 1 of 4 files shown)


PolyphonyRequiem:
@microsoft-github-policy-service agree company="Microsoft"

jrob5756 (approved):
lgtm. Thanks!

PR #100, opened 2026-04-20
fix(copilot): infer nested prompt schema from output definitions

lesandiz:
## Summary
- build the Copilot prompt schema recursively from nested `output:` definitions instead of flattening to top-level fields only
- include nested object properties, required keys, and array item schemas in the prompt-facing schema used for initial guidance and parse recovery
- add regression coverage for recursive schema generation and the actual prompt sent to Copilot

## Validation
- uv run --python 3.13 pytest tests\\test_providers\\test_copilot.py -q
- uv run --python 3.13 conductor run C:\Users\lester\.copilot\session-state\9a4adc0c-6650-46e6-9540-061d5b29a7ef\files\conductor-nested-output-repro.yaml --input topic="AI Harness Engineering"

## Context
- no exact open issue or open PR was found for this bug
- related prior work: #48 flattened an example workflow schema, but did not fix provider-side schema inference

### DIFF
--- a/src/conductor/providers/copilot.py
+++ b/src/conductor/providers/copilot.py
@@ -21,7 +21,7 @@ from conductor.exceptions import ProviderError
 from conductor.providers.base import AgentOutput, AgentProvider, EventCallback
 
 if TYPE_CHECKING:
-    from conductor.config.schema import AgentDef
+    from conductor.config.schema import AgentDef, OutputField
 
 logger = logging.getLogger(__name__)
 
@@ -505,13 +505,7 @@ class CopilotProvider(AgentProvider):
         # Build schema description for output schema (used in prompt and recovery)
         schema_for_prompt: dict[str, Any] | None = None
         if agent.output:
-            schema_for_prompt = {
-                name: {
-                    "type": field.type,
-                    "description": field.description or f"The {name} field",
-                }
-                for name, field in agent.output.items()
-            }
+            schema_for_prompt = self._build_prompt_schema(agent.output)
             schema_desc = json.dumps(schema_for_prompt, indent=2)
             full_prompt += (
                 f"\n\n**IMPORTANT: You MUST respond with a JSON object matching this schema:**\n"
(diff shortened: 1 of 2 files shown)


lesandiz:

src/conductor/providers/copilot.py
```diff
@@ -1082,6 +1076,51 @@ def _build_parse_recovery_prompt(
             f"than the raw JSON object."
         )
 
+    def _build_prompt_schema(self, schema: dict[str, OutputField]) -> dict[str, Any]:
```
jrob5756: **Missing recursion depth limit** — These three new recursive methods (`_build_prompt_schema` → `_build_prompt_field_schema` → `_build_prompt_schema` / `_build_prompt_item_schema` → `_build_prompt_item_schema`) have no depth guard. A pathologically deep schema would cause a `RecursionError` crash.

The Claude provider already protects against this with `_max_schema_depth = 10` and checks at the top of its corresponding methods (`_build_json_schema_properties` and `_build_single_field_schema`).

Suggested fix: add a `depth` parameter (default 0) to all three methods and raise `ValidationError` if `depth > MAX_SCHEMA_DEPTH`, matching Claude's pattern:

```python
MAX_PROMPT_SCHEMA_DEPTH = 10

def _build_prompt_schema(self, schema: dict[str, OutputField], depth: int = 0) -> dict[str, Any]:
    if depth > self.MAX_PROMPT_SCHEMA_DEPTH:
        raise ValidationError(...)
    ...
```

jrob5756 (commented):
Hello - Thank you so much for your contribution to conductor. One small recommendation. If you can add this check, I'd be happy to merge.

src/conductor/providers/copilot.py
lesandiz: Thanks for reviewing Jason. I applied your suggestion, following existing pattern in claude.py.

jrob5756:
Sorry @lesandiz - it looks like there are some build failures. Could you please address? Thanks!

PR #104, opened 2026-04-21
feat(providers): add claude-agent-sdk provider

lesandiz:
## Summary

- Adds a new `claude-agent-sdk` provider that delegates the agentic loop, tool execution, retry logic, and structured output extraction to the Claude Code CLI via the `claude-agent-sdk` package
- Unlike the raw `claude` provider (~1900 lines), this provider is ~250 lines — the SDK handles the heavy lifting
- Tools and MCP servers are managed by the CLI's own configuration, not workflow YAML fields

## What's included

**New files:**
- `src/conductor/providers/claude_agent_sdk.py` — `ClaudeAgentSdkProvider` implementation
- `tests/test_providers/test_claude_agent_sdk.py` — 22 tests using real SDK types
- `examples/test-claude-agent-sdk.yaml` — example workflow

**Modified files:**
- `src/conductor/config/schema.py` — added `"claude-agent-sdk"` to provider Literal types
- `src/conductor/providers/factory.py` — added factory case for the new provider
- `src/conductor/providers/__init__.py` — exported `ClaudeAgentSdkProvider`
- `pyproject.toml` — added `claude-agent-sdk` dependency
- Docs: `AGENTS.md`, `README.md`, `docs/cli-reference.md`, `docs/configuration.md`, `docs/providers/comparison.md`, `.claude/skills/conductor/SKILL.md`, `.claude/skills/conductor/references/yaml-schema.md`

## Key design decisions

- **Message dispatch uses `type(message).__name__`** — the Python SDK classes don't set a `.type` attribute (unlike the TypeScript SDK)
- **Tool results paired via `pending_tools` dict** — tracks `tool_use_id → tool_name` from AssistantMessage ToolUseBlocks, then matches them to ToolResultBlocks in subsequent UserMessages
- **Tests use real SDK types** (not fakes) — imports `AssistantMessage`, `TextBlock`, etc. directly from `claude_agent_sdk` to catch API contract drift
- **Tools/MCP servers not passed through** — the CLI manages its own toolset; workflow-level `tools` and `runtime.mcp_servers` are ignored (documented in comparison.md and yaml-schema.md)
- **Provider parity** — achieves event and output parity with copilot/claude providers; retry and tool execution are delegated to the SDK (documented in AGENTS.md)

## Test plan

- [x] `uv run ruff check src tests` — lint passes
- [x] `uv run ruff format --check src tests` — format passes
- [x] `uv run pytest tests/test_providers/test_claude_agent_sdk.py -v` — 22 tests pass
- [x] `uv run pytest -m "not real_api and not performance"` — full suite passes
- [x] `uv run conductor validate examples/test-claude-agent-sdk.yaml` — validates
- [x] `uv run conductor run examples/test-claude-agent-sdk.yaml --input question="What is Claude Design?"` — executes successfully with web search


jrob5756 (approved):
Apologies for the delay on this one. I wanted to make sure we had the right model in place for what I'm now calling **experimental providers** before merging — providers that delegate part of the agentic loop to an upstream SDK or framework and therefore can't honor every parity rule in `AGENTS.md` (streaming events, MCP passthrough, full `reasoning.effort` support, native structured output, etc.).

That framework is now captured in #241: a `ProviderCapabilities` descriptor every provider declares against, `conductor validate` cross-checks workflow features against those capabilities, console/dashboard surface an "experimental" banner with the upstream pin, and `AGENTS.md` gets explicit allowed carve-outs and promotion criteria so the tier doesn't become permanent purgatory.

This PR is the canonical first member of that tier — narrow, predictable carve-outs (tools/MCP delegated to the CLI, retries delegated to the SDK), event/output parity preserved. It's the proof point that the model works.

Plan from here:

1. Merge this as-is.
2. Implement #241 in a follow-up PR — adds capabilities + tier metadata + validate-time cross-checks + the banner. At that point `claude-agent-sdk` gets reclassified as `tier: experimental` with declared capabilities, and Copilot/Claude pick up `tier: stable` capability declarations too (they need them so the validator catches model-level mismatches, not just provider-level ones).

Thanks for the patience and for the careful tests-against-real-SDK-types choice — that catches API drift exactly where it should be caught.

PR #107, opened 2026-04-21
feat(config): add optional metadata dict to workflow definition

### LINKED ISSUE #106: feat(config): add optional metadata dict to workflow definition
## Problem

External tooling (dashboards, trackers, CI integrations) that consumes conductor event logs currently has no structured way to know *what kind* of workflow is running or how to enrich its data. The dashboard, for example, hardcodes ADO URL templates, twig SQLite DB paths, and work-item extraction logic — all tightly coupled to specific workflow implementations.

This coupling means:
- Non-twig workflows get wrong ADO links or no enrichment at all
- Adding support for a new tracker (GitHub Issues, Jira) requires dashboard code changes
- Workflow authors can't declare their own integration metadata

## Proposal

Add an optional `metadata` dict to the `workflow` section of the YAML config:

```yaml
workflow:
  name: twig-sdlc
  entry_point: intake
  metadata:
    tracker: ado
    project_url: https://dev.azure.com/org/Project
    work_item_id_agent: intake
    work_item_id_field: epic_id
```

The metadata is:
- Passed through to the `workflow_started` event as-is
- Available to any event log consumer without parsing YAML source
- Completely optional — existing workflows emit `"metadata": {}`

### Dashboard enrichment plugins
A dashboard can dispatch to enricher modules based on `metadata.tracker`:
- `tracker: ado` → look up work item hierarchy from twig SQLite DB, generate ADO links
- `tracker: github` → fetch issue status from GitHub API
- `tracker: jira` → query Jira REST API for ticket details

### Work item ID extraction
Instead of hardcoding "look for `epic_id` in the `intake` agent output", the dashboard reads `metadata.work_item_id_agent` and `metadata.work_item_id_field` to generically extract the tracking ID from any agent's structured output.

### Custom workflow tagging
Teams can add arbitrary tags for filtering, grouping, or routing:
```yaml
metadata:
  team: platform
  environment: staging
  priority: high
```

## Implementation

Two changes (~8 lines total):
1. Add `metadata: dict[str, Any] = Field(default_factory=dict)` to `WorkflowDef` in `schema.py`
2. Include `"metadata": self.config.workflow.metadata` in the `workflow_started` event emission in `workflow.py`

No breaking changes. No migration needed. Default is empty dict.

PolyphonyRequiem:
## Summary

Adds an optional `metadata` dict to workflow configuration with two binding paths:

1. **Static** — declared in the workflow YAML
2. **Dynamic** — injected at runtime via `--metadata` / `-m` CLI flags

Both are merged (CLI wins on conflicts) and included verbatim in the `workflow_started` event, enabling downstream consumers to adapt behavior without parsing YAML source.

Closes #106

## Changes

- **`src/conductor/config/schema.py`**: Add `metadata: dict[str, Any]` field to `WorkflowDef` (empty dict default)
- **`src/conductor/engine/workflow.py`**: Include `metadata` in the `workflow_started` event data
- **`src/conductor/cli/app.py`**: Add `--metadata` / `-m` CLI option, parsed separately from `--input`
- **`src/conductor/cli/run.py`**: Accept `metadata` param, merge CLI metadata on top of YAML metadata after config load
- **`src/conductor/cli/bg_runner.py`**: Forward `--metadata` flags to background child process

### YAML (static)
```yaml
workflow:
  name: twig-sdlc
  entry_point: intake
  metadata:
    tracker: ado
    project_url: https://dev.azure.com/org/Project
```

### CLI (dynamic, merged on top)
```bash
conductor run twig-sdlc.yaml --metadata work_item_id=1814
```

### Result in event log
```json
{
  "type": "workflow_started",
  "data": {
    "name": "twig-sdlc",
    "metadata": {
      "tracker": "ado",
      "project_url": "https://dev.azure.com/org/Project",
      "work_item_id": "1814"
    }
  }
}
```

## Backward Compatibility

- `metadata` defaults to `{}` — existing workflows and CLI invocations are completely unaffected
- `--metadata` is optional — omitting it changes nothing
- No changes to event format beyond the additive `metadata` key
- All existing schema tests pass (100/100)

### DIFF
--- a/src/conductor/cli/app.py
+++ b/src/conductor/cli/app.py
@@ -238,6 +238,14 @@ def run(
             help="Workflow inputs in name=value format. Can be repeated.",
         ),
     ] = None,
+    raw_metadata: Annotated[
+        list[str] | None,
+        typer.Option(
+            "--metadata",
+            "-m",
+            help="Workflow metadata in key=value format. Merged on top of YAML metadata. Can be repeated.",
+        ),
+    ] = None,
     dry_run: Annotated[
         bool,
         typer.Option(
@@ -300,12 +308,14 @@ def run(
 
     Execute a multi-agent workflow defined in the specified YAML file.
     Workflow inputs can be provided using --input flags.
+    Metadata can be provided using --metadata flags (merged on top of YAML metadata).
 
     \b
     Examples:
         conductor run workflow.yaml
         conductor run workflow.yaml --input question="What is Python?"
         conductor run workflow.yaml -i question="Hello" -i context="Programming"
+        conductor run workflow.yaml --metadata tracker=ado -m work_item_id=1814
         conductor run workflow.yaml --provider copilot
         conductor run workflow.yaml --dry-run
         conductor run workflow.yaml --skip-gates
@@ -377,6 +387,11 @@ def run(
     # Also parse --input.name=value style from sys.argv
     inputs.update(InputCollector.extract_from_args())
 
+    # Parse --metadata key=value flags (separate from inputs)
+    cli_metadata: dict[str, str] = {}
+    if raw_metadata:
+        cli_metadata.update(parse_input_flags(raw_metadata))
+
     # Resolve log file path
     resolved_log_file: Path | None = None
     if log_file is not None:
@@ -398,6 +413,7 @@ def run(
                 log_file=resolved_log_file,
                 no_interactive=True,  # Always non-interactive in background
                 web_port=web_port,
+                metadata=cli_metadata,
             )
             console.print(f"[bold cyan]Dashboard:[/bold cyan] {url}")
             console.print(
@@ -422,6 +438,7 @@ def run(
                 web=web,
                 web_port=web_port,
                 web_bg=web_bg,
+                metadata=cli_metadata,
             )
         )
 
--- a/src/conductor/cli/app.py
+++ b/src/conductor/cli/app.py
@@ -363,6 +363,7 @@ def run(
         display_execution_plan,
         generate_log_path,
         parse_input_flags,
+        parse_metadata_flags,
         run_workflow_async,
     )
 
@@ -390,10 +391,10 @@ def run(
     # Also parse --input.name=value style from sys.argv
     inputs.update(InputCollector.extract_from_args())
 
-    # Parse --metadata key=value flags (separate from inputs)
+    # Parse --metadata key=value flags (no type coercion — values stay as strings)
     cli_metadata: dict[str, str] = {}
     if raw_metadata:
-        cli_metadata.update(parse_input_flags(raw_metadata))
+        cli_metadata.update(parse_metadata_flags(raw_metadata))
 
     # Resolve log file path
     resolved_log_file: Path | None = None
--- a/src/conductor/cli/app.py
+++ b/src/conductor/cli/app.py
@@ -243,7 +243,10 @@ def run(
         typer.Option(
             "--metadata",
             "-m",
-            help="Workflow metadata in key=value format. Merged on top of YAML metadata. Can be repeated.",
+            help=(
+                "Workflow metadata in key=value format. "
+                "Merged on top of YAML metadata. Can be repeated."
+            ),
         ),
     ] = None,
     dry_run: Annotated[
--- a/src/conductor/cli/bg_runner.py
+++ b/src/conductor/cli/bg_runner.py
@@ -112,7 +112,7 @@ def launch_background(
     # Forward metadata
     if metadata:
         for key, value in metadata.items():
-            cmd.extend(["--metadata", f"{key}={value}"])
+            cmd.extend(["--metadata", f"{key}={_serialize_value(value)}"])
 
     if provider_override:
         cmd.extend(["--provider", provider_override])
--- a/src/conductor/cli/bg_runner.py
+++ b/src/conductor/cli/bg_runner.py
@@ -62,6 +62,7 @@ def launch_background(
     log_file: Path | None = None,
     no_interactive: bool = True,
     web_port: int = 0,
+    metadata: dict[str, str] | None = None,
 ) -> str:
     """Fork a detached child process running the workflow with a web dashboard.
 
@@ -77,6 +78,7 @@ def launch_background(
         log_file: Optional log file path.
         no_interactive: Whether to disable interactive mode (always True for bg).
         web_port: Desired port (0 = auto-select).
+        metadata: Optional CLI metadata key=value pairs.
 
     Returns:
         The dashboard URL (e.g. ``http://127.0.0.1:8080``).
@@ -106,6 +108,11 @@ def launch_background(
     # Forward inputs
     for key, value in inputs.items():
         cmd.extend(["--input", f"{key}={_serialize_value(value)}"])
+
+    # Forward metadata
+    if metadata:
+        for key, value in metadata.items():
+            cmd.extend(["--metadata", f"{key}={value}"])
 
     if provider_override:
         cmd.extend(["--provider", provider_override])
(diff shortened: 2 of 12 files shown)


src/conductor/engine/workflow.py
```diff
@@ -411,6 +422,43 @@ def _conductor_version() -> str:
...
+        import platform as _platform
+        import sys
+        from datetime import UTC, datetime
+
+        system: dict[str, Any] = {
+            "pid": os.getpid(),
+            "platform": sys.platform,
+            "python_version": _platform.python_version(),
+            "conductor_version": self._conductor_version(),
+            "cwd": os.getcwd(),
```
jrob5756: **Critical: `os.getcwd()` can raise unhandled `OSError`**

If the working directory is deleted between process start and this call (CI runners, containers, temp-dir cleanup), this raises `FileNotFoundError`/`OSError`. Unlike `_conductor_version()` which is wrapped in try/except, this method has no protection — and it's called at the top of `_execute_loop()`, so an unhandled exception here crashes the entire workflow before any agent runs with a confusing error.

```python
try:
    cwd = os.getcwd()
except OSError:
    cwd = "<unavailable>"
```

Or wrap the entire `_build_system_metadata()` body in try/except to match `_conductor_version()`'s pattern.

src/conductor/web/server.py
```diff
@@ -156,6 +156,25 @@ async def favicon() -> FileResponse:
...
+            for event in self._event_history:
+                if event.get("type") == "workflow_started":
+                    data = event.get("data", {})
+                    info = {
+                        "run_id": data.get("run_id", ""),
+                        "log_file": data.get("log_file", ""),
+                        "workflow_name": data.get("name", ""),
+                        "started_at": event.get("timestamp", 0),
+                        "metadata": data.get("metadata", {}),
+                        "system": data.get("system", {}),
```
jrob5756: **Critical: Unauthenticated endpoint exposes sensitive system info**

The `system` dict contains PID, filesystem paths (`cwd`, `log_file`), platform details, and parent PID — all served via an unauthenticated HTTP GET to any network client that can reach the dashboard port.

Consider:
- Omitting `system` from `/api/info` entirely (it's still in the event log for diagnostics)
- Or limiting to non-sensitive fields only (e.g., `conductor_version`, `started_at`)

src/conductor/cli/app.py
```diff
@@ -377,6 +390,11 @@ def run(
     # Also parse --input.name=value style from sys.argv
     inputs.update(InputCollector.extract_from_args())
 
+    # Parse --metadata key=value flags (separate from inputs)
+    cli_metadata: dict[str, str] = {}
+    if raw_metadata:
+        cli_metadata.update(parse_input_flags(raw_metadata))
```
jrob5756: **Important: `parse_input_flags` silently coerces metadata values**

`parse_input_flags` calls `coerce_value()`, which converts `"42"` → `int`, `"true"` → `bool`, `"null"` → `None`. So `--metadata work_item_id=42` silently becomes `{"work_item_id": 42}` (int, not string). The type annotation says `dict[str, str]` but actual values will be `int | bool | None | list | dict`.

Metadata values should stay as strings since they're opaque key-value pairs. Consider a dedicated `parse_metadata_flags()` that splits on first `=` without coercion, or add a `coerce=False` parameter to `parse_input_flags`.

src/conductor/cli/bg_runner.py
```diff
@@ -107,6 +109,11 @@ def launch_background(
     for key, value in inputs.items():
         cmd.extend(["--input", f"{key}={_serialize_value(value)}"])
 
+    # Forward metadata
+    if metadata:
+        for key, value in metadata.items():
+            cmd.extend(["--metadata", f"{key}={value}"])
```
jrob5756: **Important: Missing `_serialize_value()` — nested metadata breaks in background mode**

Inputs (line 110) use `_serialize_value(value)` to handle non-string types (dicts, lists → JSON). Metadata uses bare `f"{key}={value}"`. If YAML metadata contains nested dicts:

```yaml
metadata:
  config:
    base_url: https://example.com
```

…this produces `config={'base_url': 'https://example.com'}` (Python repr, not JSON), which fails to parse on the child side.

```suggestion
            cmd.extend(["--metadata", f"{key}={_serialize_value(value)}"])
```

src/conductor/cli/run.py
```diff
@@ -1116,6 +1122,10 @@ async def run_workflow_async(
                 event_emitter=emitter,
                 keyboard_listener=listener,
                 web_dashboard=dashboard,
+                run_id=event_log_subscriber.run_id if event_log_subscriber else "",
+                log_file=str(event_log_subscriber.path) if event_log_subscriber else "",
+                dashboard_port=(dashboard._actual_port if dashboard is not None else None),
```
jrob5756: **Important: Accessing private `dashboard._actual_port`**

`_actual_port` is a private attribute. Consider adding a public `@property port` on `WebDashboard` that returns `self._actual_port or self._port`, then use `dashboard.port` here.

src/conductor/engine/workflow.py
```diff
@@ -270,6 +270,10 @@ def __init__(
         keyboard_listener: KeyboardListener | None = None,
         web_dashboard: WebDashboard | None = None,
         _subworkflow_depth: int = 0,
+        run_id: str = "",
+        log_file: str = "",
+        dashboard_port: int | None = None,
+        bg_mode: bool = False,
```
jrob5756: **Suggestion: Consider grouping informational params into a dataclass**

These 4 new params (`run_id`, `log_file`, `dashboard_port`, `bg_mode`) are purely informational — not used for orchestration, only passed through to event data. The constructor already has 10 params; this brings it to 14.

Consider grouping into a dataclass:

```python
@dataclass
class RunContext:
    run_id: str = ""
    log_file: str = ""
    dashboard_port: int | None = None
    bg_mode: bool = False
```

Then pass a single `run_context: RunContext | None = None` parameter. This keeps the constructor clean and makes future additions trivial.

jrob5756 (commented):
Really cool. Thanks for contributing. I have a few comments. Please take a look and we can merge!

src/conductor/engine/workflow.py
PolyphonyRequiem: Good catch! Wrapped os.getcwd() in try/except — falls back to '<unavailable>' on OSError. Matches the defensive pattern used by _conductor_version().

src/conductor/web/server.py
PolyphonyRequiem: Agreed — stripped the system dict from /api/info entirely. The endpoint now only returns non-sensitive fields: run_id, workflow_name, started_at, metadata, and conductor_version. Full diagnostics remain in the event log for debugging.

src/conductor/cli/bg_runner.py
PolyphonyRequiem: Nice find — switched to _serialize_value(value) so nested dicts/lists get proper JSON serialization instead of Python repr. Matches the pattern already used for inputs on line 110.

src/conductor/cli/run.py
PolyphonyRequiem: Added a public port property on WebDashboard that returns _actual_port or _port. Updated run.py to use dashboard.port instead of reaching into the private attribute.

src/conductor/engine/workflow.py
PolyphonyRequiem: Went ahead and did this in this pass — added a RunContext dataclass grouping run_id, log_file, dashboard_port, and bg_mode. WorkflowEngine constructor now takes a single run_context param instead of four. Should make future additions painless.

src/conductor/cli/app.py
PolyphonyRequiem: Great point — added a dedicated parse_metadata_flags() that splits on first = without any coercion. Metadata values stay as raw strings, keeping the dict[str, str] annotation honest.

jrob5756 (approved):
Thanks

Issue #103, opened 2026-04-20
feat(composition): allow self-referential sub-workflows with depth tracking


PolyphonyRequiem:
> **Provenance:** AI-drafted (GitHub Copilot CLI / Claude Opus 4.6), human-reviewed and approved, AI-submitted.

## Problem

The engine currently rejects any workflow that references its own file as a sub-workflow:

```python
# src/conductor/engine/workflow.py
if current_path is not None and sub_path == current_path:
    raise ExecutionError(
        f"Circular sub-workflow reference: agent '{agent.name}' "
        f"references its own workflow file '{agent.workflow}'.",
        suggestion="A workflow cannot reference itself as a sub-workflow.",
    )
```

This prevents recursive workflow patterns where the same logic applies at different levels of a hierarchy (e.g., planning at epic/issue/task level, or any divide-and-conquer pattern).

The depth-tracking infrastructure already exists (`MAX_SUBWORKFLOW_DEPTH=10`, `_subworkflow_depth` counter). The circular reference check is overly conservative — depth tracking already prevents infinite recursion.

## Proposed solution

Remove the file-path equality check. Rely on `MAX_SUBWORKFLOW_DEPTH` for safety (already enforced). The depth counter already increments on each `_execute_subworkflow` call and raises `ExecutionError` at the limit.

**Optional enhancement:** Add a `max_depth` field to `type: workflow` agents so authors can set tighter bounds per-agent:

```yaml
agents:
  - name: plan_children
    type: workflow
    workflow: ./self.yaml
    max_depth: 3                    # override global MAX_SUBWORKFLOW_DEPTH
    input_mapping:
      work_item_id: "{{ item.id }}"
    routes:
      - to: aggregate
```

**What about mutual recursion (A → B → A)?**
The current check only catches direct self-reference. Mutual recursion is already handled by the depth limit. Removing the self-reference check makes the behavior consistent: all circular patterns (direct and mutual) are bounded by depth, not by static analysis.

**Termination guarantee:**
- `MAX_SUBWORKFLOW_DEPTH=10` is the hard ceiling
- Per-agent `max_depth` provides author control
- `max_iterations` on each sub-workflow provides a second safety net
- These compound: a 3-level recursive workflow with 200 iterations each can do at most 600 agent executions total

## Example: fractal planning

With all three issues in this series implemented:

```yaml
# plan.yaml — recursive planner
workflow:
  name: fractal-planner
  input:
    work_item_id: { type: number }
[...]

Issue #102, opened 2026-04-20
feat(composition): allow sub-workflows in for_each groups


PolyphonyRequiem:
> **Provenance:** AI-drafted (GitHub Copilot CLI / Claude Opus 4.6), human-reviewed and approved, AI-submitted.

## Problem

`for_each` groups currently reject `type: workflow` agents:

```python
# src/conductor/config/validator.py
if for_each_group.agent.type == "workflow":
    errors.append(
        f"For-each group '{for_each_group.name}' uses a workflow step..."
    )
```

This prevents dynamic fan-out to sub-workflows. A common pattern — "for each issue in the plan, run a planning/implementation sub-workflow" — is impossible.

## Proposed solution

Remove the validator restriction and wire up `for_each` execution to call `_execute_subworkflow()` for `type: workflow` agents.

```yaml
agents:
  - name: epic_planner
    prompt: "Break this epic into issues"
    output:
      issues: { type: array }
    routes:
      - to: plan_issues

  - name: plan_issues
    type: for_each
    source: epic_planner.output.issues
    as: issue
    max_concurrent: 1   # sequential to avoid branch conflicts
    agent:
      type: workflow
      workflow: ./plan-and-review.yaml
      input_mapping:                        # from #101
        work_item_id: "{{ issue.id }}"
        title: "{{ issue.title }}"
    routes:
      - to: next_step
```

**Key design decisions:**
- `max_concurrent` should default to 1 for workflow steps (sub-workflows may have side effects — git, file system — that conflict when parallel)
- Each `for_each` iteration receives the `item`/`as` variable in scope for `input_mapping` template rendering
- Sub-workflow depth increments per nesting level, shared across all `for_each` iterations (prevents depth explosion)
- Failure mode (`fail_fast` vs `continue`) applies as with regular agents

**Engine change:**
In `_execute_for_each_group()`, when `agent.type == "workflow"`, call `_execute_subworkflow(agent, item_context)` instead of the regular agent execution path. The `input_mapping` templates are rendered with the `for_each` `item` variable in scope.

## Dependency

Requires #101 (`input_mapping`) — without dynamic inputs, every `for_each` iteration would receive identical `workflow.input.*`, defeating the purpose.

## Relationship to other requests

This is **Issue 2 of 3** in a series enabling recursive workflow composition:
- **Issue 1:** #101 — `input_mapping` (prerequisite)
- **Issue 2 (this):** Allow `type: workflow` in `for_each` groups — dynamic fan-out
[...]

PR #109, opened 2026-04-21
feat(composition): dynamic sub-workflow inputs via input_mapping (#101)

### LINKED ISSUE #101: feat(composition): dynamic sub-workflow inputs via input_mapping

## Problem

Today, sub-workflows (`type: workflow`) always receive the parent's `workflow.input.*` values verbatim. The child's `workflow.input.X` equals the parent's `workflow.input.X`. There is no way to pass intermediate results (upstream agent outputs) into a sub-workflow.

This makes sub-workflows useful only for phases that need the same static inputs as the parent. You can't call a sub-workflow in a loop with different parameters each iteration — e.g., "plan issue A" then "plan issue B" — because the sub-workflow sees the same `workflow.input.*` both times.

**Current code** (`src/conductor/engine/workflow.py`):
```python
workflow_ctx = context.get("workflow", {})
sub_inputs: dict[str, Any] = (
    dict(workflow_ctx.get("input", {})) if isinstance(workflow_ctx, dict) else {}
)
return await child_engine.run(sub_inputs)
```

## Proposed solution

Add an optional `input_mapping` field to `type: workflow` agents. Each key is a sub-workflow input name; each value is a Jinja2 expression evaluated in the parent's context.

```yaml
agents:
  - name: plan_issue
    type: workflow
    workflow: ./plan-and-review.yaml
    input_mapping:
      work_item_id: "{{ task_manager.output.current_issue_id }}"
      title: "{{ task_manager.output.current_issue_title }}"
      description: "{{ task_manager.output.current_issue_description }}"
    output:
      plan_path: { type: string }
      plan_summary: { type: string }
    routes:
      - to: next_step
```

**Resolution rules:**
1. If `input_mapping` is present, render each value as a Jinja2 template against the parent's full context, then pass the resulting dict as the sub-workflow's inputs. Parent's `workflow.input.*` is NOT inherited.
2. If `input_mapping` is absent, current behavior is preserved (parent's `workflow.input.*` is forwarded).

**Schema change** (`AgentDef`):
```python
input_mapping: dict[str, str] | None = None
"""Optional mapping of sub-workflow input names to Jinja2 expressions.
Evaluated in the parent context. Only valid for type='workflow'."""
```

**Engine change** — roughly:
```python
if agent.input_mapping:
    sub_inputs = {}
    for key, template in agent.input_mapping.items():
        sub_inputs[key] = self._render_template(template, agent_context)
else:
    # existing behavior
    workflow_ctx = context.get("workflow", {})
    sub_inputs = dict(workflow_ctx.get("input", {}))
```

**Validation:**
- `input_mapping` is only valid when `type: workflow`
- Each value must be a valid Jinja2 expression
- Keys should match the sub-workflow's declared input names (warning, not error)

## Why this matters

This is the foundation for all advanced composition patterns. Without it, sub-workflows are limited to "same inputs as parent" which makes them suitable for one-shot phases but not for parameterized or iterative use.

This is **Issue 1 of 3** in a series enabling recursive workflow composition:
- **Issue 1 (this):** `input_mapping` — foundation for parameterized sub-workflows
- **Issue 2:** Allow `type: workflow` in `for_each` groups — dynamic fan-out (depends on this)
- **Issue 3:** Allow self-referential workflows — controlled recursion (depends on Issues 1 + 2)

- **Issue 2:** #102 — Allow `type: workflow` in `for_each` groups — dynamic fan-out (depends on this)
- **Issue 3:** #103 — Allow self-referential workflows — controlled recursion (depends on Issues 1 + 2)

PolyphonyRequiem:
## Summary

Adds an optional `input_mapping` field to `type: workflow` agents, enabling parameterized sub-workflow invocation. Each key maps a sub-workflow input name to a Jinja2 expression evaluated in the parent's context.

Closes #101

## Example

```yaml
agents:
  - name: plan_issue
    type: workflow
    workflow: ./plan-and-review.yaml
    input_mapping:
      work_item_id: "{{ task_manager.output.current_issue_id }}"
      title: "{{ task_manager.output.current_issue_title }}"
```

Without `input_mapping`, parent's `workflow.input.*` is forwarded (existing behavior preserved).

## Changes

- Schema: `input_mapping: dict[str, str] | None` on `AgentDef`, validated only for `type: workflow`
- Engine: render `input_mapping` templates against parent context in `_execute_subworkflow`
- Tests included

## Dependency

Foundation for #102 (for_each with workflows) and #103 (self-referential workflows).

### DIFF
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -546,17 +546,21 @@ class WorkflowEngine:
 
         # Build sub-workflow inputs from the parent context
         sub_inputs: dict[str, Any]
-        if agent.input_mapping:
+        if agent.input_mapping is not None:
             # Dynamic inputs: render each Jinja2 expression against parent context
             renderer = TemplateRenderer()
             sub_inputs = {}
             for key, template_expr in agent.input_mapping.items():
-                rendered = renderer.render(template_expr, context)
-                # Attempt to parse rendered values as JSON for non-string types
                 try:
-                    sub_inputs[key] = json.loads(rendered)
-                except (json.JSONDecodeError, ValueError):
-                    sub_inputs[key] = rendered
+                    rendered = renderer.render(template_expr, context)
+                except Exception as e:
+                    raise ExecutionError(
+                        f"Failed to render input_mapping key '{key}' for agent "
+                        f"'{agent.name}': {e}",
+                        suggestion=f"Check that the expression '{template_expr}' "
+                        "references valid context variables.",
+                    ) from e
+                sub_inputs[key] = rendered
         else:
             # Default: forward parent's workflow.input.* values
             workflow_ctx = context.get("workflow", {})
@@ -577,14 +581,6 @@ class WorkflowEngine:
             web_dashboard=self._web_dashboard,
             _subworkflow_depth=self._subworkflow_depth + 1,
         )
-
-        # Inject parent agent outputs into the child workflow's context.
-        # This allows sub-workflow agents that declare parent agents in their
-        # input: list (e.g., task_manager.output?) to access parent state
-        # even when input_mapping doesn't cover all fields.
-        for key, value in context.items():
-            if key not in ("workflow", "context") and isinstance(value, dict):
-                child_engine.context.agent_outputs[key] = value.get("output", value)
 
         return await child_engine.run(sub_inputs)
 
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -578,6 +578,14 @@ class WorkflowEngine:
             _subworkflow_depth=self._subworkflow_depth + 1,
         )
 
+        # Inject parent agent outputs into the child workflow's context.
+        # This allows sub-workflow agents that declare parent agents in their
+        # input: list (e.g., task_manager.output?) to access parent state
+        # even when input_mapping doesn't cover all fields.
+        for key, value in context.items():
+            if key not in ("workflow", "context") and isinstance(value, dict):
+                child_engine.context.agent_outputs[key] = value.get("output", value)
+
         return await child_engine.run(sub_inputs)
 
     def _get_context_window_for_agent(self, agent: AgentDef) -> int | None:
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -545,11 +545,24 @@ class WorkflowEngine:
             ) from exc
 
         # Build sub-workflow inputs from the parent context
-        # Extract workflow.input.* values from the parent context
-        workflow_ctx = context.get("workflow", {})
-        sub_inputs: dict[str, Any] = (
-            dict(workflow_ctx.get("input", {})) if isinstance(workflow_ctx, dict) else {}
-        )
+        sub_inputs: dict[str, Any]
+        if agent.input_mapping:
+            # Dynamic inputs: render each Jinja2 expression against parent context
+            renderer = TemplateRenderer()
+            sub_inputs = {}
+            for key, template_expr in agent.input_mapping.items():
+                rendered = renderer.render(template_expr, context)
+                # Attempt to parse rendered values as JSON for non-string types
+                try:
+                    sub_inputs[key] = json.loads(rendered)
+                except (json.JSONDecodeError, ValueError):
+                    sub_inputs[key] = rendered
+        else:
+            # Default: forward parent's workflow.input.* values
+            workflow_ctx = context.get("workflow", {})
+            sub_inputs = (
+                dict(workflow_ctx.get("input", {})) if isinstance(workflow_ctx, dict) else {}
+            )
 
         # Create child engine inheriting provider/registry but with deeper depth
         child_engine = WorkflowEngine(
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -555,8 +555,7 @@ class WorkflowEngine:
                     rendered = renderer.render(template_expr, context)
                 except Exception as e:
                     raise ExecutionError(
-                        f"Failed to render input_mapping key '{key}' for agent "
-                        f"'{agent.name}': {e}",
+                        f"Failed to render input_mapping key '{key}' for agent '{agent.name}': {e}",
                         suggestion=f"Check that the expression '{template_expr}' "
                         "references valid context variables.",
                     ) from e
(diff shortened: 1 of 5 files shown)


src/conductor/engine/workflow.py
```diff
@@ -565,6 +578,14 @@ async def _execute_subworkflow(
             _subworkflow_depth=self._subworkflow_depth + 1,
         )
 
+        # Inject parent agent outputs into the child workflow's context.
+        # This allows sub-workflow agents that declare parent agents in their
+        # input: list (e.g., task_manager.output?) to access parent state
+        # even when input_mapping doesn't cover all fields.
+        for key, value in context.items():
+            if key not in ("workflow", "context") and isinstance(value, dict):
+                child_engine.context.agent_outputs[key] = value.get("output", value)
```
jrob5756: **Critical: Unconditional parent context injection breaks sub-workflow isolation**

This block runs for **every** sub-workflow, not just those with `input_mapping`. This is a backward-incompatible behavioral change — previously child workflows started with a clean `agent_outputs`.

**Bugs this causes:**
1. **Namespace collision**: If parent and child both have an agent `analyzer`, the parent's output is pre-injected. Before the child's own `analyzer` runs, `{{ analyzer.output }}` resolves to stale parent data instead of raising an error.
2. **Context bypass**: Direct assignment to `agent_outputs` skips `context.store()`, so `execution_history` is out of sync.
3. **State leakage**: `value.get("output", value)` — if no `"output"` key exists, the entire internal dict (including `model`, `tokens`, etc.) is injected.

**Fix**: Either:
- Guard with `if agent.input_mapping:` (make it opt-in), or
- Remove entirely — `input_mapping` already provides controlled access to parent data. Blanket injection is redundant and unsafe.

src/conductor/engine/workflow.py
```diff
@@ -545,11 +545,24 @@ async def _execute_subworkflow(
...
+        sub_inputs: dict[str, Any]
+        if agent.input_mapping:
+            # Dynamic inputs: render each Jinja2 expression against parent context
+            renderer = TemplateRenderer()
+            sub_inputs = {}
+            for key, template_expr in agent.input_mapping.items():
+                rendered = renderer.render(template_expr, context)
+                # Attempt to parse rendered values as JSON for non-string types
+                try:
+                    sub_inputs[key] = json.loads(rendered)
```
jrob5756: **Critical: Implicit `json.loads` coercion is a foot-gun**

This silently coerces types by content sniffing:
- `"42"` → `int` · `"true"` → `bool` · `"null"` → `None`

Users have no way to pass the literal string `"true"` as a string input. The `"null"` → `None` case is especially dangerous — a missing/empty variable becomes `None` and the sub-workflow may not catch it.

**Suggestion**: Remove `json.loads` and always pass rendered values as strings (matching how `--input` CLI works). If structured data is needed, let users use the `{{ expr | json }}` filter explicitly.

src/conductor/engine/workflow.py
```diff
@@ -545,11 +545,24 @@ async def _execute_subworkflow(
...
-        sub_inputs: dict[str, Any] = (
-            dict(workflow_ctx.get("input", {})) if isinstance(workflow_ctx, dict) else {}
-        )
+        sub_inputs: dict[str, Any]
+        if agent.input_mapping:
+            # Dynamic inputs: render each Jinja2 expression against parent context
+            renderer = TemplateRenderer()
+            sub_inputs = {}
+            for key, template_expr in agent.input_mapping.items():
+                rendered = renderer.render(template_expr, context)
```
jrob5756: **Important: Template errors lack key context**

If `renderer.render()` raises `TemplateError` (e.g., undefined variable), the error message won't indicate *which* `input_mapping` key failed. With many keys, debugging is a guessing game.

```suggestion
            for key, template_expr in agent.input_mapping.items():
                try:
                    rendered = renderer.render(template_expr, context)
                except Exception as e:
                    raise ExecutionError(
                        f"Failed to render input_mapping key '{key}' for agent "
                        f"'{agent.name}': {e}",
                        suggestion=f"Check that the expression '{template_expr}' "
                        "references valid context variables.",
                    ) from e
```

src/conductor/engine/workflow.py
```diff
@@ -545,11 +545,24 @@ async def _execute_subworkflow(
             ) from exc
 
         # Build sub-workflow inputs from the parent context
-        # Extract workflow.input.* values from the parent context
-        workflow_ctx = context.get("workflow", {})
-        sub_inputs: dict[str, Any] = (
-            dict(workflow_ctx.get("input", {})) if isinstance(workflow_ctx, dict) else {}
-        )
+        sub_inputs: dict[str, Any]
+        if agent.input_mapping:
```
jrob5756: **Suggestion: `if agent.input_mapping:` — empty dict `{}` is falsy**

An explicit `input_mapping: {}` in YAML means "pass no inputs", but `{}` is falsy in Python, so this falls through to the default behavior (forwarding all `workflow.input.*`). Consider using `if agent.input_mapping is not None:` to distinguish "not set" from "explicitly empty".

tests/test_config/test_workflow_type_schema.py
```diff
@@ -284,3 +284,58 @@ def test_workflow_with_routes_to_agents(self) -> None:
...
+                ],
+                input_mapping={"key": "{{ value }}"},
+            )
+
+    def test_input_mapping_on_script_raises(self) -> None:
+        """Test that input_mapping on a script agent raises ValidationError."""
+        with pytest.raises(ValidationError, match="input_mapping"):
+            AgentDef(
+                name="script",
+                type="script",
```
jrob5756, line 338: **Important: No runtime tests for the core feature**

These 5 schema tests are good, but there are **zero tests** for the 34 new runtime lines in `workflow.py`. Missing coverage:

1. Jinja2 rendering of `input_mapping` expressions against parent context
2. JSON type coercion behavior (`"42"` → int, `"true"` → bool)
3. Error when Jinja2 references an undefined variable
4. Parent context injection and namespace collision with child agents
5. End-to-end: parent output → `input_mapping` → child sub-workflow receives correct inputs
6. Backward compat: no `input_mapping` → existing behavior preserved

Recommend adding these in `tests/test_engine/test_subworkflow.py` using the existing mock provider pattern.

jrob5756 (commented):
Looks good. Some small comments.

src/conductor/engine/workflow.py
PolyphonyRequiem: was unaware of this, thank you!

src/conductor/engine/workflow.py
PolyphonyRequiem: I think this current commit should cover it. 



 I've removed the parent context injection block entirely. 
 input_mapping already provides explicit, controlled access to parent data via Jinja2 expressions, making
 the blanket injection both redundant and harmful for the reasons you outlined

jrob5756 (commented):
Looks good. Just remove the version increment of the uv.lock file. There's a conflict and I can update it when I cut a new release. Thanks!

PR #110, opened 2026-04-21
feat(composition): allow sub-workflows in for_each groups (#102)

### LINKED ISSUE #102: feat(composition): allow sub-workflows in for_each groups

PolyphonyRequiem:
## Summary

Removes the validator restriction that prevented `type: workflow` agents in `for_each` groups, enabling dynamic fan-out to sub-workflows.

Closes #102

## Example

```yaml
for_each:
  - name: plan_issues
    source: epic_planner.output.issues
    as: issue
    max_concurrent: 1
    agent:
      type: workflow
      workflow: ./plan-and-review.yaml
      input_mapping:
        work_item_id: "{{ issue.id }}"
        title: "{{ issue.title }}"
```

## Changes

- Remove validator rejection of workflow agents in for_each groups
- Wire up for_each execution to call `_execute_subworkflow()` for workflow agents
- Emit `subworkflow_started`/`subworkflow_completed` events per iteration
- Tests included

## Dependency

Builds on #101 (input_mapping) — PR #109. Foundation for #103 (self-referential workflows).

### DIFF
--- a/src/conductor/config/schema.py
+++ b/src/conductor/config/schema.py
@@ -475,6 +475,24 @@ class AgentDef(BaseModel):
         workflow: ./research-pipeline.yaml
     """
 
+    input_mapping: dict[str, str] | None = None
+    """Optional mapping of sub-workflow input names to Jinja2 expressions.
+
+    Each key is a sub-workflow input parameter name. Each value is a Jinja2
+    template expression evaluated against the parent workflow's context.
+
+    When present, the rendered values are passed as the sub-workflow's inputs
+    instead of forwarding the parent's workflow.input.* values.
+
+    Only valid for type='workflow' agents.
+
+    Example::
+
+        input_mapping:
+          work_item_id: "{{ task_manager.output.current_issue_id }}"
+          title: "{{ task_manager.output.current_issue_title }}"
+    """
+
     max_session_seconds: float | None = Field(None, ge=1.0)
     """Maximum wall-clock duration for this agent's session in seconds.
 
@@ -529,6 +547,8 @@ class AgentDef(BaseModel):
                 raise ValueError("human_gate agents require 'options'")
             if not self.prompt:
                 raise ValueError("human_gate agents require 'prompt'")
+            if self.input_mapping:
+                raise ValueError("human_gate agents cannot have 'input_mapping'")
         elif self.type == "script":
             if not self.command:
                 raise ValueError("script agents require 'command'")
@@ -555,6 +575,8 @@ class AgentDef(BaseModel):
                 raise ValueError("script agents cannot have 'max_agent_iterations'")
             if self.retry is not None:
                 raise ValueError("script agents cannot have 'retry'")
+            if self.input_mapping:
+                raise ValueError("script agents cannot have 'input_mapping'")
         elif self.type == "workflow":
             if not self.workflow:
                 raise ValueError("workflow agents require 'workflow' path")
@@ -578,6 +600,13 @@ class AgentDef(BaseModel):
                 raise ValueError("workflow agents cannot have 'max_agent_iterations'")
             if self.retry is not None:
                 raise ValueError("workflow agents cannot have 'retry'")
+        else:
+            # Regular agent or human_gate — input_mapping is not valid
+            if self.input_mapping:
+                raise ValueError(
+                    f"'{self.type or 'agent'}' agents cannot have 'input_mapping' "
+                    "(only workflow agents support input_mapping)"
+                )
         return self
 
 
--- a/src/conductor/config/validator.py
+++ b/src/conductor/config/validator.py
@@ -111,17 +111,12 @@ def validate_workflow_config(config: WorkflowConfig) -> list[str]:
         parallel_errors = _validate_parallel_groups(config)
         errors.extend(parallel_errors)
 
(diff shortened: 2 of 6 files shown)


PolyphonyRequiem:
Converted to draft — lint check failing. Schema and type check pass. Needs lint fixes before merging. Depends on PR #109 (input_mapping) landing first.

PolyphonyRequiem:
### Follow-up: companion commit for the recursive-gate-navigation fix in #113

Heads up — I'm pushing a small commit to this branch that wires the for_each-of-workflow event path into the dashboard breadcrumb / sticky-follow infrastructure being introduced in #113.

**What it does**

The for_each-of-workflow path in this PR currently emits `subworkflow_started` / `_completed` / `_failed` per iteration, but with no per-iteration identity beyond `agent_name`. Under `max_concurrent > 1`, the dashboard treats all concurrent iterations as the same context and stacks them as descendants of one another instead of laying them out as siblings of the parent for_each group.

This commit adds:

- `parent_path: list[str]` on the for_each subworkflow lifecycle emits (the slot-key path of the engine that spawned the iteration; `[]` at the root).
- `slot_key: str` on the same emits, formatted as `"{group.name}[{key}]"` so concurrent iterations under one group get distinct, stable identities.
- `item_key`, `iteration` on the lifecycle emits so the dashboard can display per-iteration state and labels.
- A `try/except` wrapping the per-iteration `_execute_subworkflow_with_inputs` call so a `subworkflow_failed` event is emitted on iteration failure (matching the sequential path's behavior).
- `slot_key` parameter threaded into `_execute_subworkflow_with_inputs` so the spawned child engine inherits the correct `_dashboard_context_path` (introduced in #113).
- Deduplicated `_execute_subworkflow_with_inputs` (there were two overlapping definitions from prior merges; the second was shadowing the first and breaking the for_each call site under one of the signatures).

**Why this is safe to land in either order**

This commit references `_dashboard_context_path` via `getattr(self, "_dashboard_context_path", [])`, which is the field added by #113. On this branch alone (without #113), the field doesn't exist and the emits include `parent_path: []` and `slot_key: "<group>[<key>]"` — no AttributeError, no regression. The frontend store on this branch alone still ignores those fields and stacks iterations the same way it does today (the broken behavior is unchanged by this commit until #113's frontend refactor lands).

Once #113 merges, the engine infra (`_dashboard_context_path`, auto-stamping `_emit`, sticky-follow, the store handler refactor) is in place, and these emit fields immediately become functional — concurrent for_each iterations lay out correctly as siblings, and gates inside any iteration are reachable via sticky-follow. No follow-up PR required.

Order of merge between this and #113 does not matter.

See #113 for the broader fix description and the `TestSubWorkflowDashboardPath` engine test that covers both the sequential path (in #113) and the for_each path (added here).

jrob5756 (approved):
Event-correlation concern addressed by 93eea5e (slot_key/parent_path/iteration on subworkflow lifecycle events + paired completed/failed + regression test). LGTM.

PR #111, opened 2026-04-21
feat(composition): allow self-referential sub-workflows with depth tracking (#103)

### LINKED ISSUE #103: feat(composition): allow self-referential sub-workflows with depth tracking

PolyphonyRequiem:
## Summary

Removes the overly conservative circular reference check that prevented workflows from referencing themselves. The existing `MAX_SUBWORKFLOW_DEPTH=10` limit already prevents infinite recursion — the path equality check was redundant.

Closes #103

## Example

```yaml
agents:
  - name: plan_children
    type: workflow
    workflow: ./self.yaml
    max_depth: 3
    input_mapping:
      work_item_id: "{{ item.id }}"
```

## Changes

- Remove file-path equality check in `_execute_subworkflow()`
- Add optional `max_depth` field to `AgentDef` for per-agent depth control
- Depth tracking (`_subworkflow_depth`) continues to enforce the global limit
- Tests included

## Dependency

Builds on #101 (input_mapping, PR #109) and #102 (for_each workflows, PR #110).

### DIFF
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -2688,9 +2688,7 @@ class WorkflowEngine:
                     else:
                         wf_ctx = agent_context.get("workflow", {})
                         sub_inputs = (
-                            dict(wf_ctx.get("input", {}))
-                            if isinstance(wf_ctx, dict)
-                            else {}
+                            dict(wf_ctx.get("input", {})) if isinstance(wf_ctx, dict) else {}
                         )
 
                     # Execute sub-workflow
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -580,6 +580,81 @@ class WorkflowEngine:
 
         return await child_engine.run(sub_inputs)
 
+    async def _execute_subworkflow_with_inputs(
+        self,
+        agent: AgentDef,
+        sub_inputs: dict[str, Any],
+    ) -> dict[str, Any]:
+        """Execute a sub-workflow with pre-built inputs.
+
+        Like _execute_subworkflow but accepts explicit inputs instead of
+        extracting them from context. Used by for_each groups where
+        input_mapping has already been rendered with loop variables.
+
+        Args:
+            agent: Workflow agent definition with ``workflow`` path.
+            sub_inputs: Pre-built input dict for the sub-workflow.
+
+        Returns:
+            The sub-workflow's final output dict.
+        """
+        from conductor.config.loader import load_config
+
+        if self._subworkflow_depth >= MAX_SUBWORKFLOW_DEPTH:
+            raise ExecutionError(
+                f"Sub-workflow depth limit exceeded ({MAX_SUBWORKFLOW_DEPTH}). "
+                f"Agent '{agent.name}' cannot invoke sub-workflow '{agent.workflow}'.",
+                suggestion="Check for circular sub-workflow references or reduce nesting depth.",
+            )
+
+        assert agent.workflow is not None  # noqa: S101
+
+        if self.workflow_path is not None:
+            base_dir = Path(self.workflow_path).resolve().parent
+        else:
+            base_dir = Path.cwd()
+
+        sub_path = (base_dir / agent.workflow).resolve()
+
+        if not sub_path.exists():
+            raise ExecutionError(
+                f"Sub-workflow file not found: {sub_path} (referenced by agent '{agent.name}')",
+                suggestion="Check that the 'workflow' path is correct and the file exists.",
+            )
+
+        # Detect circular references via file path
+        current_path = Path(self.workflow_path).resolve() if self.workflow_path else None
+        if current_path is not None and sub_path == current_path:
+            raise ExecutionError(
+                f"Circular sub-workflow reference: agent '{agent.name}' "
+                f"references its own workflow file '{agent.workflow}'.",
+                suggestion="A workflow cannot reference itself as a sub-workflow.",
+            )
+
+        try:
+            sub_config = load_config(sub_path)
+        except Exception as exc:
+            raise ExecutionError(
+                f"Failed to load sub-workflow '{sub_path}' "
+                f"(referenced by agent '{agent.name}'): {exc}",
+                suggestion="Check the sub-workflow YAML for syntax or validation errors.",
+            ) from exc
+
+        child_engine = WorkflowEngine(
+            config=sub_config,
+            provider=self._single_provider,
+            registry=self._registry,
+            skip_gates=self.skip_gates,
+            workflow_path=sub_path,
+            interrupt_event=self._interrupt_event,
+            event_emitter=self._event_emitter,
+            keyboard_listener=self._keyboard_listener,
+            web_dashboard=self._web_dashboard,
+            _subworkflow_depth=self._subworkflow_depth + 1,
+        )
+
+        return await child_engine.run(sub_inputs)
+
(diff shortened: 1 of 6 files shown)


PolyphonyRequiem:
Converted to draft — lint check failing. Schema and type check pass. Needs lint fixes before merging. Depends on PR #109 and PR #110 landing first.

src/conductor/engine/workflow.py
```diff
@@ -2674,7 +2747,52 @@ async def execute_single_item(item: Any, index: int, key: str) -> tuple[str, Any
...
+                    _item_elapsed = _time.time() - _item_start
+
+                    self._emit(
+                        "for_each_item_completed",
+                        {
+                            "group_name": for_each_group.name,
+                            "item_key": key,
+                            "elapsed": _item_elapsed,
+                            "tokens": 0,
+                            "cost_usd": 0.0,
```
jrob5756: **Cost/token tracking lost for nested workflows in for_each.**

`"tokens": 0` and `"cost_usd": 0.0` are hardcoded here. The sub-workflow's nested agent runs accumulate real token usage and cost, but none of it is rolled up into this event. This breaks the dashboard metrics, JSONL logger totals, and any cost reporting for for_each + workflow executions.

`child_engine.run()` should expose aggregate usage (or the engine should already track it on the child). Plumb that through and emit the real values, mirroring what the regular-agent branch below does with `output.tokens_used` and `usage.cost_usd`.

src/conductor/engine/workflow.py
```diff
@@ -2674,7 +2747,52 @@ async def execute_single_item(item: Any, index: int, key: str) -> tuple[str, Any
...
+                # Execute agent — sub-workflow or regular
+                if for_each_group.agent.type == "workflow":
+                    # Build sub-workflow inputs from input_mapping with loop vars
+                    if for_each_group.agent.input_mapping:
+                        renderer = TemplateRenderer()
+                        sub_inputs: dict[str, Any] = {}
+                        for k, tmpl in for_each_group.agent.input_mapping.items():
+                            rendered = renderer.render(tmpl, agent_context)
+                            try:
+                                sub_inputs[k] = json.loads(rendered)
```
jrob5756: **Inconsistent input rendering vs `_execute_subworkflow`.**

This branch attempts `json.loads(rendered)` and falls back to the raw string. But `_execute_subworkflow` (around line 627) does `sub_inputs[key] = rendered` with no JSON parsing — the rendered string is stored as-is.

Result: the **same `input_mapping` produces different types** depending on whether the workflow agent runs as a normal step or inside a for_each. A mapping like `count: "{{ items | length }}"` becomes `int` in for_each but `str` everywhere else. That's a silent type-drift bug users will hit.

Pick one behavior (probably JSON-parse-with-fallback, since it's more useful) and apply it in both methods — ideally by extracting a single `_build_subworkflow_inputs(agent, context)` helper that both call.

jrob5756 (commented):
Merge conflicts and two correctness issues to address before merge. See inline comments.

src/conductor/engine/workflow.py
PolyphonyRequiem: Fixed — `_execute_subworkflow_with_inputs` now returns a tuple and the for_each caller unpacks real usage:

```python
output_content, child_usage = await self._execute_subworkflow_with_inputs(...)

"tokens": child_usage.total_tokens,
"cost_usd": child_usage.total_cost_usd or 0.0,
```

src/conductor/engine/workflow.py
PolyphonyRequiem: yep, extracted `_build_subworkflow_inputs(agent, context)` with json-parse-with-fallback — both `_execute_subworkflow` and the for_each path call it now. no more type drift.

PR #119, opened 2026-04-25
fix(engine): ensure workflow.input available for script/sub-workflow templates in explicit mode

PolyphonyRequiem:
## Problem

In `explicit` context mode, `build_for_agent()` starts with `workflow.input: {}` and only populates entries declared in the agent's `input:` list. Script agents and sub-workflow `input_mapping` templates that reference `{{ workflow.input.X }}` without declaring it in their `input:` list get an empty dict, causing `TemplateError` at runtime.

### Reproduction

```yaml
workflow:
  name: explicit-script
  entry_point: detector
  context:
    mode: explicit
  input:
    work_item_id:
      type: number
      required: true
agents:
  - name: detector
    type: script
    command: pwsh
    args:
      - "-Command"
      - "Write-Output '{{ workflow.input.work_item_id }}'; exit 0"
    routes:
      - to: $end
```

`conductor run explicit-script.yaml --input work_item_id=42`

**Expected:** Script outputs `42`
**Actual:** `TemplateError: 'dict object' has no attribute 'work_item_id'`

### Root Cause

`build_for_agent()` in explicit mode starts with `workflow.input: {}` (line 168 in context.py) and only adds entries via `_add_explicit_input()` for items in the agent's `input:` list. Script agents typically don't declare workflow inputs in their `input:` list because their templates are rendered locally (no LLM prompt cost concern).

### Fix

After `build_for_agent()`, inject the full `workflow_inputs` into the template context for **script** and **workflow** (sub-workflow) agent types. These are local template renders with no LLM cost, so workflow inputs should always be available regardless of context mode.

This preserves the explicit mode contract for **LLM agents** (no unnecessary token cost from undeclared inputs in prompts).

### Changes

- `src/conductor/engine/workflow.py`: Inject full `workflow_inputs` after `build_for_agent()` for script and workflow agent types (2 sites)
- `tests/test_engine/test_workflow.py`: Regression test for explicit mode + script agent template rendering

### Testing

- 75 workflow engine tests pass (including new regression test)
- Existing explicit mode tests unchanged (LLM agents still filter correctly)

### DIFF
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -1399,6 +1399,13 @@ class WorkflowEngine:
                                 agent.input,
                                 mode=self.config.workflow.context.mode,
                             )
+                            # Script args are rendered locally (no LLM cost), so
+                            # workflow inputs must always be available for template
+                            # resolution — even in explicit mode where they'd
+                            # otherwise be filtered out.
+                            agent_context.setdefault("workflow", {})["input"] = (
+                                self.context.workflow_inputs.copy()
+                            )
                             _script_start = _time.time()
 
                             # Count how many times this specific script has been executed
@@ -1490,6 +1497,12 @@ class WorkflowEngine:
                                 agent.name,
                                 agent.input,
                                 mode=self.config.workflow.context.mode,
+                            )
+                            # input_mapping templates are rendered locally (no LLM
+                            # cost), so workflow inputs must always be available —
+                            # even in explicit mode.
+                            agent_context.setdefault("workflow", {})["input"] = (
+                                self.context.workflow_inputs.copy()
                             )
                             _sub_start = _time.time()
 
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -1398,13 +1398,7 @@ class WorkflowEngine:
                                 agent.name,
                                 agent.input,
                                 mode=self.config.workflow.context.mode,
-                            )
-                            # Script args are rendered locally (no LLM cost), so
-                            # workflow inputs must always be available for template
-                            # resolution — even in explicit mode where they'd
-                            # otherwise be filtered out.
-                            agent_context.setdefault("workflow", {})["input"] = (
-                                self.context.workflow_inputs.copy()
+                                agent_type=agent.type,
                             )
                             _script_start = _time.time()
 
@@ -1497,12 +1491,7 @@ class WorkflowEngine:
                                 agent.name,
                                 agent.input,
                                 mode=self.config.workflow.context.mode,
-                            )
-                            # input_mapping templates are rendered locally (no LLM
-                            # cost), so workflow inputs must always be available —
-                            # even in explicit mode.
-                            agent_context.setdefault("workflow", {})["input"] = (
-                                self.context.workflow_inputs.copy()
+                                agent_type=agent.type,
                             )
                             _sub_start = _time.time()
 
@@ -1586,6 +1575,7 @@ class WorkflowEngine:
                             agent.name,
                             agent.input,
                             mode=self.config.workflow.context.mode,
+                            agent_type=agent.type,
                         )
 
                         # Execute agent (get executor for multi-provider support)
@@ -2255,6 +2245,7 @@ class WorkflowEngine:
                     agent.name,
                     agent.input,
                     mode=self.config.workflow.context.mode,
+                    agent_type=agent.type,
                 )
 
                 # Execute agent (get executor for multi-provider support)
@@ -2589,6 +2580,7 @@ class WorkflowEngine:
                     for_each_group.agent.name,
                     for_each_group.agent.input,
                     mode=self.config.workflow.context.mode,
+                    agent_type=for_each_group.agent.type,
                 )
 
                 # Inject loop variables into context
--- a/tests/test_engine/test_workflow.py
+++ b/tests/test_engine/test_workflow.py
@@ -298,6 +298,43 @@ class TestWorkflowEngineContextModes:
         # Workflow.input.goal should not be in agent2's context since it's not in input list
         assert "other" not in agent2_context.get("workflow", {}).get("input", {})
 
+    @pytest.mark.asyncio
+    async def test_explicit_mode_script_gets_workflow_inputs(self) -> None:
+        """Regression: script agents in explicit mode must see workflow.input.
+
+        In explicit mode, workflow.input was empty {} for script agents that
+        didn't declare inputs. Script args are rendered locally (no LLM cost),
+        so workflow inputs must always be available for template resolution.
+        """
+        config = WorkflowConfig(
+            workflow=WorkflowDef(
+                name="explicit-script",
+                entry_point="detector",
+                context=ContextConfig(mode="explicit"),
+            ),
+            agents=[
+                AgentDef(
+                    name="detector",
+                    type="script",
+                    command="pwsh",
+                    args=[
+                        "-Command",
+                        "Write-Output '{{ workflow.input.work_item_id }}'; exit 0",
+                    ],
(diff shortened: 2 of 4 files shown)


tests/test_engine/test_workflow.py
```diff
@@ -298,6 +298,43 @@ def mock_handler(agent, prompt, context):
...
+            workflow=WorkflowDef(
+                name="explicit-script",
+                entry_point="detector",
+                context=ContextConfig(mode="explicit"),
+            ),
+            agents=[
+                AgentDef(
+                    name="detector",
+                    type="script",
+                    command="pwsh",
```
jrob5756: **`pwsh` is not portable.** PowerShell Core isn't installed on standard macOS or many Linux CI runners — this test fails locally with `command not found: 'pwsh'`, even though the templating itself works correctly (you can see `'42'` rendered in captured stderr).

The repo convention for script tests is `command=sys.executable` (see `tests/test_engine/test_script_workflow.py:56`). Suggest:

```python
command=sys.executable,
args=[
    "-c",
    "print('{{ workflow.input.work_item_id }}')",
],
```

The assertion on `stdout.strip() == "42"` still holds.

src/conductor/engine/workflow.py
```diff
@@ -1399,6 +1399,13 @@ async def _execute_loop(self, current_agent_name: str) -> dict[str, Any]:
                                 agent.input,
                                 mode=self.config.workflow.context.mode,
                             )
+                            # Script args are rendered locally (no LLM cost), so
+                            # workflow inputs must always be available for template
+                            # resolution — even in explicit mode where they'd
+                            # otherwise be filtered out.
+                            agent_context.setdefault("workflow", {})["input"] = (
```
jrob5756: **Scope question: why only `workflow.input` and not agent outputs?** The justification in the comment (and PR description) is "script args are rendered locally (no LLM cost), so workflow inputs must always be available." By identical reasoning, prior agent outputs should also always be available — yet a script that references `{{ planner.output.foo }}` without declaring it in `input:` will still hit the same `TemplateError` in explicit mode.

Either:
1. Broaden the fix to give script/sub-workflow agents the full accumulate-mode context (since local renders have no token cost), or
2. Keep the narrow fix but document the intentional limitation (e.g., "workflow inputs are special because they're set once at startup; agent outputs must still be explicitly declared for traceability").

I'd lean toward (1) — the rationale already justifies it.

src/conductor/engine/workflow.py
```diff
@@ -1491,6 +1498,12 @@ async def _execute_loop(self, current_agent_name: str) -> dict[str, Any]:
                                 agent.input,
                                 mode=self.config.workflow.context.mode,
                             )
+                            # input_mapping templates are rendered locally (no LLM
+                            # cost), so workflow inputs must always be available —
+                            # even in explicit mode.
+                            agent_context.setdefault("workflow", {})["input"] = (
```
jrob5756: **Consistency: same fix is missing for parallel/for-each call sites.** There are two other `build_for_agent` call sites in the parallel-group worker (~line 2380) and for-each item (~line 2714) that don't get this treatment.

Today those paths only run LLM agents through `executor.execute()`, so script/workflow agents likely don't reach them — but there's no schema validation forbidding it: parallel groups reference agents by name, and `ForEachDef.agent: AgentDef` accepts any type including `script` and `workflow`.

To prevent drift, consider centralizing this in `build_for_agent` itself (e.g., an `is_local_render: bool` parameter, or a small helper like `_inject_workflow_inputs(ctx)`). Then both current sites — and any future script-in-parallel scenarios — are covered by one place.

jrob5756 (commented):
Awesome work/catch on this one. A couple small comments and we can merge!

tests/test_engine/test_workflow.py
PolyphonyRequiem: Took it as written — switched to `sys.executable` + `python -c` per the convention in `tests/test_engine/test_script_workflow.py`. The `stdout.strip() == "42"` assertion still holds. Pushed in b4b0bbb.

Thanks for the catch, @jrob5756.

src/conductor/engine/workflow.py
PolyphonyRequiem: @jrob5756 your critique reshaped this. You're right that "no LLM cost" doesn't actually carve `workflow.input` away from agent outputs. By that logic a script referencing `{{ planner.output.foo }}` should always render too. That rationale proves too much.

Going with your **option (2)**: narrow fix, rewritten justification. Skipping (1) for now.

**Why narrow.** (1) → (2) is strictly additive. Undeclared `agent.output` references error today, and under (1) they'd just start working. So shipping (2) now and broadening later if `input_mapping` users hit real friction costs us nothing on the compatibility side. The reverse is not true. Ship (1), discover it was wrong, and now we're walking back actual behavior.

**Why the new rationale.** The updated comment drops the LLM-cost framing entirely. The carve-out is now justified on interface grounds: `workflow.input` is the workflow's external interface. Set once at startup, finite, present for the lifetime of the run. Per-step agent outputs are intentionally gated by `input:` for traceability and to keep us from silently rendering templates against data that isn't there yet. That's a real joint to carve at, and it doesn't drag outputs along with it. The new `_LOCAL_RENDER_AGENT_TYPES` constant in `context.py` carries that reasoning in its docstring.

**The bet.** Sub-workflow `input_mapping` is 3 days old. `type: workflow` is ~17 days old. Installed base is basically zero, so whatever lands this week becomes the de facto contract. The user-visible cost under (2): in explicit mode you have to mirror `agent.output` references from `input_mapping` templates into the parent agent's `input:` list. If that turns into a real papercut, broadening to (1) is a one-line change to what `_LOCAL_RENDER_AGENT_TYPES` gates. The centralization from the sibling thread keeps that cheap.

Want your read before you merge. Are you comfortable with (2) as the durable contract, or would you rather we just commit to (1) now while adoption is warm enough that we can set the precedent without breaking anyone? Happy to flip.

b4b0bbb.

src/conductor/engine/workflow.py
PolyphonyRequiem: @jrob5756 done — pushed the centralization to `build_for_agent` itself.

Shape of the change:

- New `agent_type: str | None = None` parameter on `WorkflowContext.build_for_agent`.
- New module-level `_LOCAL_RENDER_AGENT_TYPES = frozenset({"script", "workflow"})` constant in `context.py` carrying the rationale in its docstring.
- All five `build_for_agent` call sites now pass `agent_type=agent.type`: LLM dispatch (~L1575), script (~L1395), sub-workflow (~L1487), parallel-group worker (~L2245), for-each item (~L2580). The two duplicated `setdefault("workflow", {})["input"] = ...` blocks in `_execute_loop` are gone.
- Future call sites get the right behavior by default if they pass `agent.type` along.

Added five direct unit tests on `build_for_agent` in `test_context.py` to lock the contract: script/workflow types get full `workflow.input`; agent outputs stay filtered for them (the deliberate narrow-scope guard from the sibling thread); LLM agents (`agent_type=None`) and `human_gate` are unchanged.

b4b0bbb.

src/conductor/engine/workflow.py
jrob5756: Ok, thanks for digging in and calling this out. Makes sense to me. Lets go with option 2 for now. Thanks!

jrob5756 (approved):
All three review comments addressed cleanly:

1. **pwsh portability** — fixed by using `sys.executable`, matching repo convention
2. **Scope** — documented intentional limitation; `_LOCAL_RENDER_AGENT_TYPES` constant has clear rationale
3. **Consistency** — centralized in `build_for_agent` via new `agent_type` parameter, applied at all 5 call sites

Bonus: 5 new unit tests covering both carve-out behavior and regression guards. CI green across the board. LGTM 🚀

PR #121, opened 2026-04-25
feat(engine): add workflow.dir, workflow.file, workflow.name template variables

PolyphonyRequiem:
## Summary

Adds three new template variables available in all agent contexts:
- `{{ workflow.dir }}` — absolute path to the workflow YAML's parent directory
- `{{ workflow.file }}` — absolute path to the workflow YAML file itself
- `{{ workflow.name }}` — workflow name from the YAML config

## Motivation

Script agents specify script paths in `args`, but these resolve relative to CWD (where `conductor run` is invoked). When workflows live in a **registry** (e.g., `~/.conductor/registries/twig/`), co-located scripts can't be found because CWD is the user's project directory, not the registry.

`{{ workflow.dir }}` solves this by letting scripts build paths relative to the workflow file:

```yaml
agents:
  - name: detector
    type: script
    command: pwsh
    args:
      - "-File"
      - "{{ workflow.dir }}/scripts/detect-state.ps1"
```

This matches how `!file` tags already resolve prompt file paths relative to the workflow file.

## Changes

- `src/conductor/engine/context.py`: Add `workflow_dir`, `workflow_file`, `workflow_name` fields to `WorkflowContext`. Include them in `build_for_agent()` output under the `workflow` dict (all context modes).
- `src/conductor/engine/workflow.py`: Set the new fields from `workflow_path` during `WorkflowEngine.__init__()`.
- `tests/test_engine/test_context.py`: Tests for metadata in accumulate/explicit modes, and empty-string omission.

## Design decisions

- **Available in all context modes** — workflow metadata is lightweight (3 strings) and fundamental. Unlike agent outputs, these shouldn't be filtered by explicit mode.
- **Empty strings omitted** — when `workflow_path` is None (e.g., tests), the fields are empty strings and excluded from the context dict to avoid polluting templates.
- **No breaking changes** — existing `workflow.input` behavior is unchanged. New fields are additive.

## Testing

- 139 context + workflow engine tests pass
- New tests cover accumulate mode, explicit mode, and empty-metadata omission


src/conductor/engine/context.py
```diff
@@ -78,6 +78,18 @@ class WorkflowContext:
...
+    workflow_dir: str = ""
+    """Directory containing the workflow YAML file (resolved absolute path).
+    Available in templates as ``{{ workflow.dir }}``."""
+
+    workflow_file: str = ""
+    """Absolute path to the workflow YAML file.
+    Available in templates as ``{{ workflow.file }}``."""
+
+    workflow_name: str = ""
+    """Name of the workflow from the YAML config.
```
jrob5756, line 98: **Resume drops these fields.**

The new `workflow_dir/file/name` fields aren't included in `to_dict()` (line 486) or restored by `from_dict()` (line 502). On resume, `cli/run.py:1538` calls `engine.set_context(restored_context)`, which **replaces** the engine's context (built with the metadata in `__init__`) with one that has the metadata wiped to empty strings. After resume, `{{ workflow.dir/file/name }}` will silently disappear from templates — exactly the registry-based script-path scenario this PR exists for.

Recommended fix: repopulate metadata inside `WorkflowEngine.set_context()` from `self.workflow_path` / `self.config`. Keeps absolute paths out of checkpoint files (which would otherwise become stale if the workflow moves) and keeps the source of truth co-located with the engine that knows the path.

Please also add a regression test in `tests/test_engine/test_resume.py` (or similar) asserting that `workflow.dir` survives a checkpoint round-trip + resume.

tests/test_engine/test_context.py
```diff
@@ -151,6 +154,50 @@ def test_last_only_mode_empty_history(self) -> None:
...
+
+    def test_empty_metadata_omitted(self) -> None:
+        """Test that empty workflow metadata fields are not included."""
+        ctx = WorkflowContext()
+
+        agent_ctx = ctx.build_for_agent("agent", [], mode="accumulate")
+
+        assert "dir" not in agent_ctx["workflow"]
+        assert "file" not in agent_ctx["workflow"]
+        assert "name" not in agent_ctx["workflow"]
```
jrob5756, line 198: **Suggestion: add an engine-level wiring test.**

These tests verify `WorkflowContext.build_for_agent()` formatting given pre-set fields, but nothing asserts that `WorkflowEngine.__init__(config, workflow_path=...)` actually wires the path through into those fields (the change at `workflow.py:311-315`). A small test in `tests/test_engine/test_workflow.py` like:

```python
def test_engine_populates_workflow_metadata(tmp_path):
    wf_file = tmp_path / "wf.yaml"
    wf_file.write_text("...")  # or use a fixture config
    engine = WorkflowEngine(config, workflow_path=wf_file)
    assert engine.context.workflow_dir == str(tmp_path.resolve())
    assert engine.context.workflow_file == str(wf_file.resolve())
    assert engine.context.workflow_name == config.workflow.name
```

would catch any future regression in the wiring (e.g., if someone refactors `__init__` and reverts to the old `WorkflowContext()`).

jrob5756 (commented):
Merge conflict and some small comments. Thanks!

src/conductor/engine/context.py
PolyphonyRequiem: Yeah, you're right, the round trip drops them and `set_context` was happily clobbering the engine's view. Fixed in `eaf6c9f`: `set_context` now repopulates `workflow_dir/file/name` from `self.workflow_path` and `self.config` after the replace, so the engine stays the source of truth and we keep absolute paths out of the checkpoint. Added `test_set_context_repopulates_workflow_metadata` that round-trips through `to_dict`/`from_dict`, calls `set_context`, and verifies via `build_for_agent` that templates actually resolve. Also covered the no-`workflow_path` case in `test_set_context_without_workflow_path_still_sets_name`.

tests/test_engine/test_context.py
PolyphonyRequiem: Fair, there was no guard that `__init__` was actually wiring the path into the context. Added `test_engine_populates_workflow_metadata` and `test_engine_workflow_metadata_empty_without_path` in `tests/test_engine/test_workflow.py` to lock that down. Same commit, `eaf6c9f`.

PR #122, opened 2026-04-25
feat(engine): auto-parse script agent JSON stdout into output fields

PolyphonyRequiem:
## Summary

When a script agent's stdout is valid JSON, the parsed fields are now merged into the output dict alongside `stdout`/`stderr`/`exit_code`. This makes parsed fields accessible in route conditions and downstream templates as `output.field_name` — matching LLM structured output behavior.

## Motivation

Script agents are the deterministic workhorses of conductor workflows (P8). They emit structured data via JSON stdout, but currently that data is only accessible as a raw string (`output.stdout`). Routing on specific fields requires opaque exit codes instead of readable field-based conditions.

**Before:**
```yaml
# Must use exit codes for routing
routes:
  - to: planner
    when: "exit_code == 20"  # What does 20 mean?
```

**After:**
```yaml
# Route on parsed JSON fields directly
routes:
  - to: planner
    when: "route == 'planning'"  # Self-documenting
```

## Changes

- `src/conductor/engine/workflow.py`: After capturing script output, attempt `json.loads(stdout)`. If it parses as a dict, merge fields into `output_content` (7 lines).
- `tests/test_engine/test_workflow.py`: End-to-end test verifying JSON parsing, field access, and field-based routing.

## Design decisions

- **Backward compatible** — `stdout`, `stderr`, `exit_code` always present. Non-JSON stdout is unchanged. JSON arrays/scalars are ignored (only objects merged).
- **No schema changes** — parsed fields are dynamic, not declared. This matches how script output is already untyped.
- **Merge semantics** — parsed fields could theoretically shadow `stdout`/`stderr`/`exit_code` if a script outputs those as JSON keys. This is intentional: scripts that output `{"exit_code": 42}` probably want the JSON value to win over the raw field.

## Testing

- 75 workflow engine tests pass (including new JSON parsing test)
- Test covers: field parsing, backward compat (stdout still present), field-based routing

### DIFF
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -1455,6 +1455,13 @@ class WorkflowEngine:
                             try:
                                 parsed = json.loads(script_output.stdout)
                                 if isinstance(parsed, dict):
+                                    shadowed = set(parsed.keys()) & set(output_content.keys())
+                                    if shadowed:
+                                        logger.debug(
+                                            "Script '%s' JSON output shadows built-in fields: %s",
+                                            agent.name,
+                                            ", ".join(sorted(shadowed)),
+                                        )
                                     output_content.update(parsed)
                             except (json.JSONDecodeError, ValueError):
                                 pass
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -1448,6 +1448,16 @@ class WorkflowEngine:
                                 "stderr": script_output.stderr,
                                 "exit_code": script_output.exit_code,
                             }
+                            # Auto-parse JSON stdout: if stdout is valid JSON
+                            # object, merge its fields into output so they're
+                            # accessible as output.field_name in templates and
+                            # route conditions (like LLM structured outputs).
+                            try:
+                                parsed = json.loads(script_output.stdout)
+                                if isinstance(parsed, dict):
+                                    output_content.update(parsed)
+                            except (json.JSONDecodeError, ValueError):
+                                pass
                             self.context.store(agent.name, output_content)
                             self.limits.record_execution(agent.name)
                             self.limits.check_timeout()
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -1463,7 +1463,7 @@ class WorkflowEngine:
                                             ", ".join(sorted(shadowed)),
                                         )
                                     output_content.update(parsed)
-                            except (json.JSONDecodeError, ValueError):
+                            except json.JSONDecodeError:
                                 pass
                             self.context.store(agent.name, output_content)
                             self.limits.record_execution(agent.name)
--- a/tests/test_engine/test_workflow.py
+++ b/tests/test_engine/test_workflow.py
@@ -298,6 +298,46 @@ class TestWorkflowEngineContextModes:
         # Workflow.input.goal should not be in agent2's context since it's not in input list
         assert "other" not in agent2_context.get("workflow", {}).get("input", {})
 
+    @pytest.mark.asyncio
+    async def test_script_json_stdout_parsed_into_output(self) -> None:
+        """Test that script stdout containing JSON is auto-parsed into output fields."""
+        config = WorkflowConfig(
+            workflow=WorkflowDef(name="json-script", entry_point="detector"),
+            agents=[
+                AgentDef(
+                    name="detector",
+                    type="script",
+                    command="pwsh",
+                    args=[
+                        "-Command",
+                        'Write-Output \'{"plan_exists": true, "route": "planning"}\'; exit 0',
+                    ],
+                    routes=[
+                        RouteDef(to="planner", when="route == 'planning'"),
+                        RouteDef(to="$end"),
+                    ],
+                ),
+                AgentDef(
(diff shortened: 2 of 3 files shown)


tests/test_engine/test_workflow.py
```diff
@@ -298,6 +298,46 @@ def mock_handler(agent, prompt, context):
...
         assert "other" not in agent2_context.get("workflow", {}).get("input", {})
 
+    @pytest.mark.asyncio
+    async def test_script_json_stdout_parsed_into_output(self) -> None:
+        """Test that script stdout containing JSON is auto-parsed into output fields."""
+        config = WorkflowConfig(
+            workflow=WorkflowDef(name="json-script", entry_point="detector"),
+            agents=[
+                AgentDef(
+                    name="detector",
```
jrob5756: **Portability.** Similar to #119, `pwsh` is not installed by default on macOS or minimal Linux. Running `uv run pytest -k json_stdout` locally on macOS fails with:

```
ExecutionError: Script 'detector': command not found: 'pwsh'
```

CI happens to pass because `ubuntu-latest` pre-installs PowerShell, but this breaks the local dev loop. Every other script test in this repo uses `sys.executable` for portability (see `tests/test_engine/test_script_workflow.py:55,89`).

Suggested fix:

```python
import sys
...
command=sys.executable,
args=[
    "-c",
    'import json; print(json.dumps({"plan_exists": True, "route": "planning"}))',
],
```

And similarly for the `planner` script.

tests/test_engine/test_workflow.py
```diff
@@ -298,6 +298,46 @@ def mock_handler(agent, prompt, context):
         # Workflow.input.goal should not be in agent2's context since it's not in input list
         assert "other" not in agent2_context.get("workflow", {}).get("input", {})
 
+    @pytest.mark.asyncio
```
jrob5756: **Coverage gaps.** This single test only validates the happy path. Consider adding tests for the documented behaviors so they don't silently regress:

- Non-JSON stdout (e.g. `"hello world"`) — `output.stdout` preserved, no extra fields, no exception.
- JSON array / scalar (`[1,2,3]`, `42`) — ignored, not merged (the `isinstance(parsed, dict)` guard).
- Shadowing (`{"exit_code": "ok"}`) — PR description says JSON value wins; no test asserts this contract.
- Empty stdout — doesn't crash.

These are 4–5 lines each and lock in the design decisions called out in the PR description.

**Minor:** consider moving this test to `tests/test_engine/test_script_workflow.py`, which is the dedicated home for script-step behaviors. It currently sits inside `TestWorkflowEngineContextModes`, which is a bit misleading.

src/conductor/engine/workflow.py
```diff
@@ -1448,6 +1448,23 @@ async def _execute_loop(self, current_agent_name: str) -> dict[str, Any]:
...
+                                if isinstance(parsed, dict):
+                                    shadowed = set(parsed.keys()) & set(output_content.keys())
+                                    if shadowed:
+                                        logger.debug(
+                                            "Script '%s' JSON output shadows built-in fields: %s",
+                                            agent.name,
+                                            ", ".join(sorted(shadowed)),
+                                        )
+                                    output_content.update(parsed)
+                            except (json.JSONDecodeError, ValueError):
```
jrob5756: **Nit:** `json.JSONDecodeError` is a subclass of `ValueError`, so the tuple is redundant. Either of these is equivalent and clearer:

```python
except json.JSONDecodeError:
    pass
```

or just:

```python
except ValueError:
    pass
```

Not a correctness issue, just cleanup.

jrob5756 (commented):
Looks good, just a few small comments

tests/test_engine/test_workflow.py
PolyphonyRequiem: @jrob5756 same call as #119, same fix. Swapped both `pwsh` invocations for `sys.executable` per the convention in `test_script_workflow.py`. Side benefit: gets this test off `ubuntu-latest`-only and unbreaks the macOS local loop. 11109e1.

tests/test_engine/test_workflow.py
PolyphonyRequiem: @jrob5756 took both halves.

Relocation: moved into `tests/test_engine/test_script_workflow.py` under a new `TestScriptJsonStdout` class. The old home was `TestWorkflowEngineContextModes`, which had nothing to do with stdout parsing.

Coverage: you were right, these were design decisions in prose with no tests. Now they're contracts:
- non-JSON stdout preserved as-is, no extras, no exception
- JSON arrays/scalars ignored (parametrized over array / int / string / bool)
- shadowing: `{"exit_code": "ok"}` locks in the "JSON value wins" rule from the PR description
- empty stdout doesn't crash

11109e1.

src/conductor/engine/workflow.py
PolyphonyRequiem: @jrob5756 correct, `JSONDecodeError` is a `ValueError` subclass. Narrowed to `except json.JSONDecodeError:` — more specific, more self-documenting. 11109e1.

jrob5756 (approved):
All review feedback addressed cleanly. Tests pass locally on macOS. LGTM.

PR #125, opened 2026-04-25
feat(validate): catch template reference errors before runtime

PolyphonyRequiem:
## Summary

Enhanced `conductor validate` to catch Jinja2 template reference errors before runtime. Previously, `conductor validate` only checked YAML schema (Pydantic) — stale agent references, missing workflow inputs, and undeclared dependencies only surfaced at runtime after minutes of execution and real API costs.

### Errors (validation fails)
- **Stale agent references**: `{{ old_agent_name.output.field }}` where `old_agent_name` doesn't exist
- **Missing workflow input references**: `{{ workflow.input.missing_field }}` where `missing_field` isn't declared in `input:`
- **Stale references in `prompt`, `system_prompt`, `command`, `args`, `working_dir`, `input_mapping`, parallel-group inputs, and workflow `output:` templates** — all template-bearing fields are scanned

### Warnings (validation passes with notes)
- **Undeclared dependencies in explicit mode**: Agent prompt references `{{ a.output.val }}` but doesn't declare `a.output` in its `input:` list

## Example

```yaml
agents:
  - name: writer
    prompt: |
      Based on {{ old_agent_name.output.findings }}
      Write about {{ workflow.input.nonexistent_field }}
```

```
$ conductor validate stale.yaml
╭─ Validation Failed ─╮
│ - agent 'writer' prompt references unknown agent 'old_agent_name'. Available: researcher, writer
│ - agent 'writer' prompt references unknown workflow input 'nonexistent_field'. Declared inputs: topic
╰──────────────────────╯
```

## Changes

- `src/conductor/config/validator.py`:
  - Added `_extract_template_refs()` — parses each template via `jinja2.Environment().parse()` and walks `nodes.Getattr` chains rooted at the truly-free names returned by `meta.find_undeclared_variables`. Loop variables, `{% set %}` bindings, and macro params are excluded automatically because they don't appear in `find_undeclared_variables`. String literals are `nodes.Const` and never enter the walk.
  - Installed `_TolerantNameMap(dict)` on `_JINJA_ENV.filters` / `.tests` so workflow-specific filters (registered on a different env at render time, e.g. `| json`) don't cause `meta.find_undeclared_variables` to raise `TemplateAssertionError` during validation.
  - Added `_collect_template_strings()` to gather all Jinja2 template strings from an agent. Includes `input_mapping` for sub-workflow agents (added by #109 closing #101) — uses `getattr` so it stays a no-op on branches that haven't merged that schema field yet.
  - Added `_validate_template_references()` with agent name, workflow input, and explicit-mode checks.
  - Extended `INPUT_REF_PATTERN` to accept `{group}.errors`, `{group}.errors.{member}`, and bare `{group}.outputs` (with optional `?` suffix on each).
  - Updated `validate_workflow_config()` signature to accept optional `workflow_path`.
- `src/conductor/cli/validate.py`:
  - **Wired semantic validation** (`validate_workflow_config`) into the CLI `validate` command — it previously only ran Pydantic schema validation.
  - Displays warnings for non-fatal issues.

## Testing

- 351 tests pass (`tests/test_config` + `tests/test_cli/test_validate.py`), no regressions.
- New test classes in `tests/test_config/test_validator.py`:
  - `TestExtractTemplateRefs` — unit tests on the new extractor (loop vars, string literals, `{% set %}`, macros, builtins, malformed templates, unknown filters/tests).
  - `TestInputRefPatternExtensions` — parametrized regex shape tests for the new `.errors` / bare `.outputs` forms, plus negative cases.
  - `TestTemplateReferenceValidation` — end-to-end stale-ref detection in `prompt`, `system_prompt`, and parallel-group inputs, plus regression tests for the false-positive classes flagged in review (Jinja2 loop variables, string literals).
  - `TestExplicitModeWarnings` — warnings vs. no-warning paths.
  - `TestOutputTemplateValidation` — workflow `output:` template stale-ref detection.
  - `TestExamplesRegression` — loops over `examples/*.yaml` to prove no example regresses.
  - `TestInputMappingTemplateCollection` — covers `input_mapping` collection + stale-ref detection via a duck-typed agent (so the test runs regardless of whether this branch has merged the `input_mapping` schema field from #109).
- New `TestSemanticValidationIntegration` class in `tests/test_cli/test_validate.py` — 5 CLI-level tests covering exit codes, warning rendering, and the false-positive case being non-blocking.
- All `examples/*.yaml` validate cleanly.

### DIFF
--- a/src/conductor/cli/validate.py
+++ b/src/conductor/cli/validate.py
@@ -40,7 +40,6 @@ def validate_workflow(
 
     try:
         config = load_config(workflow_path)
-        return True, config
     except ConductorError as e:
         # Display structured error
         display_validation_error(e, workflow_path, output_console)
@@ -55,6 +54,20 @@ def validate_workflow(
             )
         )
         return False, None
+
+    # Semantic validation: cross-field references, template refs, etc.
+    try:
+        from conductor.config.validator import validate_workflow_config
+
+        warnings = validate_workflow_config(config, workflow_path=workflow_path)
+        if warnings:
+            for warning in warnings:
+                output_console.print(f"  [yellow]⚠[/yellow] {warning}")
+    except ConductorError as e:
+        display_validation_error(e, workflow_path, output_console)
+        return False, None
+
+    return True, config
 
 
 def display_validation_error(
--- a/src/conductor/config/validator.py
+++ b/src/conductor/config/validator.py
@@ -8,12 +8,13 @@ tool references.
 from __future__ import annotations
 
 import re
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 from conductor.exceptions import ConfigurationError
 
 if TYPE_CHECKING:
-    from conductor.config.schema import WorkflowConfig
+    from conductor.config.schema import AgentDef, WorkflowConfig
 
 
 # Matches agent/group name references in output templates.
@@ -22,6 +23,18 @@ if TYPE_CHECKING:
 # (which only matches {{ }} blocks with .output singular). This pattern also matches {% %} blocks
 # and .outputs plural for path coverage analysis.
 _OUTPUT_REF_PATTERN = re.compile(r"(?:\{\{|\{%)[^}%]*?(\w+)\.outputs?\b")
+
+# Matches agent/workflow template references in Jinja2 expressions:
+#   {{ agent_name.output.field }}
+#   {{ agent_name.output }}
+#   {% if agent_name.output.field %}
+# Excludes built-in namespaces: workflow, context, item, _index, _key
+_TEMPLATE_REF_PATTERN = re.compile(r"(?:\{\{|\{%)[^}%]*?\b(\w+)\.(?:output|outputs)\b")
+
+# Matches workflow.input.X references in templates
+_WORKFLOW_INPUT_REF_PATTERN = re.compile(r"(?:\{\{|\{%)[^}%]*?\bworkflow\.input\.(\w+)\b")
+
+_BUILTIN_NAMES = frozenset({"workflow", "context", "item", "_index", "_key", "loop"})
 
 # DFS path cap: larger workflows may get partial coverage analysis
 _MAX_ENUMERATED_PATHS = 100
@@ -34,20 +47,25 @@ _MAX_ENUMERATED_PATHS = 100
 INPUT_REF_PATTERN = re.compile(
     r"^(?:"
     r"(?P<agent>[a-zA-Z_][a-zA-Z0-9_]*)\.output(?:\.(?P<field>[a-zA-Z_][a-zA-Z0-9_]*))?|"
-    r"(?P<parallel>[a-zA-Z_][a-zA-Z0-9_]*)\.outputs\.(?P<pg_agent>[a-zA-Z_][a-zA-Z0-9_]*)(?:\.(?P<pg_field>[a-zA-Z_][a-zA-Z0-9_]*))?|"
+    r"(?P<parallel>[a-zA-Z_][a-zA-Z0-9_]*)\.(?:outputs|errors)(?:\.(?P<pg_agent>[a-zA-Z_][a-zA-Z0-9_]*)(?:\.(?P<pg_field>[a-zA-Z_][a-zA-Z0-9_]*))?)?|"
     r"workflow\.input\.(?P<input>[a-zA-Z_][a-zA-Z0-9_]*)"
     r")(?P<optional>\?)?$"
 )
 
 
-def validate_workflow_config(config: WorkflowConfig) -> list[str]:
+def validate_workflow_config(
+    config: WorkflowConfig,
+    workflow_path: Path | None = None,
+) -> list[str]:
     """Perform comprehensive validation of a workflow configuration.
 
     This function performs semantic validation beyond what Pydantic can check,
-    including cross-field references and consistency checks.
+    including cross-field references, consistency checks, and Jinja2 template
+    reference validation.
 
     Args:
         config: The WorkflowConfig to validate.
+        workflow_path: Optional path to the workflow file (for !file resolution).
 
     Returns:
         A list of warning messages (non-fatal issues).
@@ -97,6 +115,7 @@ def validate_workflow_config(config: WorkflowConfig) -> list[str]:
             agent_names,
             parallel_names,
             set(config.workflow.input.keys()),
+            for_each_names,
         )
         errors.extend(input_errors)
         warnings.extend(input_warnings)
@@ -134,6 +153,11 @@ def validate_workflow_config(config: WorkflowConfig) -> list[str]:
 
     # Check output templates against conditional execution paths (warnings only)
     warnings.extend(_validate_output_path_coverage(config))
+
+    # Validate Jinja2 template references across all agents
+    tmpl_errors, tmpl_warnings = _validate_template_references(config, workflow_path)
+    errors.extend(tmpl_errors)
+    warnings.extend(tmpl_warnings)
 
     if errors:
         raise ConfigurationError(
@@ -179,6 +203,7 @@ def _validate_input_references(
     agent_names: set[str],
     parallel_names: set[str],
     workflow_inputs: set[str],
+    for_each_names: set[str] | None = None,
 ) -> tuple[list[str], list[str]]:
     """Validate input reference formats and targets.
 
@@ -188,12 +213,14 @@ def _validate_input_references(
         agent_names: Set of valid agent names.
         parallel_names: Set of valid parallel group names.
         workflow_inputs: Set of valid workflow input parameter names.
+        for_each_names: Set of valid for-each group names.
 
     Returns:
         Tuple of (error messages, warning messages).
     """
     errors: list[str] = []
     warnings: list[str] = []
+    group_names = parallel_names | (for_each_names or set())
 
     for input_ref in inputs:
         match = INPUT_REF_PATTERN.match(input_ref)
@@ -220,9 +247,9 @@ def _validate_input_references(
                     f"Agent '{agent_name}' references unknown agent '{ref_agent}' in input"
                 )
 
-        # Check if referencing parallel group output
+        # Check if referencing parallel/for-each group output
         ref_parallel = match.group("parallel")
-        if ref_parallel and ref_parallel not in parallel_names:
+        if ref_parallel and ref_parallel not in group_names:
             is_optional = match.group("optional") == "?"
             if is_optional:
                 warnings.append(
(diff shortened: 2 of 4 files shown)


src/conductor/config/validator.py
```diff
@@ -23,6 +24,18 @@
 # and .outputs plural for path coverage analysis.
 _OUTPUT_REF_PATTERN = re.compile(r"(?:\{\{|\{%)[^}%]*?(\w+)\.outputs?\b")
 
+# Matches agent/workflow template references in Jinja2 expressions:
+#   {{ agent_name.output.field }}
+#   {{ agent_name.output }}
+#   {% if agent_name.output.field %}
+# Excludes built-in namespaces: workflow, context, item, _index, _key
+_TEMPLATE_REF_PATTERN = re.compile(r"(?:\{\{|\{%)[^}%]*?\b(\w+)\.(?:output|outputs)\b")
```
jrob5756: **Critical: false positives on valid workflows.**

This regex has no awareness of Jinja2 syntax beyond the `{{`/`{%` delimiters. Two problem cases I verified:

1. **Jinja2 `{% for %}` loop variables** — `{% for r in researcher.outputs %}{{ r.output.text }}{% endfor %}` matches `r.output` and reports `r` as 'unknown agent'. This is a normal Jinja2 idiom and these workflows used to validate fine.
2. **String literals inside templates** — `{{ x | replace("foo.output", "") }}` reports `foo` as 'unknown agent'.

Repro:
```python
import re
p = re.compile(r'(?:\{\{|\{%)[^}%]*?\b(\w+)\.(?:output|outputs)\b')
[m.group(1) for m in p.finditer('{% for x in y %}{{ x.output }}{% endfor %}')]
# => ['x']
[m.group(1) for m in p.finditer('{{ x | replace("foo.output", "") }}')]
# => ['foo']
```

Because these become **errors** (not warnings), `conductor validate` will hard-fail on previously-valid workflows. Either:
- Replace with a real Jinja2 AST walk (`jinja2.Environment().parse()` + `jinja2.meta.find_undeclared_variables`), which gives string- and scope-aware variable tracking, or
- Downgrade these to warnings until the AST approach lands.

src/conductor/config/validator.py
```diff
@@ -34,20 +47,25 @@
 INPUT_REF_PATTERN = re.compile(
     r"^(?:"
     r"(?P<agent>[a-zA-Z_][a-zA-Z0-9_]*)\.output(?:\.(?P<field>[a-zA-Z_][a-zA-Z0-9_]*))?|"
-    r"(?P<parallel>[a-zA-Z_][a-zA-Z0-9_]*)\.outputs\.(?P<pg_agent>[a-zA-Z_][a-zA-Z0-9_]*)(?:\.(?P<pg_field>[a-zA-Z_][a-zA-Z0-9_]*))?|"
+    r"(?P<parallel>[a-zA-Z_][a-zA-Z0-9_]*)\.(?:outputs|errors)(?:\.(?P<pg_agent>[a-zA-Z_][a-zA-Z0-9_]*)(?:\.(?P<pg_field>[a-zA-Z_][a-zA-Z0-9_]*))?)?|"
```
jrob5756, line 82: **Needs test coverage for the new accepted forms.**

This pattern was extended to accept `.errors` (in addition to `.outputs`) and to make the trailing `.pg_agent[.pg_field]` optional. Please add tests asserting the new accepted shapes — at minimum:

- `{group}.errors`
- `{group}.errors.{member}`
- `{group}.outputs` (bare, no member)
- Optional `?` suffix on each of the above

Without tests, future refactors can silently regress this regex.

src/conductor/config/validator.py
```diff
@@ -577,3 +604,171 @@ def _validate_output_path_coverage(config: WorkflowConfig) -> list[str]:
...
+
+    if agent.prompt:
+        templates.append((f"agent '{agent.name}' prompt", agent.prompt))
+    if agent.system_prompt:
+        templates.append((f"agent '{agent.name}' system_prompt", agent.system_prompt))
+    if agent.command:
+        templates.append((f"agent '{agent.name}' command", agent.command))
+    for i, arg in enumerate(agent.args):
+        templates.append((f"agent '{agent.name}' args[{i}]", arg))
+    input_mapping: dict[str, str] | None = getattr(agent, "input_mapping", None)
```
jrob5756, line 726: **Dead code.**

`AgentDef` has no `input_mapping` field — confirmed by grep across the entire repo (no occurrences in `src/`, `tests/`, or `examples/`). Pydantic v2 defaults to `extra='ignore'`, so `getattr(agent, 'input_mapping', None)` always returns `None` and this whole branch is unreachable.

The PR description still claims:
> Stale references in script args, **input_mapping**, system_prompt: All template-bearing fields are scanned

Either remove this block or, if forward-looking, gate it behind an actual schema field added in this PR.

src/conductor/config/validator.py
```diff
@@ -577,3 +604,171 @@ def _validate_output_path_coverage(config: WorkflowConfig) -> list[str]:
...
+def _resolve_prompt_file(agent_name: str, prompt: str, workflow_path: Path | None) -> str | None:
+    """If a prompt looks like it came from a !file tag, try to load the file content.
+
+    The !file tag is already resolved during YAML loading, so the prompt field
+    contains the file content. This function is for scanning prompts that may
+    have been loaded from files for additional template references.
+
+    Returns the prompt as-is (it's already resolved by the loader).
+    """
+    return prompt if prompt else None
```
jrob5756: **Dead code — please remove.**

`_resolve_prompt_file` is never called anywhere (verified via grep). The body is a no-op (`return prompt if prompt else None`) and the docstring explicitly says "the prompt is already resolved by the loader." Either delete this entire function, or replace it with the real `!file` resolution it implies (and call it).

src/conductor/config/validator.py
```diff
@@ -577,3 +604,171 @@ def _validate_output_path_coverage(config: WorkflowConfig) -> list[str]:
...
+
+
+def _validate_template_references(
```
jrob5756, line 734: **Suggestion: use Jinja2's own parser instead of regex.**

Jinja2 ships with `jinja2.Environment().parse(source)` and `jinja2.meta.find_undeclared_variables(ast)` — these give you string-aware, comment-aware, and scope-aware (recognizing `{% for x in y %}`, `{% set x = ... %}`, macro params, etc.) variable tracking. This eliminates both classes of false positives flagged on `_TEMPLATE_REF_PATTERN` with no ongoing regex maintenance burden, and naturally handles cases like `{{ workflow.input.x | filter('foo.output') }}`.

Rough sketch:
```python
from jinja2 import Environment, meta

env = Environment()
ast = env.parse(template)
undeclared = meta.find_undeclared_variables(ast)  # set of top-level names
# Then resolve each name to either: builtin, agent, parallel/foreach group, workflow input, loop var
```

Also: `find_referenced_templates(ast)` for `{% include %}`/`{% extends %}` if those become relevant.

If this is too much for this PR, please at least downgrade missing-reference detection from errors to warnings so false positives don't block valid workflows.

src/conductor/cli/validate.py
```diff
@@ -56,6 +55,20 @@ def validate_workflow(
...
+    try:
+        from conductor.config.validator import validate_workflow_config
+
+        warnings = validate_workflow_config(config, workflow_path=workflow_path)
+        if warnings:
+            for warning in warnings:
+                output_console.print(f"  [yellow]⚠[/yellow] {warning}")
+    except ConductorError as e:
+        display_validation_error(e, workflow_path, output_console)
+        return False, None
```
jrob5756, line 68: **Missing tests for the new validation behavior.**

This PR adds ~200 lines of nontrivial regex/AST-style validation logic (in `validator.py`) and wires it into the CLI here, but adds zero new tests. Codecov reports 16 patch lines uncovered.

At minimum, please add:
- **Positive regression tests**: every workflow under `examples/` should still validate successfully (this would catch issues #1 today).
- **Negative tests** for each new error path: stale agent ref, stale workflow input ref, references in `system_prompt`/`command`/`args`/`working_dir`.
- **Explicit-mode warning tests**: undeclared agent ref + undeclared `workflow.input.X` ref in explicit mode.
- **For-each tests**: inline-agent template scanning, `for_each_names` accepted in `_validate_input_references`.
- **Pattern-extension tests**: new `INPUT_REF_PATTERN` accepts `.errors` and bare `{group}.outputs`.
- **CLI integration**: this branch — verify warnings render and that errors return `(False, None)`.

jrob5756 (commented):
Some testing and cleanup

src/conductor/config/validator.py
PolyphonyRequiem: Confirmed both repros locally — good catches.

Replaced the regex with `jinja2.Environment().parse(...)` + `meta.find_undeclared_variables`, then walked `nodes.Getattr` chains rooted at the truly-free names. Both classes of false positive are now structurally impossible:

- The loop variable `r` is bound by `{% for r in ... %}` and so doesn't appear in `find_undeclared_variables` — it's filtered out before we ever look at the chain.
- String literals are `nodes.Const`, not `nodes.Getattr`, so they never enter the walk.

Added regression tests for both in `TestExtractTemplateRefs.test_for_loop_variable_is_not_a_ref` and `test_string_literal_is_not_a_ref`, plus end-to-end versions in `TestTemplateReferenceValidation`.

Fix: 278b389

src/conductor/config/validator.py
PolyphonyRequiem: Added — see `TestInputRefPatternExtensions` in `tests/test_config/test_validator.py`. Parametrized coverage for the new shapes:

- `{group}.errors` and `{group}.errors?`
- `{group}.errors.{member}` and `{group}.errors.{member}.{field}`
- `{group}.outputs` (bare) and `{group}.outputs?`

Plus a legacy-shape test (`agent.output`, `workflow.input.param`, etc.) so the regex can't drift backward, and a negative test that explicitly rejects bare `workflow.input` and other malformed shapes — so a future refactor that loosens the pattern fails a test instead of silently regressing.

Fix: 278b389

src/conductor/config/validator.py
PolyphonyRequiem: Right on both counts — the field doesn't exist on `AgentDef` and the branch is unreachable. Removed the dead block in 278b389.

Will update the PR description to drop the `input_mapping` claim. The forward-looking version belongs with #101 when that lands, not here.

src/conductor/config/validator.py
PolyphonyRequiem: Deleted in 278b389. No callers, no behavior — confirmed via grep. If we want real `!file` resolution at validate time it's a separate concern; the loader already handles it for run.

src/conductor/config/validator.py
PolyphonyRequiem: Done — switched to `Environment().parse(...)` + `meta.find_undeclared_variables` + a `nodes.Getattr` walk. See `_extract_template_refs` in 278b389. Both false-positive classes from the regex are now structurally impossible (loop vars are bound and so excluded from `find_undeclared_variables`; string literals are `nodes.Const` and never enter the walk).

One Jinja2 quirk worth flagging for future maintainers: `meta.find_undeclared_variables` actually runs the code generator over the AST, so it raises `TemplateAssertionError` for unknown filters/tests at *find* time, not just at render time. Conductor registers `| json` and friends on a different env at render time, which would have made every `meta.find_undeclared_variables` call blow up on real workflows. Worked around with a `_TolerantNameMap(dict)` installed onto `_JINJA_ENV.filters` and `.tests` that returns an identity function for any unknown name. Covered in `TestExtractTemplateRefs.test_unknown_filter_does_not_raise` and `test_unknown_test_does_not_raise`.

src/conductor/cli/validate.py
PolyphonyRequiem: Added a substantial test suite in 278b389:

In `tests/test_config/test_validator.py`:
- `TestExtractTemplateRefs` — 12 unit tests on the new extractor (loop vars, string literals, `{% set %}`, macros, builtins, malformed templates, unknown filters/tests).
- `TestInputRefPatternExtensions` — parametrized regex shape tests for the new `.errors` / bare `.outputs` forms (and a negative test for shapes that should still be rejected).
- `TestTemplateReferenceValidation` — end-to-end stale-ref detection in `prompt`, `system_prompt`, and parallel-group inputs, plus regression tests for both false-positive classes.
- `TestExplicitModeWarnings` — warnings vs no-warning paths for `context.mode: explicit`.
- `TestOutputTemplateValidation` — workflow `output:` template stale-ref detection.
- `TestExamplesRegression` — loops over `examples/*.yaml` to prove no example regresses (your suggested positive regression test).

In `tests/test_cli/test_validate.py`:
- `TestSemanticValidationIntegration` — 5 CLI-level tests covering exit codes, warning rendering, and the false-positive case being non-blocking.

Total: ~600 net lines of test coverage. Locally the new tests pull the patch coverage well above the 80% bar; will see what codecov reports on the next run.

src/conductor/config/validator.py
PolyphonyRequiem: Correcting myself — #101 actually closed via #109 on May 1, which added `input_mapping` to `AgentDef`. So the field DOES exist on main, the validator block in main collects from it, and dropping it from this PR would have been a regression once it merges back.

Restored the collection in 8cad8e3, kept the `getattr` shim so the code works on this branch (which is behind main and doesn't have the field yet) and on main (which does). Added `TestInputMappingTemplateCollection` exercising the path via a duck-typed agent — three tests covering the present/absent paths and stale-ref detection inside an `input_mapping` expression.

PR description updated to put `input_mapping` back in the list of scanned fields.

jrob5756 (approved):
LGTM — all 6 review comments addressed substantively. Switch to Jinja2 AST is well-implemented, regression tests cover both false-positive classes, and example workflows still validate.

PR #130, opened 2026-05-01
feat(dialog): agent dialog mode with web dashboard support

PolyphonyRequiem:
## Summary

Add **dialog mode** for agents, enabling multi-turn conversational interactions within workflows.

### Core
- **Dialog gate** (\gates/dialog.py\) — full dialog gate implementation supporting multi-turn conversations
- **Dialog evaluator** (\ngine/dialog_evaluator.py\) — condition evaluation for dialog transitions
- **Schema** — new dialog-related Pydantic models in \config/schema.py\
- **Workflow engine** — dialog execution integration in \ngine/workflow.py\
- **Linkify utility** (\xecutor/linkify.py\) — link rendering for dialog content

### Provider Support
- Both **Copilot** and **Claude** providers updated with dialog support

### Web Dashboard
- New React components: \DialogDetail\, \DialogEngagementPrompt\, \DialogOverlay\
- Updated workflow store, event types, detail panel, and layout
- Rebuilt static assets

### Tests
- \	est_dialog_schema.py\ — schema validation tests
- \	est_dialog_evaluator.py\ — evaluator unit tests  
- \	est_dialog.py\ — dialog gate integration tests

### Docs & Examples
- \workflow-syntax.md\ — dialog mode documentation
- \xamples/dialog-mode.yaml\ and \xamples/dialog-test.yaml\


src/conductor/web/frontend/src/components/detail/DialogDetail.tsx
```diff
@@ -0,0 +1,239 @@
...
+              {children}
+            </pre>
+          ),
+          strong: ({ children }) => <strong className="font-semibold">{children}</strong>,
+          em: ({ children }) => <em className="italic">{children}</em>,
+          a: ({ href, children }) => {
+            if (isRelativeFileLink(href)) {
+              const handleOpenInEditor = (e: React.MouseEvent) => {
+                e.preventDefault();
+                fetch('/api/open-file', {
```
jrob5756: **Missing endpoint** — This POSTs to `/api/open-file`, but no such route is registered in `src/conductor/web/server.py` (`grep open-file` returns nothing). Clicks will silently 404.

Either:
1. Implement the endpoint with strict validation (reject `..`, resolve symlinks, enforce containment within the workflow directory), OR
2. Remove the link rewriting in `DialogMarkdown` and render plain anchors.

Additionally, even with the endpoint, `isRelativeFileLink` (lines 13–19) accepts paths like `../../../../etc/passwd` — server-side validation is mandatory.

src/conductor/executor/linkify.py
```diff
@@ -0,0 +1,263 @@
...
+    # Must not look like a URL (already handled)
+    if re.match(r"https?://", stripped):
+        return None
+
+    # If base_dir is provided, verify file exists
+    if base_dir is not None:
+        try:
+            candidate = (base_dir / stripped).resolve()
+            # Security: must be within base_dir
+            if not str(candidate).startswith(str(base_dir.resolve())):
```
jrob5756: **Path traversal** — `str(candidate).startswith(str(base_dir.resolve()))` is a string-prefix check, not a path-aware containment check.

Example bypass: `base_dir` resolves to `/foo/bar`; a candidate resolving to `/foo/barbaz/x.md` passes the check (`'/foo/barbaz/x.md'.startswith('/foo/bar')` is `True`).

**Fix** (Python 3.9+):
```python
try:
    candidate.relative_to(base_dir.resolve())
except ValueError:
    return None
```
or `candidate.is_relative_to(base_dir.resolve())`.

src/conductor/gates/dialog.py
```diff
@@ -0,0 +1,705 @@
...
+            )
+
+            # Check if user is dismissing the dialog
+            if self._is_dismiss(user_input):
+                result.user_dismissed = True
+                self._display_dialog_end(dismissed_by="user")
+                break
+
+            # Send to agent and get response
+            history.append({"role": "user", "content": user_input})
```
jrob5756, line 292: **History corruption on exception (CLI mode)** — User input is appended to `history` *before* `execute_dialog_turn`. On exception, the handler at line 280–289 logs and `continue`s without rolling back. The next iteration appends another user message, leaving two consecutive `role=user` entries with no agent response between them — confusing the LLM on retry.

**Fix**: append after successful response, or explicitly `history.pop()` in the exception handler.

src/conductor/gates/dialog.py
```diff
@@ -0,0 +1,705 @@
...
+                    "turn_count": len(result.messages),
+                    "user_dismissed": True,
+                },
+            )
+            return result
+
+        # Dialog loop
+        while True:
+            # Send to agent and get response
+            history.append({"role": "user", "content": user_input})
```
jrob5756, line 455: **Same history corruption pattern in web mode** — Same issue as the CLI loop above. Line 431 appends user_input before the try, but on exception the new user_input from `wait_for_dialog_message` (line 460) is appended again at the next iteration's line 431. The failed user message stays in history forever.

**Fix**: pop the failed entry in the exception handler before waiting for the next message, or restructure to append after success.

src/conductor/gates/dialog.py
```diff
@@ -0,0 +1,705 @@
...
+                {
+                    "dialog_id": dialog_id,
+                    "agent_name": agent.name,
+                    "role": "agent",
+                    "content": agent_response,
+                },
+            )
+
+            # Check if agent proposed completion
+            if "[READY_TO_CONTINUE]" in agent_response:
```
jrob5756: **Suggestion: `[READY_TO_CONTINUE]` is user-injectable** — A user could include the literal string in their message; if the agent quotes it back, the dialog ends prematurely. Consider:
- Stripping the marker from user input before sending to the agent, OR
- Requiring it to be at start/end of the response, not anywhere in it.

src/conductor/engine/dialog_evaluator.py
```diff
@@ -0,0 +1,167 @@
...
+            output_str = json.dumps(output, indent=2, default=str)
+        except (TypeError, ValueError):
+            output_str = str(output)
+
+        system_prompt = EVALUATOR_SYSTEM_PROMPT.format(
+            trigger_prompt=agent.dialog.trigger_prompt,
+        )
+        user_prompt = EVALUATOR_USER_PROMPT.format(
+            agent_name=agent.name,
+            agent_output=output_str[:4000],  # Truncate to avoid excessive tokens
```
jrob5756: **Silent JSON truncation** — `output_str[:4000]` can split a large JSON output mid-structure, leaving malformed context for the evaluator. The `try/except` at line 121 swallows the error and the evaluator returns no trigger, hiding the issue.

Consider appending a `"\n…[truncated]"` marker so the evaluator LLM understands it has partial data, and add a test exercising 5000+ char outputs.

src/conductor/engine/dialog_evaluator.py
```diff
@@ -0,0 +1,167 @@
...
+
+        Returns:
+            Parsed DialogEvaluation.
+        """
+        try:
+            text = response.strip()
+            # Handle markdown code blocks
+            if text.startswith("```"):
+                lines = text.split("\n")
+                text = "\n".join(lines[1:-1]) if len(lines) > 2 else text
```
jrob5756: **Code-block parser strips valid content on missing fence** — `lines[1:-1]` assumes the response ends with a closing ```` ``` ````. If the LLM omits it, the last line of valid JSON is sliced off and parsing fails.

```python
if text.startswith("```"):
    lines = text.split("\n")
    if len(lines) > 2 and lines[-1].strip().startswith("```"):
        text = "\n".join(lines[1:-1])
    elif len(lines) > 1:
        text = "\n".join(lines[1:])
```

src/conductor/web/frontend/src/stores/workflow-store.ts
```diff
@@ -236,6 +244,12 @@ interface WorkflowState {
   _wsSend: ((data: object) => void) | null;
   setWsSend: (fn: ((data: object) => void) | null) => void;
   sendGateResponse: (agentName: string, selectedValue: string, additionalInput?: Record<string, string>) => void;
+  // Dialog state
+  activeDialog: { agentName: string; dialogId: string } | null;
+  dialogEngaged: boolean;
```
jrob5756, line 332: **`replayState` doesn't reset these new fields** — `activeDialog` and `dialogEngaged` are declared here but `replayState` (around line 395, outside this diff) doesn't reset them in its explicit reset block. The spread `...state` carries them over from the prior workflow, so late-joiners or re-replays can show a stale dialog overlay.

**Fix**: add to the reset object in `replayState`:
```ts
activeDialog: null,
dialogEngaged: false,
```

src/conductor/web/frontend/src/stores/workflow-store.ts
```diff
@@ -321,6 +335,44 @@ export const useWorkflowStore = create<WorkflowState>((set) => ({
...
+        type: 'dialog_message',
+        agent_name: agentName,
+        dialog_id: dialogId,
+        content: message,
+      });
+      // Optimistically mark awaiting response
+      set((state) => {
+        const nodes = { ...state.nodes };
+        if (nodes[agentName]) {
+          nodes[agentName] = { ...nodes[agentName]!, dialog_awaiting_response: true };
```
jrob5756: **Optimistic UI race** — The optimistic `dialog_awaiting_response: true` is a separate `set()` from the eventual `dialog_message` handler that flips it back to `false`. If the agent response arrives before this update is committed, the UI can stick in 'awaiting' until the next message. Low likelihood, but easily fixed by either:
- Removing the optimistic update (let the echo from the server set it), OR
- Using a single atomic `set()` keyed off the send action.

tests/test_gates/test_dialog.py
```diff
@@ -0,0 +1,285 @@
...
+    """Tests for DialogResult dataclass."""
+
+    def test_default_values(self) -> None:
+        """Test DialogResult has sensible defaults."""
+        result = DialogResult(dialog_id="test-123")
+        assert result.dialog_id == "test-123"
+        assert result.messages == []
+        assert result.user_dismissed is False
+        assert result.user_declined is False
+        assert result.agent_proposed_continue is False
```
jrob5756, line 425: **Zero coverage for web dialog flow** — `_web_handle_dialog()` in `gates/dialog.py:337-557` is 220 lines with no tests. Missing coverage:
- `wait_for_dialog_message` happy path
- `dialog_decline` message handling (lines 386, 457, 495, 527)
- Web-mode agent approval flow (lines 494–513)
- Exception/retry path (lines 439–474)

All existing tests use CLI mode with `web_dashboard=None`. Please add a `test_web_dialog.py` (or extend this file) mocking `WebDashboard.wait_for_dialog_message`.

tests/test_engine/test_dialog_evaluator.py
```diff
@@ -0,0 +1,151 @@
...
+        provider.execute_dialog_turn = AsyncMock(
+            return_value='{"trigger": false, "reason": "nope"}'
+        )
+
+        await self.evaluator.evaluate(agent, {"result": "hello"}, provider)
+
+        # Verify the system prompt contains the trigger criteria
+        call_args = provider.execute_dialog_turn.call_args
+        assert "Trigger when confused" in call_args.kwargs["system_prompt"]
+        assert "my_agent" in call_args.kwargs["user_message"]
```
jrob5756, line 206: **No integration test for dialog → re-execute** — `_handle_dialog()` in `workflow.py:561-638` formats the dialog transcript into re-execution guidance and re-runs the agent. The end-to-end flow (evaluate → dialog → re-execute with transcript appended → return new output) is not exercised — existing tests mock the DialogHandler.

Please add an integration test that runs a real workflow with dialog enabled and asserts the agent re-executes with the transcript present in its guidance/context.

src/conductor/providers/copilot.py
```diff
@@ -1709,6 +1709,97 @@ async def validate_connection(self) -> bool:
                 is_retryable=False,
             ) from e
 
+    async def execute_dialog_turn(
```
jrob5756, line 1778: **No tests for `execute_dialog_turn` on either provider** — Both providers gained ~45 lines of new code with different message-format strategies (Copilot builds a `User: …\n\nAssistant: …` string, Claude uses a message list). Per AGENTS.md provider parity rules, these need parallel test coverage.

Please add tests in `tests/test_providers/test_copilot.py` and `tests/test_providers/test_claude.py` covering: empty history, multi-turn history, model override, error propagation, and equivalent output for the same logical input.

jrob5756 (commented):
Cool change. A couple of recommendations to consider before merging.

src/conductor/web/frontend/src/components/detail/DialogDetail.tsx
PolyphonyRequiem: the missing endpoint is from another PR (the metadata enrichment on human gates)

src/conductor/executor/linkify.py
PolyphonyRequiem: fixed in 889bb65 — switched to `Path.is_relative_to` (3.12+) against the resolved base. drops the string-prefix gotcha entirely.

src/conductor/gates/dialog.py
PolyphonyRequiem: fixed in 889bb65. CLI loop now pops the user turn from `history` in the exception branch before `continue`. `result.messages` (the user-visible transcript) is left intact — orphan user msg without an agent reply is the honest representation of what happened. added `test_cli_exception_pops_user_history` covering the success/fail/success sequence.

src/conductor/gates/dialog.py
PolyphonyRequiem: fixed in 889bb65. same pop-on-exception in the web `_web_handle_dialog` loop. also caught a related bug while in there — the READY-decline branch was double-appending `approval` to `history` (once explicitly, once at the loop top), creating two consecutive user turns even on the happy path. dropped the explicit append. added `TestWebDialogFlow` with 5 tests covering decline, happy path, exception recovery, READY-decline-no-duplicate-history, and READY-approval.

src/conductor/gates/dialog.py
PolyphonyRequiem: fixed in 889bb65. `_extract_ready_marker` now uses `rstrip().endswith()` so the marker is only treated as terminal when it's actually at the end. mid-message marker is left as literal text, agent's reply is stored cleanly in `result.messages` with the marker stripped. `TestReadyMarkerTerminalOnly` covers both cases.

src/conductor/engine/dialog_evaluator.py
PolyphonyRequiem: fixed in 889bb65. added `_truncate_for_evaluator(text, limit)` that appends `\n…[truncated]` so the evaluator (and any human inspecting the prompt) knows context was cut. `TestEvaluatorTruncation` has the 5500-char round-trip test.

src/conductor/engine/dialog_evaluator.py
PolyphonyRequiem: fixed in 889bb65. parser now uses `splitlines()` and only strips the last line when it's actually a closing fence. unterminated fences keep the JSON intact. tests: `test_parse_unterminated_code_fence_keeps_last_line` and `test_parse_terminated_code_fence_strips_both`.

src/conductor/web/frontend/src/stores/workflow-store.ts
PolyphonyRequiem: fixed in 889bb65. all three replay-state mutators (`replayState`, `setReplayMode`, `setReplayPosition`) now reset `activeDialog: null, dialogEngaged: false`. you were right that `setReplayPosition` was the same hazard — easy to scrub through replay and end up with a stale dialog panel.

src/conductor/web/frontend/src/stores/workflow-store.ts
PolyphonyRequiem: adopted in 889bb65 — dropped the optimistic set in `sendDialogMessage`. the `dialog_message` event handler is now the single source of truth: sets `dialog_awaiting_response = true` on `role === 'user'` echo, `false` on `role === 'agent'`. server round-trip is fast enough that the perceived snappiness loss is negligible, and the race is just gone.

tests/test_gates/test_dialog.py
PolyphonyRequiem: fixed in 889bb65. added `TestWebDialogFlow` in `test_dialog.py` with 5 tests covering the web flow end-to-end: engagement decline, happy path, provider exception → recovery (no orphan user in history), READY-decline (asserts no duplicate history append), and READY-approval. plus a `wait_for_dialog_message` fix in `web/server.py` to filter on `dialog_id` not just `agent_name` — caught while writing the tests.

tests/test_engine/test_dialog_evaluator.py
PolyphonyRequiem: fixed in 889bb65. new `tests/test_engine/test_dialog_integration.py` runs the real `WorkflowEngine` end-to-end with a `CopilotProvider(mock_handler=...)` plus mocked `execute_dialog_turn` and patched dialog-handler interactive prompts. asserts (a) the agent runs twice, (b) the second run's prompt contains `--- DIALOG WITH USER ---`, `--- END DIALOG ---`, and the user's actual message text.

src/conductor/providers/copilot.py
PolyphonyRequiem: fixed in 889bb65. added 4 tests each to `test_claude.py` and `test_copilot.py` (`TestClaudeExecuteDialogTurn` / `TestCopilotExecuteDialogTurn`) covering empty history, multi-turn history serialization, model override, and error wrapping. for Copilot the test mocks the `create_session` + event-driven flow and asserts the prompt is built correctly with `User:` / `Assistant:` blocks.

src/conductor/web/frontend/src/components/detail/DialogDetail.tsx
PolyphonyRequiem: follow-up: also tightened `isRelativeFileLink` in 889bb65 to reject `..` segments and Windows drive letters (`C:\path`) as defense-in-depth. when the `/api/open-file` endpoint lands from the metadata PR, the client side won't be the only line of defense.

PolyphonyRequiem:
review pass complete in 889bb65. summary by category:

**real bugs (fixed):**
- `linkify.py` path traversal: `Path.is_relative_to` instead of `str.startswith`
- `dialog.py` CLI + web: pop user turn from provider history on exception (no more orphan-user → consecutive-user-turn corruption)
- `dialog.py` web READY-decline: dropped duplicate `history.append` (loop top already does it)
- `dialog_evaluator.py` code-fence parser: only strip last line if it's an actual closing fence
- `workflow-store.ts` replay: `setReplayPosition` now resets dialog state too (was missing)
- `web/server.py` `wait_for_dialog_message`: filter on `dialog_id` not just `agent_name` (caught while writing tests)

**suggestions adopted:**
- READY_TO_CONTINUE: terminal-only via `rstrip().endswith()`, marker stripped from stored transcript
- evaluator truncation: append `\n…[truncated]` marker so the evaluator knows context was cut
- optimistic UI in `sendDialogMessage`: dropped; `dialog_message` event handler is now single source of truth for `dialog_awaiting_response`
- `DialogDetail.tsx`: defense-in-depth — `isRelativeFileLink` rejects `..` and Windows drive letters

**tests added:**
- `TestWebDialogFlow` (5): engagement decline, happy path, exception recovery, READY-decline-no-duplicate-history, READY-approval
- `TestDialogHandlerExceptionRecovery`: 3-call provider sequence asserting no consecutive user turns
- `TestReadyMarkerTerminalOnly`: mid-message vs terminal marker
- `TestEvaluatorTruncation`: end-to-end 5500-char round-trip
- `test_dialog_integration.py`: full `WorkflowEngine` end-to-end with dialog → re-execute, asserts transcript appears in re-execution guidance
- `TestClaudeExecuteDialogTurn` + `TestCopilotExecuteDialogTurn` (4 each): empty/multi-turn history, model override, error wrapping
- code-fence parser: terminated + unterminated cases

169 dialog/evaluator/provider/gate tests pass. ruff clean. ty has only pre-existing termios platform diagnostics.

note: TS source only — frontend static assets need a CI rebuild before this is mergeable.

thanks for the thorough pass — the history-corruption catches were genuinely subtle and the test-coverage call-outs forced the integration test that I'd been deferring.

src/conductor/web/frontend/src/components/detail/DialogDetail.tsx
jrob5756: It looks like this was in #114 but that PR was closed

jrob5756 (approved):
All 12 review comments addressed. Tests pass (157), lint/format clean, merged with latest main. LGTM 🚀

PR #131, opened 2026-05-01
feat(gates): markdown rendering and auto-linkification in human gate prompts

PolyphonyRequiem:
## Summary

Adds full Markdown rendering support to human gate prompts in both the terminal CLI and the web dashboard.

### Rich Terminal Rendering
- Gate prompts are now rendered through Rich Markdown in the terminal, supporting headings, bold, lists, code blocks, and tables

### Auto-Linkification (\linkify.py\)
- New markdown-aware post-processor that converts bare file paths and URLs in rendered gate prompts into clickable links
- File paths are validated against the workflow root (path traversal blocked)
- Skips fenced code blocks, inline code, and existing markdown links
- Applied to both terminal (Rich) and web dashboard rendering

### Web Dashboard: File Viewing
- \GET /api/files/{path}\ endpoint serves local files relative to the workflow root with security guards (path traversal protection, extension allowlist, 1MB size limit)
- \FileViewer\ modal component renders linked files with Markdown support
- Relative links in gate prompts open in the FileViewer; external URLs open in new tabs

### Web Dashboard: GFM Table Support
- Added \
emark-gfm\ plugin for GitHub Flavored Markdown table parsing
- Added table/th/td component overrides to both GateDetail and FileViewer

### Documentation
- Added 'Markdown in Gate Prompts' section to \docs/workflow-syntax.md\ documenting all supported features with examples

## Tests
- 13 file API security tests (path traversal, extension allowlist, size limits)
- 3 gate Markdown rendering tests
- Auto-linkify unit tests (213 lines covering URLs, file paths, code block skipping, edge cases)


src/conductor/executor/linkify.py
```diff
@@ -0,0 +1,263 @@
...
+    # Must not look like a URL (already handled)
+    if re.match(r"https?://", stripped):
+        return None
+
+    # If base_dir is provided, verify file exists
+    if base_dir is not None:
+        try:
+            candidate = (base_dir / stripped).resolve()
+            # Security: must be within base_dir
+            if not str(candidate).startswith(str(base_dir.resolve())):
```
jrob5756: **Important — fragile string-prefix containment check.**

```python
if not str(candidate).startswith(str(base_dir.resolve())):
```

This has the classic prefix-confusion bug: `/foo/bar` is a string prefix of `/foo/bar2`. The new `server.py` (line ~290) correctly uses `target.relative_to(self._workflow_root)` in a `try/except ValueError` — please apply the same pattern here, or use `candidate.is_relative_to(base_dir.resolve())` (stdlib since 3.9; project requires 3.12+).

Not exploitable in practice today because `candidate` is also resolved and gate roots don't usually have sibling-prefix names, but it's a footgun worth removing.

src/conductor/executor/linkify.py
```diff
@@ -0,0 +1,263 @@
...
+        ".cfg",
+        ".ini",
+        ".csv",
+        ".xml",
+        ".sh",
+        ".bat",
+        ".ps1",
+        ".plan.md",
+    }
+)
```
jrob5756, line 45: **Important — extension allowlist already drifting between modules.**

This `LINKABLE_EXTENSIONS` (21 entries, includes `.plan.md`) and `web/server.py`'s `_FILE_ALLOWED_EXTENSIONS` (20 entries, missing `.plan.md`) are intended to stay in sync — the docstring above even says so. They've already drifted on this very PR.

Suggest: have `server.py` import this constant directly, e.g. `from conductor.executor.linkify import LINKABLE_EXTENSIONS`.

Side note: the `.plan.md` entry here is dead anyway — `Path.suffix` returns only `.md` for `foo.plan.md`, so it's already covered by the `.md` entry in server.py. Consolidating to one source of truth makes this kind of confusion go away.

src/conductor/executor/linkify.py
```diff
@@ -0,0 +1,263 @@
...
+
+# Bare file path: contains at least one /, ends with a known extension.
+# Must start at a word boundary or line start.  Avoids matching inside
+# URLs (already handled) by requiring no scheme prefix.
+_FILE_PATH_RE = re.compile(
+    r"(?<![a-zA-Z0-9_/\\])"  # not preceded by path-like chars (avoids partial matches)
+    r"(?!https?://)"  # not a URL
+    r"(?:[a-zA-Z0-9_.][a-zA-Z0-9_./-]*[a-zA-Z0-9_]"  # path chars with at least one /
+    r")"
+)
```
jrob5756: **Suggestion — dead regex.**

`_FILE_PATH_RE` is defined here but never referenced (`grep -rn _FILE_PATH_RE src/ tests/` returns only this definition). The actual file-path linkification uses `re.split(r"(\s+)", segment)` token-by-token in `_try_linkify_path`. Safe to delete (along with the explanatory comment block immediately above).

src/conductor/executor/linkify.py
```diff
@@ -0,0 +1,263 @@
...
+
+    # Build markdown link with forward slashes (for dashboard API)
+    link_target = normalized
+    return f"{prefix}[{stripped}]({link_target}){suffix}"
+
+
+def _has_linkable_extension(path: str) -> bool:
+    """Check if a path ends with a known linkable extension."""
+    lower = path.lower()
+    return any(lower.endswith(ext) for ext in LINKABLE_EXTENSIONS)
```
jrob5756: **Nit — `str.endswith` accepts a tuple natively.**

```python
def _has_linkable_extension(path: str) -> bool:
    lower = path.lower()
    return any(lower.endswith(ext) for ext in LINKABLE_EXTENSIONS)
```

Can be:

```python
_LINKABLE_SUFFIXES = tuple(LINKABLE_EXTENSIONS)
...
def _has_linkable_extension(path: str) -> bool:
    return path.lower().endswith(_LINKABLE_SUFFIXES)
```

No Python-level loop, slightly faster, more idiomatic.

src/conductor/executor/linkify.py
```diff
@@ -0,0 +1,263 @@
...
+            if not str(candidate).startswith(str(base_dir.resolve())):
+                return None
+            if not candidate.is_file():
+                return None
+        except (OSError, ValueError):
+            return None
+
+    # Build markdown link with forward slashes (for dashboard API)
+    link_target = normalized
+    return f"{prefix}[{stripped}]({link_target}){suffix}"
```
jrob5756: **Nit — dead local.**

```python
link_target = normalized
return f"{prefix}[{stripped}]({link_target}){suffix}"
```

Inline `normalized` directly into the f-string and drop the `link_target` line.

src/conductor/web/server.py
```diff
@@ -217,6 +247,81 @@ async def resume_agent() -> JSONResponse:
...
+            # Reject absolute, drive-qualified, UNC, and scheme-prefixed paths
+            if (
+                file_path.startswith(("/", "\\"))
+                or "://" in file_path
+                or (len(file_path) >= 2 and file_path[1] == ":")
+            ):
+                return JSONResponse(
+                    {"error": "Absolute paths are not allowed"},
+                    status_code=403,
+                )
```
jrob5756, line 255: **Nit — hand-rolled absolute-path detection.**

The combined absolute / UNC / drive-letter check could be `PurePosixPath(file_path).is_absolute() or PureWindowsPath(file_path).is_absolute()` (or `PurePath`). Keep the `"://"` scheme check separate. Reduces a brittle hand-rolled check to a documented stdlib call.

src/conductor/engine/workflow.py
```diff
@@ -830,16 +831,18 @@ async def _handle_gate_with_web(
...
+            gate_base = Path(self.workflow_path).resolve().parent if self.workflow_path else None
+            return await self.gate_handler.handle_gate(agent, agent_context, base_dir=gate_base)
 
         # Race CLI vs web input. We start the web task unconditionally (not only
         # when a client is currently connected), because the human often opens
         # the per-run dashboard AFTER seeing the gate-waiting notification.
         # If we bail early when ``has_connections()`` is False, a later click
         # in the dashboard pushes a message to ``_gate_response_queue`` that
         # nobody is awaiting, and the workflow hangs forever.
+        gate_base = Path(self.workflow_path).resolve().parent if self.workflow_path else None
```
jrob5756: **Suggestion — triplicated `gate_base` derivation.**

The same expression appears at lines 834, 843, and 1433-1436 (and a similar pattern at line 583 for `workflow_dir`):

```python
gate_base = Path(self.workflow_path).resolve().parent if self.workflow_path else None
```

Extract a cached property:

```python
@property
def _workflow_dir(self) -> Path | None:
    return Path(self.workflow_path).resolve().parent if self.workflow_path else None
```

and reuse at all four call sites.

jrob5756 (commented):
A few comments before we merge.

jrob5756 (approved):
Approving — security review came back clean (defense-in-depth on the file API, ReactMarkdown sanitization, scheme filtering on relative-link detection, good test coverage).

The 7 inline comments from my [earlier review](https://github.com/microsoft/conductor/pull/131#pullrequestreview-4221694382) are still applicable but none are merge-blockers. Worth doing as a follow-up:

- **`linkify.py:248`** — string-prefix containment check. Switch to `Path.is_relative_to` / `relative_to(...) try/except` to match `server.py`'s pattern. Not exploitable today but a footgun.
- **`linkify.py` ↔ `server.py` allowlist drift** — already happening on this PR (`.plan.md`). Consolidate to one source of truth.
- Misc dead code (`_FILE_PATH_RE`, `link_target` local) and small DRY wins.

Ship it. 🚀

PolyphonyRequiem:
Addressed all 7 review comments in commit 6c37af1:

1. **Path containment (Important)** - Resolved via upstream merge; now uses is_relative_to() instead of the fragile string-prefix comparison.

2. **Extension allowlist drift (Important)** - server.py now imports LINKABLE_EXTENSIONS directly from linkify.py (single source of truth). Removed the dead .plan.md entry (Path.suffix returns .md anyway).

3. **Dead regex** - Removed _FILE_PATH_RE (defined but never referenced).

4. **endswith tuple idiom (Nit)** - _has_linkable_extension now uses path.lower().endswith(_LINKABLE_SUFFIXES) with a pre-computed tuple. No Python-level loop.

5. **Dead local (Nit)** - Inlined normalized directly into the f-string, removed link_target.

6. **Absolute-path detection (Nit)** - Replaced hand-rolled check with PurePosixPath(file_path).is_absolute() or PureWindowsPath(file_path).is_absolute(). Kept the :// scheme check separate as suggested.

7. **Triplicated gate_base derivation (Suggestion)** - Extracted a _workflow_dir property on WorkflowEngine and replaced all three call sites.

All 87 linkify+server tests and 517 engine tests pass.

PR #139, opened 2026-05-03
fix(engine): coerce Python literal "True"/"False"/"None" in workflow output

PolyphonyRequiem:
## Problem

Workflow output templates pass through `_maybe_parse_json` (`engine/workflow.py:3398`) to convert JSON-shaped strings back into native Python types. The function only recognized **lowercase** JSON literals (`true` / `false` / `null`).

In practice, workflow author expressions like:

```yaml
output:
  matched: "{{ left == right }}"
  done:    "{{ count >= threshold }}"
```

render their `bool` results via Jinja's default `str()`, producing the strings `"True"` / `"False"`. These then survived `_maybe_parse_json` unchanged — as **truthy non-empty strings** — so downstream route `when:` clauses comparing them against `true` / `false` silently misbehaved. The string `"False"` is truthy and `== true` is also `false`, so behavior was hard to reason about and harder to spot.

A `{{ none }}` expression has the same shape: renders as `"None"`, survives as a string.

## Fix

Recognize the three Python-literal forms (`"True"` / `"False"` / `"None"`) explicitly before the existing JSON-literal check. Lowercase JSON literals continue to coerce as before.

Three lines of behavior change in `_maybe_parse_json`:

```python
if stripped == "True":
    return True
if stripped == "False":
    return False
if stripped == "None":
    return None
```

## Tests

Added 3 cases to `TestWorkflowEngineOutputTemplates`:

- `test_output_template_python_bool_literals` — verifies `{{ a == b }}` / `{{ a != b }}` produce native `bool`.
- `test_output_template_python_none_literal` — verifies `{{ none }}` produces native `None`.
- `test_output_template_lowercase_json_literals_still_work` — regression check for `true` / `false` / `null`.

All 77 tests in `test_engine/test_workflow.py` pass; ruff clean.

## Why this matters

The mismatch between Jinja's `str(bool)` output and the JSON-literal recognition list was a stable footgun: workflows lint clean, validate clean, run without errors, and silently take wrong route branches. Even authors who knew about it had to reach for awkward workarounds like `{{ (a == b) | string | lower }}` or `{{ 1 if a == b else 0 }}`.

This patch is intentionally small and additive — no behavior change for any input that already coerced correctly. Happy to adjust scope if you'd prefer a different approach (e.g. coerce at template-render time rather than at output-map time).

### DIFF
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -3018,6 +3018,12 @@ class WorkflowEngine:
     def _maybe_parse_json(value: str) -> Any:
         """Attempt to parse a string as JSON.
 
+        Also coerces Python literal string forms ("True", "False", "None") that
+        commonly arise from Jinja expressions like ``{{ a == b }}`` rendering a
+        Python ``bool`` via ``str()``. Without this, those values survive as
+        truthy non-empty strings downstream and silently misbehave in route
+        ``when:`` clauses.
+
         Args:
             value: The string to parse.
 
@@ -3025,6 +3031,14 @@ class WorkflowEngine:
             Parsed JSON value if successful, original string otherwise.
         """
         stripped = value.strip()
+        # Python literal forms produced by str(bool) / str(None) — common from
+        # Jinja expressions in workflow output templates.
+        if stripped == "True":
+            return True
+        if stripped == "False":
+            return False
+        if stripped == "None":
+            return None
         if stripped.startswith(("{", "[", '"')) or stripped in ("true", "false", "null"):
             try:
                 return json.loads(stripped)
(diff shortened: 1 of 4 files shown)


PolyphonyRequiem:
The previous CI run surfaced 2 failing tests. Both were asserting the broken behavior this PR fixes (with inline comments acknowledging the workaround):

- `test_examples.py:214` — `result["syntax_passed"] == "True"  # Templates return strings`
- `test_parallel_workflows.py:410` — `result["success"] == "True"  # Boolean rendered as string`

Pushed [4a61195](https://github.com/PolyphonyRequiem/conductor/commit/4a61195) updating both to assert native `bool` (`is True`).

This is also a useful signal: those tests effectively codified the bug as expected behavior. Worth flagging to anyone reviewing — the change is intentional, not an oversight in the assertions.

jrob5756 (approved):
LGTM. Surgical fix with clear docstring and good test coverage (new behavior + None + regression for lowercase JSON literals). The two updated existing tests honestly correct previously-asserted-but-buggy behavior.

PR #146, opened 2026-05-04
fix(web): dedupe phantom edges, for-each dive-in nav, PR #113 review nits (#145)

### LINKED ISSUE #145: Follow-up from PR #113: tighten proactor AssertionError suppression + small comment/docstring nits
Follow-up to #113 (merged as b3e3485d). Three items from the review were not addressed and are deferred here.

## I3 — Over-broad `AssertionError` suppression during server shutdown (Important)

**File:** `src/conductor/web/server.py:425`

`_is_proactor_shutdown_race` is documented to return True "only when … the traceback (if available) originates from asyncio internals", but the implementation falls through to an unconditional `return True` when the traceback is missing or the last frame isn't from asyncio. Combined with the DEBUG-level logging in `_loop_exception_handler` (line 435) and `_guarded_serve` (line 469), unrelated `AssertionError`s raised during shutdown — e.g., from a workflow callback finishing late — get silently swallowed at default log levels.

**Scope:** Only fires when uvicorn's `should_exit` is set, narrowing blast radius. Windows + Python 3.14+ is the documented trigger.

**Suggested fix (pick one):**
- Tighten the gate: `return tb is not None and bool(frames) and 'asyncio' in frames[-1].filename` (false otherwise) — matches the docstring contract.
- OR keep the permissive fallback but log it at `WARNING` with `exc_info=True` so operators can see what was suppressed.

## S2 — Clear-before-raise rationale comment (Suggestion)

**File:** `src/conductor/engine/workflow.py` around `_check_interrupt` (lines ~1145-1152)

The interrupt event is cleared just before the conditional `raise InterruptError` for `_subworkflow_depth > 0`. Correct today (`is_set()` is the only consumer; the unwind chain doesn't re-check), but brittle if anyone later adds a between-agent recheck after the catch — they'll find the event already gone. A short comment would prevent future foot-guns.

## S3 — `resolveSlotPath` docstring orphaning warning (Suggestion)

**File:** `src/conductor/web/frontend/src/stores/workflow-store.ts` (function around line 404)

The implementation correctly matches the newest matching slot for re-runs. The docstring says nothing about the consequence: older iterations of the same slot become unreachable for late-arriving events targeting them. Worth a one-line note in the docstring.

---
Original review: https://github.com/microsoft/conductor/pull/113#pullrequestreview-4221689127

PolyphonyRequiem:
Follow-ups for #113 — bundles four fixes surfaced after merge.

## 1. Phantom edges in subworkflow graphs (commit fcab784)
Workflows that use a conditional + bare catch-all route to the same target (e.g. `- to:  when: X` followed by `- to: `) emitted two RouteEdges with identical (source, target). `graph-layout.ts` rendered each as a separate edge, which dagre then laid out as overlapping/diverging strands — hence the visual "phantom" edges. Fix: dedupe by `(from, to)`, first route wins (matches engine first-match-wins).

Repro: open dashboard at root, click into `implementation` subworkflow → confirmed gone after the fix.

## 2. For-each iteration dive-in (this commit)
The detail panel for a `for_each` group listed iterations but offered no way to navigate into a specific iteration's subworkflow context. `pg_execution_group[0]` could not be linked into. Added a Layers icon button per iteration row that resolves the slot key (`<group>[<key>]`) against `subworkflowContexts` and calls `navigateIntoSubworkflow`. Click + Enter/Space supported.

## 3. Tighten `_is_proactor_shutdown_race` — #145 (I3)
Per Jason's review: the proactor race gate fell back to permissive `True` when no traceback was attached, which could swallow unrelated AssertionErrors raised by user/workflow code during shutdown. Now requires the deepest traceback frame to originate from asyncio internals. Tests updated to assert the tightened contract (asyncio frame → True; non-asyncio frame → False; missing traceback → False).

## 4. Doc nits — #145 (S2, S3)
* `engine/workflow.py`: comment on why `_check_interrupt` clears the interrupt event before raising (single-listener reset semantics, no race).
* `stores/workflow-store.ts`: docstring on `resolveSlotPath` noting that newest-wins semantics make older for-each iterations unreachable by stale slot paths.

## Validation
* 137 web + engine tests pass
* `ruff check` + `ruff format --check` clean
* Frontend bundle rebuilt; `index.html` references new asset filenames

## Refs
* Closes #145
* Builds on #113

🤖 Co-authored with [Copilot CLI](https://github.com/github/gh-copilot)


src/conductor/web/server.py
```diff
@@ -405,24 +405,25 @@ def _is_proactor_shutdown_race(self, context: dict[str, Any]) -> bool:
...
-            frames = tb_mod.extract_tb(tb)
-            if frames and "asyncio" in frames[-1].filename:
-                return True
-        # If no traceback but server is shutting down, still suppress —
-        # the only known source of AssertionError during shutdown is this race.
-        return True
+        if tb is None:
+            return False
+        frames = tb_mod.extract_tb(tb)
+        return bool(frames) and "asyncio" in frames[-1].filename
```
jrob5756, line 464: **`_guarded_serve` retains the same permissive `AssertionError` handling that was just tightened here.**

Issue #145 calls out **both** silent-swallow sites — `_loop_exception_handler` *and* `_guarded_serve` (lines 443-458 below). The wrapper still catches *any* `AssertionError` whenever `should_exit` is set, with no traceback inspection:

```python
except AssertionError:
    if self._server is not None and getattr(self._server, "should_exit", False):
        logger.debug(...)  # Suppressed silently
    else:
        raise
```

Suggested options:
- Apply the same asyncio-frame check inline in the `except`, or
- Refactor: synthesize a context dict from the exception and call `self._is_proactor_shutdown_race(ctx)` so both sites share one gate.

src/conductor/web/frontend/src/components/graph/graph-layout.ts
```diff
@@ -192,13 +192,24 @@ export function buildGraphElements(
...
     const from = childToParent.get(r.from) ?? r.from;
     const to = childToParent.get(r.to) ?? r.to;
     if (!nodeIds.has(from) || !nodeIds.has(to)) continue;
     // Skip self-loops created by remapping (e.g. group member → group member)
     if (from === to) continue;
-    const edgeId = `${from}->${to}${r.when ? `[${r.when}]` : ''}`;
+    const pairKey = `${from}->${to}`;
+    if (seenPairs.has(pairKey)) continue;
+    seenPairs.add(pairKey);
+    const edgeId = `${pairKey}${r.when ? `[${r.when}]` : ''}`;
```
jrob5756, line 222: **Nit/Suggestion — Edge dedupe is lossy on `when` metadata.**

The surviving edge keeps `data: { when: r.when }` from the *first* route only. For the documented repro (conditional + bare catch-all) that's fine — the second route has no `when`. But for two-conditional routes to the same target (e.g. `when: success` then `when: error`), the visual edge will display only `[success]` even though the actual graph means "go to $end on either condition."

Consider clearing `when` (or marking it `'*'` / `'multiple'`) when collapsing routes with non-equivalent conditions, e.g.:

```ts
const existing = seenPairs.get(pairKey);
if (existing && existing !== r.when) {
  // Multiple distinct conditions collapse — drop the label.
  flowEdges[existing.idx].data = { when: undefined };
  continue;
}
```

Minor UX nit, not blocking.

jrob5756 (commented):
two small fixes

PolyphonyRequiem:
Addressed both review comments in commit 9c0940f:

1. **_guarded_serve tightened** - now calls _is_proactor_shutdown_race(ctx) with a synthesized context dict instead of the permissive should_exit-only check. Both exception-handler sites share the same strict asyncio-frame gate. Added test_guarded_serve_reraises_non_asyncio_assertion_during_shutdown to verify the tightened contract.

2. **Edge dedupe preserves label accuracy** - changed seenPairs from a Set to a Map tracking both the when value and the edge index. When collapsing routes with different conditions to the same target, the label is cleared (undefined) so the UI doesn't misleadingly show only one condition.

All 48 web server tests pass.

jrob5756 (approved):
LGTM — both review findings addressed cleanly. Shared gate for the proactor race is the right call.

PR #150, opened 2026-05-05
feat: add per-agent timeout_seconds for hard wall-clock timeouts (#82)

### LINKED ISSUE #82: Feature: Agent-Level Timeouts
## Summary

Add per-agent `timeout_seconds` configuration so individual slow agents don't block entire workflows.

## Motivation

Conductor only supports workflow-level `timeout_seconds` today. With agents becoming more autonomous and models varying widely in response time, one slow agent can consume the entire workflow's time budget. Research on "context rot" shows agent performance degrades after ~60% context fill — timeouts prevent wasted tokens on degraded agents.

This is already identified as a gap in the conductor docs.

## Proposed Design

```yaml
agents:
  - name: slow_researcher
    model: claude-opus-4.5
    timeout_seconds: 120          # agent-specific hard limit

  - name: fast_classifier
    model: gpt-5.2-mini
    timeout_seconds: 15           # tight limit for simple tasks
```

### Behavior

- Agent execution cancelled after `timeout_seconds` wall-clock time
- Raises `TimeoutError` — handled by existing error semantics (`fail_fast`, `continue_on_error`)
- Workflow-level timeout still applies as an overall cap
- Events emitted: `agent_timeout` with agent name, elapsed time, and configured limit

### Interaction with Other Features

| Feature | Interaction |
|---------|------------|
| **Retry policies** (#80) | `retry_on: [timeout]` enables automatic retry on timeout |
| **Fallback models** (#84) | Timed-out agent retries with `fallback_model` for faster response |
| **Parallel execution** | Timed-out agent in parallel group follows the group's failure mode |
| **Workflow timeout** | Agent timeout cannot exceed remaining workflow timeout |

### Example: Timeout + Retry + Fallback

```yaml
agents:
  - name: researcher
    model: claude-opus-4.5
    timeout_seconds: 120
    fallback_model: claude-haiku-4.5    # faster model on retry
    retry:
      max_attempts: 2
      retry_on:
        - timeout
```

If `claude-opus-4.5` times out at 120s → retry with `claude-haiku-4.5` (much faster) → complete within budget.

## Why It Fits Conductor

- Trivial YAML addition — single field per agent
- Builds on existing `LimitEnforcer` infrastructure
- Implementation: wraps agent execution in `asyncio.wait_for()` with the configured timeout
- Combined with retry + fallback_model, creates a complete resilience story

## Effort Estimate

Low — wraps agent execution in `asyncio.wait_for()`, adds one schema field, emits one new event.

brrusino:
Add timeout_seconds field to AgentDef that wraps agent execution in asyncio.wait_for() at the engine level. This provides hard cancellation for slow agents without blocking entire workflows.

Key behaviors:
- Effective timeout = min(agent.timeout_seconds, remaining_workflow_timeout)
- When workflow timeout is stricter, it owns the error (no mislabeling)
- Emits agent_timeout event with agent name, elapsed time, and limit
- Raises AgentTimeoutError (subclass of TimeoutError) for existing error handling semantics (fail_fast, continue_on_error)
- Scoped to provider-backed agents only (script uses 'timeout', human_gate/workflow types rejected)
- Applied at all execution sites: main loop, parallel groups, for-each groups

Schema:
- New field: timeout_seconds: float | None (ge=1.0)
- Rejected for script, human_gate, and workflow agent types

Exception:
- New AgentTimeoutError class with agent_name attribute

Console:
- New verbose_log_agent_timeout handler for agent_timeout events

Closes #82 .

### DIFF
--- a/src/conductor/config/schema.py
+++ b/src/conductor/config/schema.py
@@ -582,6 +582,12 @@ class AgentDef(BaseModel):
     or workflow types). This is a hard cancellation — unlike
     ``max_session_seconds`` which checks between provider iterations.
 
+    Because this is a hard cancellation, in-flight provider sessions,
+    MCP tool calls, and HTTP connections receive ``CancelledError``
+    mid-flight and may not get a clean shutdown. External state (e.g.,
+    partially-written files, open MCP tool handles) may be left
+    inconsistent.
+
     Note: Agent-level timeouts are non-retryable. The retry policy
     operates inside the provider and is cancelled along with the agent.
 
--- a/src/conductor/config/schema.py
+++ b/src/conductor/config/schema.py
@@ -567,6 +567,29 @@ class AgentDef(BaseModel):
         max_depth: 3  # Allow at most 3 levels of recursion
     """
 
+    timeout_seconds: float | None = Field(None, ge=1.0)
+    """Hard wall-clock timeout for this agent's execution in seconds.
+
+    When set, the engine wraps the entire agent execution in
+    ``asyncio.wait_for()``. If exceeded, raises ``AgentTimeoutError``
+    which is handled by existing error semantics (``fail_fast``,
+    ``continue_on_error``).
+
+    The effective timeout is ``min(timeout_seconds, remaining_workflow_timeout)``
+    so agent timeouts never exceed the workflow-level limit.
+
+    Only applies to provider-backed agents (not script, human_gate,
+    or workflow types). This is a hard cancellation — unlike
+    ``max_session_seconds`` which checks between provider iterations.
+
+    Note: Agent-level timeouts are non-retryable. The retry policy
+    operates inside the provider and is cancelled along with the agent.
+
+    Example::
+
+        timeout_seconds: 120  # Cancel agent after 2 minutes
+    """
+
     max_session_seconds: float | None = Field(None, ge=1.0)
     """Maximum wall-clock duration for this agent's session in seconds.
 
@@ -666,6 +689,8 @@ class AgentDef(BaseModel):
                 raise ValueError("human_gate agents cannot have 'max_depth'")
             if self.reasoning is not None:
                 raise ValueError("human_gate agents cannot have 'reasoning'")
+            if self.timeout_seconds is not None:
+                raise ValueError("human_gate agents cannot have 'timeout_seconds'")
         elif self.type == "script":
             if not self.command:
                 raise ValueError("script agents require 'command'")
@@ -700,6 +725,11 @@ class AgentDef(BaseModel):
                 raise ValueError("script agents cannot have 'max_depth'")
             if self.reasoning is not None:
                 raise ValueError("script agents cannot have 'reasoning'")
+            if self.timeout_seconds is not None:
+                raise ValueError(
+                    "script agents cannot have 'timeout_seconds' "
+                    "(use 'timeout' for script-specific timeouts)"
+                )
         elif self.type == "workflow":
             if not self.workflow:
                 raise ValueError("workflow agents require 'workflow' path")
@@ -725,6 +755,8 @@ class AgentDef(BaseModel):
                 raise ValueError("workflow agents cannot have 'retry'")
             if self.dialog is not None:
                 raise ValueError("workflow agents cannot have 'dialog'")
+            if self.timeout_seconds is not None:
+                raise ValueError("workflow agents cannot have 'timeout_seconds'")
         else:
             # Regular agent or human_gate — input_mapping is not valid
             if self.input_mapping is not None:
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -24,6 +24,7 @@ from conductor.engine.router import Router, RouteResult
 from conductor.engine.usage import UsageTracker, WorkflowUsage
 from conductor.events import WorkflowEvent, WorkflowEventEmitter
 from conductor.exceptions import (
+    AgentTimeoutError,
     ConductorError,
     ExecutionError,
     InterruptError,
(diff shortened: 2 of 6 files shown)


src/conductor/engine/workflow.py
```diff
@@ -601,6 +602,60 @@ async def _execute_script(self, agent: AgentDef, context: dict[str, Any]) -> Scr
...
+                    "elapsed": elapsed,
+                    "timeout_seconds": agent.timeout_seconds,
+                },
+            )
+            raise AgentTimeoutError(
+                agent_name=agent.name,
+                elapsed_seconds=elapsed,
+                timeout_seconds=agent.timeout_seconds,
+            ) from None
+
```
jrob5756, line 659: Consider `from e` instead of `from None` here.

`from None` clears the exception chain, but for a hard-cancellation feature where users may need to debug what was running when cancelled, preserving the underlying `asyncio.TimeoutError` (`from e` after `except TimeoutError as e:`) gives more diagnostic context with no significant downside in user-facing tracebacks.

src/conductor/config/schema.py
```diff
@@ -532,6 +532,29 @@ class AgentDef(BaseModel):
         max_depth: 3  # Allow at most 3 levels of recursion
     """
 
+    timeout_seconds: float | None = Field(None, ge=1.0)
```
jrob5756, line 570: Worth adding a sentence to this docstring noting the operational implication of hard cancellation: in-flight provider sessions, MCP tool calls, and HTTP connections receive `CancelledError` mid-flight and may not get a clean shutdown. That helps users understand both why this is non-retryable and why external state (e.g., partially-written files, open MCP tool handles) may be left inconsistent.

src/conductor/exceptions.py
```diff
@@ -443,6 +443,45 @@ def __init__(
...
+        super().__init__(
+            message,
+            elapsed_seconds=elapsed_seconds,
+            timeout_seconds=timeout_seconds,
+            current_agent=agent_name,
+            suggestion=suggestion,
+        )
+        # Set after super().__init__() because ExecutionError.__init__
+        # defaults agent_name to None and would overwrite an earlier assignment.
+        self.agent_name = agent_name
```
jrob5756, line 484: Minor: `self.agent_name` and `self.current_agent` (set via `super().__init__(current_agent=agent_name, ...)`) now both hold the same value. The comment correctly explains the override ordering, but consider either:

1. Just relying on `current_agent` (consistent with parent `TimeoutError`), or
2. Documenting why both exist if downstream consumers depend on `agent_name` specifically.

The parallel-group error wrapper at `workflow.py:~2871` uses `_parallel_agent_name` rather than either of these, so it's not clear who consumes `.agent_name` here.

tests/test_engine/test_agent_timeout.py
```diff
@@ -0,0 +1,575 @@
...
+                    routes=[RouteDef(to="$end")],
+                ),
+            ],
+            output={"answer": "{{ agent1.output.answer }}"},
+        )
+
+        provider = CopilotProvider(mock_handler=lambda a, p, c: {"answer": "42"})
+        engine = WorkflowEngine(config, provider)
+        result = await engine.run({"question": "test"})
+        assert result["answer"] == 42 or result["answer"] == "42"
```
jrob5756: Brittle assertion — the `or` suggests uncertainty about the output type. With `OutputField(type="string")` and `mock_handler` returning `{"answer": "42"}`, the result should be deterministic. Pin it to one expected value (likely `"42"`) so the test fails loudly if coercion behavior changes.

tests/test_engine/test_agent_timeout.py
```diff
@@ -0,0 +1,575 @@
...
+
+    def test_agent_timeout_error_suggestion(self) -> None:
+        """AgentTimeoutError has a helpful suggestion."""
+        err = AgentTimeoutError(
+            agent_name="researcher",
+            elapsed_seconds=120.5,
+            timeout_seconds=120.0,
+        )
+        assert "timeout_seconds" in err.suggestion
+        assert "researcher" in err.suggestion
```
jrob5756, line 576: Missing for-each integration coverage. The helper `_execute_with_agent_timeout` is applied at three sites in `workflow.py` (main loop, parallel groups, for-each groups), but only main loop and parallel are exercised end-to-end here. Consider adding a `TestAgentTimeoutForEach` class mirroring `TestAgentTimeoutParallel` (both `fail_fast` and `continue_on_error` modes) so the for-each call site doesn't regress silently.

jrob5756 (commented):
Nice, focused PR. CI is green and the helper-based design is clean. A few suggestions inline. Most are minor; the missing for-each integration test and the brittle output-type assertion are the two I'd address before merging. Also looks like there is a merge conflict :)

src/conductor/engine/workflow.py
brrusino: Good call — changed to from e so the original asyncio.TimeoutError is preserved in the chain.

src/conductor/config/schema.py
brrusino: Yeah, that's a great point. Added a paragraph noting that in-flight provider sessions, MCP tool calls, and HTTP connections receive CancelledError mid-flight and external state may be left inconsistent.

src/conductor/exceptions.py
brrusino: Good question. I kept both because ExecutionError exposes .agent_name and TimeoutError exposes .current_agent — downstream consumers could check either depending on which exception type they catch. Updated the class docstring to document that both are set to the same value and why.

tests/test_engine/test_agent_timeout.py
brrusino: Yep — pinned it to 42 (int). The output template renders the string through Jinja2 which coerces it.

tests/test_engine/test_agent_timeout.py
brrusino: Added TestAgentTimeoutForEach with both fail_fast and continue_on_error modes — same structure as the parallel group tests.

Issue #195, opened 2026-05-15
`--web-bg` child process dies when launched from a job-managed shell on Windows (missing CREATE_BREAKAWAY_FROM_JOB)


jrob5756:
### Summary

`conductor run <workflow> --web-bg` works fine from a regular interactive PowerShell window, but the bg child process is killed within ~10 seconds when launched from any shell wrapper that runs commands inside a Windows job object with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`. Common environments that trigger this: GitHub Actions runners, VS Code integrated terminal, JetBrains IDE terminals, and the GitHub Copilot CLI shell tool.

The bg child inherits the parent''s job object because `subprocess.Popen` is called with `CREATE_NEW_PROCESS_GROUP` only — there is no `CREATE_BREAKAWAY_FROM_JOB`. When the synchronous wrapper completes and tears down the job, every process in that job dies, including the supposedly-detached bg child.

### Repro

From inside a job-managed shell (the GitHub Copilot CLI shell tool reproduces it 100% on my machine):

```powershell
conductor run sdd-plan --web-bg --input purpose="anything"
```

Output:
```
Dashboard: http://127.0.0.1:50699
Workflow running in background. Dashboard auto-shuts down after workflow completes and all clients disconnect.
```

Within ~10 seconds:
- `Get-Process -Id <pid>` returns nothing
- `Test-NetConnection -Port <port>` returns `False`
- The PID file at `~/.conductor/runs/plan-<port>.pid` is left behind, never cleaned up
- The bg child''s `events.jsonl` contains exactly one event: `workflow_started`
- The bg child''s `--log-file auto` file ends at `Starting workflow execution...`

### Diagnostic isolation

I reproduced the exact `bg_runner.launch_background()` Popen call from a Python wrapper script in the same shell:

```python
proc = subprocess.Popen(
    [python, "-m", "conductor", "--silent", "run", workflow,
     "--web", "--web-port", "63990", "--no-interactive",
     "--input", "purpose=test"],
    stdout=subprocess.DEVNULL,
    stderr=subprocess.DEVNULL,
    stdin=subprocess.DEVNULL,
    creationflags=subprocess.CREATE_NEW_PROCESS_GROUP,
    env={**os.environ, "CONDUCTOR_WEB_BG": "1", "CONDUCTOR_WEB_PORT": "63990"},
)
sys.exit(0)  # parent exits immediately, mimicking bg_runner
```

[...]

Issue #116, opened 2026-04-24
bug: intermittent startup crash — process dies between agent_started and first prompt render


franklixuefei:
## Problem

Conductor intermittently crashes during startup, after emitting `workflow_started` and `agent_started` events but before rendering the first agent prompt. The process exits silently with no error event logged. Retrying the same command succeeds.

Observed across 56 workflow runs of `pr-reviewer` (v0.1.8 and v0.1.9) on Windows over 8 days. 26 of 56 runs (46%) crashed on startup. Every crash is recoverable by simply re-running the same command.

### Crash signature

All crash logs have exactly 2 events and nearly identical file sizes (90,748-90,750 bytes):

```
Event 1: workflow_started  (timestamp T)
Event 2: agent_started     (timestamp T + 0.001s)
(EOF — no further events, no error event)
```

Time between `workflow_started` and `agent_started` is ~1-2ms. The process dies before the next event (`agent_prompt_rendered`) would be emitted.

### Crash rate

| Metric | Value |
|--------|-------|
| Total pr-reviewer runs | 56 |
| 2-event crashes | 26 (46%) |
| Successful runs | 30 |
| Crash log sizes | 90,748 (4x), 90,749 (6x), 90,750 (16x) |
| Successful log sizes | 92KB - 722KB |

### Crash-before-success patterns

Every crash was eventually followed by a successful run of the same workflow with the same arguments:

```
5 crash(es) then success
1 crash then success
3 crash(es) then success
1 crash then success
1 crash then success
1 crash then success
2 crash(es) then success
1 crash then success
3 crash(es) then success
3 crash(es) then success
1 crash then success
2 crash(es) then success
2 crash(es) then success
```

Worst case: 5 consecutive crashes before success. No crash was permanent -- retrying always works.

### Near-crash logs (survived slightly longer)

Some runs survived past `agent_started` but died before completing the first turn:

| Events | Size | Last event |
|--------|------|------------|
| 2 | 90,750 bytes | `agent_started` (crash) |
| 3 | 92,645 bytes | `agent_prompt_rendered` (crash) |
| 4 | 92,764 bytes | `agent_turn_start` (crash) |

This suggests the failure window extends from `agent_started` through early `agent_turn_start`, with most crashes happening before `agent_prompt_rendered`.

## Environment

- Conductor: v0.1.9 (also observed on v0.1.8)
- Python: 3.13.13
- OS: Windows NT 10.0.26220.0 (Windows 11)
[...]

PR #204, opened 2026-05-18
fix(windows): make --web-bg startup crashes diagnosable (#116)

jrob5756:
## Summary

Phase 1 (diagnostics-only) for issue #116. **This PR does not yet fix the underlying Windows crash** — it makes the crash forensically traceable so Phase 2 can root-cause it from a real Windows repro.

Today, when a `conductor run <wf> --web-bg` (or `resume --web-bg`) child dies on Windows between `agent_started` and `agent_prompt_rendered` (~46% of runs of `pr-reviewer` per the report), it leaves:

- No traceback (parent redirects child stderr to `DEVNULL`).
- No native-crash dump (`faulthandler` was never enabled).
- No `workflow_failed` event (a `SystemExit` from a misbehaving library escapes the engine's `except Exception` arm).

This PR closes all three gaps.

## Changes

| File | Change |
|---|---|
| `src/conductor/__init__.py` | `faulthandler.enable(file=sys.__stderr__, all_threads=True)` at import time. Covers both `python -m conductor` and the installed `conductor` console script (which bypasses `__main__.py`). |
| `src/conductor/cli/bg_runner.py` | Capture child stdout/stderr to `\$TMPDIR/conductor/conductor-<name>-<ts>-<runid>.bg.{stderr,stdout}.log` instead of `DEVNULL`. Pass `CONDUCTOR_RUN_ID` / `CONDUCTOR_BG_STDERR_LOG` / `CONDUCTOR_BG_STDOUT_LOG` to the child via env so the events JSONL and bg log files share an 8-hex run id in their filenames. Close parent file handles in `try/finally`. Returns a new `BackgroundLaunch` dataclass (`url`, `stderr_log`, `stdout_log`, `run_id`). Threads the stderr-log path into every `RuntimeError` raised by `_finalize_background_launch`. |
| `src/conductor/engine/event_log.py` | `EventLogSubscriber` honours `CONDUCTOR_RUN_ID` (validated as 1–32 hex chars, lowercased), falls back to random when unset or malformed. |
| `src/conductor/engine/workflow.py` | `_build_system_metadata` surfaces `bg_stderr_log` / `bg_stdout_log`. `_execute_loop` gets `except asyncio.CancelledError: raise` (preserves dashboard-stop semantics) followed by `except BaseException` that emits `workflow_failed` with `is_base_exception: true`. `on_error` hook is deliberately **not** invoked for `BaseException` cases because its declared signature is `Exception \| None`. |
| `src/conductor/cli/app.py` | Both `--web-bg` callers updated for `BackgroundLaunch`; stderr-log path printed beside the dashboard URL. |

## Tests

**15 new, 5 updated.**

- `tests/test_cli/test_bg_runner.py` (new, 9 tests): `BackgroundLaunch` shape and filename pattern; file handles vs `DEVNULL` contract; parent-handle closure on success and on `_finalize_background_launch` failure; `CONDUCTOR_RUN_ID` / log-path env wiring; stderr-log path included in early-exit and dashboard-timeout `RuntimeError` messages; filename sanitisation.
- `tests/test_engine/test_event_log.py` (+4): `CONDUCTOR_RUN_ID` honoured, case-normalised, rejected when non-hex / overlong, empty falls back to random.
- `tests/test_engine/test_workflow.py` (+2): `TestExecuteLoopExceptionArms` covers `SystemExit` → `workflow_failed` with `is_base_exception=true`, and `CancelledError` → **no** spurious `workflow_failed` event.
- `tests/test_cli/test_web_flags.py` and `test_resume_command.py` updated for the new `BackgroundLaunch` return type and the no-DEVNULL contract (now asserts file-like stdout/stderr with `.bg.{stderr,stdout}.log` names and the new env vars).

All checks green locally on Python 3.14:
- \`make lint\` — clean.
- \`make typecheck\` — same single pre-existing unrelated warning in `dialog_evaluator.py:131`. Zero new diagnostics.
- \`uv run pytest -m "not real_api and not performance and not install_scripts"\` — **2627 passed, 3 skipped, 46 deselected** (baseline 2612; +15 new tests, no regressions).

## Docs

- `CHANGELOG.md` — entry under Unreleased / Fixed.
- `AGENTS.md` — `bg_runner.py` bullet describes the new capture behaviour and return type; new "Debugging \`--web-bg\` failures" section under Architecture / Key Patterns.

## API surface changes

- `launch_background()` and `launch_background_resume()` in `conductor.cli.bg_runner` now return a `BackgroundLaunch` dataclass instead of a bare URL string. Existing callers in `conductor.cli.app` are updated; external callers (if any) need to access `.url`.

## Intentionally NOT addressed

- **Root-cause fix for the Windows crash.** That's Phase 2, gated on having a real traceback from one of the new captured `.bg.stderr.log` files. Plan: ask the original reporter (\@franklixuefei) to retry on Windows once this lands, attach the captured log, and act on whatever it shows.
- **\`CREATE_BREAKAWAY_FROM_JOB\` on Windows Popen creationflags (#195).** Out of scope per direction — that's a separately tracked sibling bug with a different signature (one event before death, not two).

## Closes / refs

Refs #116 (diagnostics; the issue stays open until Phase 2 lands the actual fix).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -67,7 +67,7 @@ make validate-examples    # validate all examples
 - **cli/**: Typer-based CLI with commands `run`, `validate`, `init`, `templates`, `stop`, `update`, `resume`, `checkpoints`
   - `app.py` - Main entry point, defines the Typer application
   - `run.py` - Workflow execution command with verbose logging helpers
-  - `bg_runner.py` - Background process forking for `--web-bg` mode
+  - `bg_runner.py` - Background process forking for `--web-bg` mode. Captures the detached child's stdout/stderr to `$TMPDIR/conductor/conductor-<name>-<ts>-<runid>.bg.{stderr,stdout}.log` so silent crashes (uncaught Python exceptions, `faulthandler` dumps) leave a forensic trail — DEVNULL is **not** used for stdout/stderr. Passes `CONDUCTOR_RUN_ID`, `CONDUCTOR_BG_STDERR_LOG`, and `CONDUCTOR_BG_STDOUT_LOG` to the child via env so the child's `EventLogSubscriber` shares a run id with the bg log files and surfaces both paths in `workflow_started` system metadata. Returns a `BackgroundLaunch` dataclass (`url`, `stderr_log`, `stdout_log`, `run_id`).
   - `pid.py` - PID file utilities for tracking/stopping background processes
   - `update.py` - Update check and version comparison. Upgrades are delegated to the install script (`install.ps1`/`install.sh`); in-process self-upgrade was removed because on Windows the running Python interpreter sits inside the venv `uv tool install --force` is trying to recreate, which fails with "Access is denied". `conductor update` prints the OS-appropriate install-script one-liner; `conductor update --apply` spawns the installer detached (Windows: new console window; POSIX: `os.execvpe` replace) and exits the current process so file locks release. The startup hint is suppressed by `CONDUCTOR_NO_UPDATE_CHECK=1`, `--silent`, `--help`/`--version`, and the `update` subcommand itself.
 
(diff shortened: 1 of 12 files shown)


jrob5756:
Pushed 70049cb addressing the comprehensive PR review.

**Critical fix (silent-failure-hunter):**
- The new `except BaseException` arm in `_execute_loop` now wraps both `_emit` and `_save_checkpoint_on_failure` in their own `try/except Exception`. A misbehaving subscriber or checkpoint writer can no longer mask the original `BaseException` on its way out — which would have defeated the whole point of this PR.

**Simplification (code-simplifier):**
- Extracted `_spawn_bg_child` helper containing the previously-duplicated tail of `launch_background` and `launch_background_resume` (~45 LOC of byte-identical code per side). Both public functions now end with a one-line call. Keeps the detachment-flag + log-redirect + env-var contract in a single place.
- Collapsed the redundant `pid_workflow_ref is None` pragma'd guard.
- Hoisted `os` / `re` / `secrets` to module level in `event_log.py`; compiled `_RUN_ID_PATTERN` once at module load.

**Type invariants (type-design-analyzer):**
- Added `BackgroundLaunch.__post_init__` validation: enforces 8-hex `run_id`, `http://127.0.0.1:` URL prefix, and that the `run_id` is embedded in both log filenames. Catches construction bugs immediately at the type boundary.

**Hardening (silent-failure-hunter MEDIUMs):**
- `_close_quietly` now prints a stderr warning on close failure (disk-full / Windows-locked-file scenarios) instead of suppressing silently.
- `conductor/__init__.py` prints a stderr warning if `faulthandler.enable()` fails, so the user knows the diagnostic didn't arm.

**Tests (pr-test-analyzer 8-10 priority gaps):**
- `test_regular_exception_does_not_set_is_base_exception_flag` — Phase 2 contract baseline.
- `test_cancellederror_via_external_task_cancel` — replaces the synthetic synchronous-raise with real `task.cancel()` against an async `AgentExecutor.execute` patch, matching production dashboard-stop semantics.
- `test_bg_log_paths_in_workflow_started_system_metadata` + negative case.
- `CONDUCTOR_RUN_ID` 32-char accepted / 33-char rejected boundary tests.

**Comment fixes (comment-analyzer):**
- Tightened the subprocess-stdio mechanism explanation in `conductor/__init__.py`.
- Extended the `BaseException` / `CancelledError` arm comments with the rationale for skipping `on_error` and not saving a checkpoint on cancellation.

**Intentionally not actioned:**
- Pre-existing `_terminate_child` silent suppression — out of scope.
- `_open_bg_log_files` OSError handling already raises a contextual `RuntimeError` that `print_error` surfaces.

**Validation:** `make lint` clean; `make typecheck` no new diagnostics; full test suite 2632 passed (was 2627, +5 new tests).

PR #213, opened 2026-05-19
feat(notifications): user-defined domain notifications (type: notification)

PolyphonyRequiem:
## Summary

Adds a new `type: notification` step that lets workflow authors publish typed, declared **domain events** to a dedicated JSONL stream for external tooling to hook off.

Notifications are **distinct from execution events** (`agent_started`, `route_evaluated`, ...): execution events describe what the runtime is doing; notifications describe what the workflow is asserting to the outside world.

> **Notifications are a visibility primitive, not a control-flow primitive.** They do not block, do not wait for an ack, and do not gate routing. If a decision needs to flow back into the workflow, use `human_gate`.

## YAML surface

```yaml
workflow:
  notifications:
    namespace: polyphony.feature_pr   # defaults to slugified workflow name
    correlation: [apex_id]            # inherited by sub-workflows
    types:
      pr_ready:
        version: 1
        payload:
          pr_url: { type: string }
          pr_id:  { type: number }

agents:
  - name: announce_ready
    type: notification
    notification: pr_ready
    payload:
      pr_url: "{{ open_pr.output.url }}"
      pr_id:  "{{ open_pr.output.id }}"
    routes:
      - to: next_step       # emission is independent of routing
```

## Envelope shape

- `emission_id` = `<run_id>:<dotted_step_path>:<iteration>` — stable across resume/replay so the first downstream consumer can dedupe.
- `schema_id` = `<namespace>.<type>@<version>` — consumers MUST pin on this, not on `notification_type` alone.
- `correlation` — first-class field; auto-merged from declared keys and inherited from the parent workflow (parent wins on collision so the upstream trail survives deep nesting).
- `payload` — Jinja2-rendered then type-validated against the declared `OutputField` schema.

## Delivery (v1)

Every emission flows through `WorkflowEventEmitter` and is written by `NotificationLogSubscriber` to a dedicated `…notifications.jsonl` file next to the existing event log (path printed at end-of-run).

No webhooks, shell hooks, or model-driven emission in v1.

## What's in this PR

- `src/conductor/config/schema.py` — `NotificationsConfig` + `NotificationTypeDef`; `notification` and `payload` fields on `AgentDef`; strict positive/negative validators.
- `src/conductor/config/validator.py` — cross-references undeclared types, payload field mismatches, unknown correlation keys.
- `src/conductor/executor/notification.py` — pure envelope builder (rendering, coercion, type validation, emission-id construction).
- `src/conductor/engine/workflow.py` — dispatch branch + correlation snapshot inherited into sub-workflows.
- `src/conductor/engine/event_log.py` — `NotificationLogSubscriber` filters to `notification` events only.
- `src/conductor/cli/run.py` — wires the subscriber in both `run` and `resume` paths.
- `examples/notifications.yaml` — runnable example with namespace, correlation, mid-workflow emission.
- `AGENTS.md` — full design + schema-version-bump policy + "visibility not control flow" callout.
- Tests: `tests/test_config/test_notifications.py` + `tests/test_executor/test_notification.py` (18 focused tests).

## Schema-version-bump policy (documented, not enforced)

- Adding an optional field is non-breaking — do **not** bump `version`.
- Removing/renaming a field, changing a field's type, or making an optional field required is breaking — bump `version` and emit under the new `schema_id`.
- During a transition a workflow MAY emit both `pr_ready@1` and `pr_ready@2` from adjacent steps.
- Consumers MUST pin on `schema_id`.

## Validation

Undeclared type, payload field mismatch, unknown correlation key, or a type mismatch on a rendered value all fail the workflow loudly with `ValidationError`. The contract is the whole point.

## Testing

- `uv run ruff check` — clean
- `uv run ruff format --check` — clean
- `uv run pytest tests/test_config tests/test_executor tests/test_engine tests/test_cli -m "not performance"` — 1692 passed, 5 skipped
- `uv run conductor validate examples/notifications.yaml` — clean
- All other `examples/*.yaml` still validate clean.

## Deliberately deferred

- Edge-attached notifications (`notify:` on `RouteDef`) — possible v1.5 sugar reusing the same envelope.
- Webhooks / shell hooks for delivery.
- MCP tool for model-driven emission.
- Ack / response semantics (use `human_gate` instead).

### DIFF
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -1232,6 +1232,7 @@ async def run_workflow_async(
     # Always create event emitter and JSONL log subscriber
     emitter = WorkflowEventEmitter()
     event_log_subscriber: Any = None
+    notification_log_subscriber: Any = None
     dashboard: Any = None
 
     if web:
@@ -1278,10 +1279,19 @@ async def run_workflow_async(
         verbose_log(f"Agents: {len(config.agents)}")
 
         # Start JSONL event log subscriber (always-on structured diagnostics)
-        from conductor.engine.event_log import EventLogSubscriber
+        from conductor.engine.event_log import EventLogSubscriber, NotificationLogSubscriber
 
         event_log_subscriber = EventLogSubscriber(config.workflow.name)
         emitter.subscribe(event_log_subscriber.on_event)
+
+        # Dedicated notifications.jsonl for external tooling that only cares
+        # about domain notifications, not execution telemetry. Share the same
+        # run_id so the two files have aligned filename stems.
+        notification_log_subscriber = NotificationLogSubscriber(
+            config.workflow.name,
+            run_id=event_log_subscriber.run_id,
+        )
+        emitter.subscribe(notification_log_subscriber.on_event)
 
         # Subscribe console output to the event emitter
         console_subscriber = ConsoleEventSubscriber()
@@ -1466,6 +1476,11 @@ async def run_workflow_async(
         if event_log_subscriber is not None:
             event_log_subscriber.close()
             _verbose_console.print(f"[dim]Event log written to: {event_log_subscriber.path}[/dim]")
+        if notification_log_subscriber is not None:
+            notification_log_subscriber.close()
+            _verbose_console.print(
+                f"[dim]Notifications written to: {notification_log_subscriber.path}[/dim]"
+            )
 
         # Report log file path to stderr and close file logging
         if log_file is not None and _file_console is not None:
@@ -1748,6 +1763,7 @@ async def resume_workflow_async(
     # Always create event emitter and JSONL log subscriber (parity with run)
     emitter = WorkflowEventEmitter()
     event_log_subscriber: Any = None
+    notification_log_subscriber: Any = None
     dashboard: Any = None
 
     try:
@@ -1907,6 +1923,16 @@ async def resume_workflow_async(
                 existing_run_id=cp.run_id or None,
             )
             emitter.subscribe(event_log_subscriber.on_event)
+
+            # Dedicated notifications.jsonl (parity with run). Shares the
+            # run_id stem with the event log so the two files line up.
+            from conductor.engine.event_log import NotificationLogSubscriber
+
+            notification_log_subscriber = NotificationLogSubscriber(
+                config.workflow.name,
+                run_id=event_log_subscriber.run_id,
+            )
+            emitter.subscribe(notification_log_subscriber.on_event)
 
             # Subscribe console output to the event emitter (parity with run)
             console_subscriber = ConsoleEventSubscriber()
@@ -2068,6 +2094,11 @@ async def resume_workflow_async(
         if event_log_subscriber is not None:
             event_log_subscriber.close()
             _verbose_console.print(f"[dim]Event log written to: {event_log_subscriber.path}[/dim]")
+        if notification_log_subscriber is not None:
+            notification_log_subscriber.close()
+            _verbose_console.print(
+                f"[dim]Notifications written to: {notification_log_subscriber.path}[/dim]"
+            )
 
         # Report log file path to stderr and close file logging
         if log_file is not None and _file_console is not None:
--- a/src/conductor/config/schema.py
+++ b/src/conductor/config/schema.py
@@ -961,7 +961,7 @@ class AgentDef(BaseModel):
           reason: "{{ precheck.output.reason }}"
     """
 
-    notification: str | None = None
+    emit: str | None = None
     """For ``type=notification`` steps: the declared notification type name
     to emit. Must reference a key under ``workflow.notifications.types``.
     """
@@ -1291,9 +1291,9 @@ class AgentDef(BaseModel):
             if self.duration is not None:
                 raise ValueError("terminate agents cannot have 'duration' (only 'wait' agents do)")
         elif self.type == "notification":
-            if not self.notification:
+            if not self.emit:
                 raise ValueError(
-                    "notification agents require 'notification' "
+                    "notification agents require 'emit' "
                     "(the declared notification type name to emit)"
                 )
             if self.payload is None:
@@ -1365,11 +1365,11 @@ class AgentDef(BaseModel):
                     f"'{self.type or 'agent'}' agents cannot have 'output_type' "
                     "(only 'set' agents support output_type)"
                 )
-        # Cross-type guard: notification/payload fields only valid on notification steps
+        # Cross-type guard: emit/payload fields only valid on notification steps
         if self.type != "notification":
-            if self.notification is not None:
+            if self.emit is not None:
                 raise ValueError(
-                    f"'{self.type or 'agent'}' agents cannot have 'notification' "
+                    f"'{self.type or 'agent'}' agents cannot have 'emit' "
                     "(only type=notification steps emit notifications)"
                 )
             if self.payload is not None:
--- a/src/conductor/config/schema.py
+++ b/src/conductor/config/schema.py
@@ -6,6 +6,7 @@ workflow YAML configuration files.
 
 from __future__ import annotations
 
+import re
 from typing import Any, Literal
 
 from pydantic import (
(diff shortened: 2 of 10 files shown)


src/conductor/engine/workflow.py
```diff
@@ -2479,6 +2526,111 @@ async def _execute_loop(self, current_agent_name: str) -> dict[str, Any]:
                                 )
                             continue
 
+                        # Handle notification steps (fire-and-forget visibility)
+                        if agent.type == "notification":
```
jrob5756, line 3033: **Parallel / for-each containment is undefined.** This dispatch lives only in the sequential path; behavior inside `parallel:` or `for_each:` is unspecified and `get_agent_execution_count` won't give per-slot iteration counts (breaking `emission_id` uniqueness).

Either disallow at validation time, or thread emission through the group dispatchers + add tests.

src/conductor/engine/workflow.py
```diff
@@ -2479,6 +2526,111 @@ async def _execute_loop(self, current_agent_name: str) -> dict[str, Any]:
...
+                                    "notification_type": agent.notification,
+                                },
+                            )
+
+                            notif_config = self.config.workflow.notifications
+                            # The schema/validator both guarantee a non-None
+                            # notifications block exists when a notification step
+                            # is present, but assert defensively at runtime so
+                            # the type narrowing is explicit for callers.
+                            assert notif_config is not None  # noqa: S101
```
jrob5756: `assert` is stripped under `python -O` — this becomes a no-op and the next line dereferences `None`. Make it an explicit `raise`.

src/conductor/engine/workflow.py
```diff
@@ -2479,6 +2526,111 @@ async def _execute_loop(self, current_agent_name: str) -> dict[str, Any]:
...
+                                        "error_type": type(exc).__name__,
+                                        "message": str(exc),
+                                    },
+                                )
+                                raise
+
+                            # The domain notification itself — this is what
+                            # external tooling tails the notifications.jsonl
+                            # subscriber for.
+                            self._emit("notification", envelope)
```
jrob5756, line 3087: **Dashboard rendering for the new event types is missing.** Per your discussion with Daniel about a bottom-pane tab at the workflow level — worth landing alongside this PR (or as an immediate follow-up) so `--web` users actually see emissions. Currently the four new event types (`notification_started`, `notification`, `notification_completed`, `notification_failed`) and the new step type aren't handled in `workflow-store.ts` or the Cytoscape view.

src/conductor/engine/workflow.py
```diff
@@ -2479,6 +2526,111 @@ async def _execute_loop(self, current_agent_name: str) -> dict[str, Any]:
...
+                                    "notification_type": agent.notification,
+                                    "emission_id": envelope["emission_id"],
+                                    "schema_id": envelope["schema_id"],
+                                },
+                            )
+
+                            # Store an empty dict so routes can fire without
+                            # exposing notification internals as agent output
+                            # (consistent with the "no response flow" contract).
+                            self.context.store(agent.name, {})
```
jrob5756, line 3102: Storing `{}` means `{{ announce_ready.output.x }}` downstream silently renders empty under `accumulate` mode — quiet footgun for authors who confuse a notification with a regular agent. Either document the behavior in AGENTS.md, or store a sentinel that lets templates fail loudly.

src/conductor/cli/run.py
```diff
@@ -1775,6 +1791,16 @@ async def resume_workflow_async(
             )
             emitter.subscribe(event_log_subscriber.on_event)
 
+            # Dedicated notifications.jsonl (parity with run). Shares the
+            # run_id stem with the event log so the two files line up.
+            from conductor.engine.event_log import NotificationLogSubscriber
+
+            notification_log_subscriber = NotificationLogSubscriber(
```
jrob5756, line 1931: Quick confirmation: on resume the dashboard already replays the main event log (`dashboard.replay_events_from_jsonl` at line 1823), and `notification` events flow through the same emitter — so old emissions *will* render in the dashboard on resume. The separate `.notifications.jsonl` is fresh per resume, which is fine for external tailers given the stable `emission_id` dedup. No action needed; flagging only to confirm the behavior is intentional.

src/conductor/executor/notification.py
```diff
@@ -0,0 +1,235 @@
...
+    becomes a list. Falls back to the raw value if parsing fails or the
+    value is already non-string (e.g. a dict literal passed through).
+    """
+    if field_type == "string":
+        return value
+    if not isinstance(value, str):
+        return value
+    try:
+        return json.loads(value)
+    except (json.JSONDecodeError, ValueError):
```
jrob5756, line 121: Silent fallback to the raw string means a bad template (e.g. `"3.14abc"` for a `number` field) fails one frame later with "got str" rather than "failed to coerce 'X' as number". Better author UX to raise here with the offending value.

src/conductor/executor/notification.py
```diff
@@ -0,0 +1,235 @@
...
+            "schema_id": schema_id,
+            "notification_type": type_name,
+            "namespace": namespace,
+            "version": type_def.version,
+            "run_id": run_id,
+            "workflow": workflow_name,
+            "source_agent": agent.name,
+            "subworkflow_path": list(subworkflow_path),
+            "correlation": dict(correlation),
+            "workflow_metadata": dict(workflow_metadata),
```
jrob5756: **Nit:** `workflow_metadata` feels redundant in the envelope. Consumers already get `run_id` + `workflow` + `subworkflow_path` and can join against the event log if they need more. Splatting the whole metadata dict into every emission bloats the wire format and couples the notification schema to whatever ad-hoc keys workflows happen to use. Consider dropping it (or opt-in allowlist).

src/conductor/config/schema.py
```diff
@@ -671,6 +795,18 @@ class AgentDef(BaseModel):
           effort: high
     """
 
+    notification: str | None = None
```
jrob5756: **Nit:** `type: notification` + `notification: pr_ready` reads redundant. Siblings use `script`→`command`, `human_gate`→`dialog`. Something like `emit:` or `event:` would parse more naturally — easier to rename now than after it ships.

jrob5756 (commented):
Nice work. I have a few questions/comments below

PolyphonyRequiem:
Rebased \eat/notifications\ onto v0.1.18 for local polyphony dogfood testing. One mechanical conflict in \schema.py\ (combined \AgentDef.type\ Literal enum to include \set/terminate/wait\ from v0.1.18 plus \
otification\ from this branch, plus adjacent field and validator blocks). Conflict was fully mechanical — no overlapping logic. Pushed to [PolyphonyRequiem/conductor feat/notifications](https://github.com/PolyphonyRequiem/conductor/tree/feat/notifications). New HEAD: 7d46793.

src/conductor/engine/workflow.py
PolyphonyRequiem: Applied. Added validator disallowance for `type: notification` inside both parallel groups and `for_each` inline agents, matching the existing guard pattern for `script`, `wait`, `terminate`, and `workflow` steps. Error messages direct authors to place the notification step in the main agent list and route to it from the group's `routes:` block instead. Threading dispatch through the group executors (iteration-scoped `emission_id` uniqueness) is a follow-up; the conservative static guard is in commit 27006af.

src/conductor/engine/workflow.py
PolyphonyRequiem: Applied. Replaced `assert notif_config is not None` with an explicit `raise ExecutionError(...)` (commit 27006af). The `ExecutionError` matches the pattern used for other internal invariant violations in `workflow.py`.

src/conductor/engine/workflow.py
PolyphonyRequiem: Acknowledged — deferring dashboard rendering as an immediate follow-up PR per the 'or as an immediate follow-up' suggestion. This PR's scope is the backend notification pipeline (YAML surface, executor, event emission, JSONL subscriber). The four new event types (`notification_started`, `notification`, `notification_completed`, `notification_failed`) and the new step type will need handlers in `workflow-store.ts` and the Cytoscape view; that's a standalone frontend change.

src/conductor/engine/workflow.py
PolyphonyRequiem: Applied documentation approach. Added a note to AGENTS.md under the Notifications section documenting that notification steps store `{}` in the workflow context and that downstream `accumulate`-mode templates referencing `step_name.output.field` will silently render empty — this is intentional fire-and-forget behavior, and authors who need a value to flow back should use a regular agent step instead. Commit 27006af.

src/conductor/cli/run.py
PolyphonyRequiem: Confirmed — that's intentional. Notification events flow through `WorkflowEventEmitter` → `EventLog` → `notifications.jsonl` subscriber on the same path as all other events. The `replay_events_from_jsonl` call at resume replays everything from the main event log, so past emissions will be visible in the dashboard tab on resume. The separate `.notifications.jsonl` is a fresh-per-resume streaming target for external tailers; `emission_id` dedup in consumers handles the fresh start.

src/conductor/executor/notification.py
PolyphonyRequiem: Applied. `_coerce_rendered` now raises `ValidationError` on `json.loads` failure instead of silently returning the raw string, and the call site in `build_envelope` catches `ValidationError` from the coerce and re-raises with the field name for context (e.g. `Payload field 'pr_id' for notification step 'bad': Failed to coerce rendered value 'not-a-number' to declared type 'number'`). Commit 27006af.

src/conductor/executor/notification.py
PolyphonyRequiem: Applied. Dropped `workflow_metadata` from the `build_envelope` signature and the envelope dict. Consumers already have `run_id`, `workflow`, and `subworkflow_path` to join against the event log for any workflow-level metadata they need. Commit 27006af.

src/conductor/config/schema.py
PolyphonyRequiem: Applied, using `emit:`. The field on `AgentDef` is now `emit: str | None` (schema.py, validator.py, executor, workflow.py, tests, examples, AGENTS.md all updated). The YAML surface reads `type: notification` + `emit: pr_ready` which parses naturally as 'this step emits the pr_ready notification type'. Agreed that now is the right time — commit 27006af.

PR #211, opened 2026-05-19
fix(cli): gate remaining dashboard URL prints behind is_verbose() for --silent compliance

wjgong001:
## Description

Follow-up to PR #203 (merged, by @sjh9714). PR #203 fixed the web-bg startup dashboard URL leak in `src/conductor/cli/app.py`, but the `--silent` flag is supposed to suppress ALL dashboard URL output, not just the startup print.

## Remaining fixes

This PR gates the remaining dashboard URL prints behind `is_verbose()`:

1. **`src/conductor/cli/run.py` — `run_workflow_async` function (line 1174)**: The `--silent` flag should suppress this dashboard URL print during foreground --web runs, not just web-bg mode (#203's scope).

2. **`src/conductor/cli/run.py` — `resume` function (line 1832)**: Same issue in the resume code path — dashboard URL printed regardless of --silent.

3. **`src/conductor/cli/app.py` — `_run_replay` function (line 994)**: Replay dashboard URL leaked even in --silent mode.

4. **`src/conductor/cli/run.py` — "Workflow complete. Dashboard still running at..." shutdown messages (lines 1321, 1884)**: Both the `run` and `resume` functions printed the dashboard URL during shutdown regardless of --silent mode.

## Testing

Existing tests from #203 should continue to pass. The fix follows the same pattern as #203: lazy-import `is_verbose()` from `app.py` and gate the print call.

Co-authored-by: sjh9714 <sjh9714@users.noreply.github.com>

PR #223, opened 2026-05-21
fix(cli): make _verbose_console silent-aware and gate replay prints (#209)

### LINKED ISSUE #209: `--silent` still prints dashboard URL in foreground `--web` and `replay`
## Summary

Follow-up to #201 / #203, which fixed `--silent --web-bg` for both `run` and `resume`. Three other dashboard-URL prints share the same shape and still violate the `--silent` contract ("No progress output. Only JSON result on stdout."):

1. `src/conductor/cli/app.py:990` — `replay` command:
   ```python
   console.print(f"\n[bold green]▶ Replay dashboard:[/] {dashboard.url}\n")
   console.print("[dim]Press Ctrl+C to exit[/dim]\n")
   ```
2. `src/conductor/cli/run.py:1173` — foreground `run --web`:
   ```python
   _verbose_console.print(f"[bold cyan]Dashboard:[/bold cyan] {dashboard.url}")
   ```
3. `src/conductor/cli/run.py:1685` — foreground `resume --web`:
   ```python
   _verbose_console.print(f"[bold cyan]Dashboard:[/bold cyan] {dashboard.url}")
   ```

Note that `_verbose_console = Console(stderr=True, highlight=False)` (`run.py:33`) is **not** silent-aware — it's a plain stderr Console that doesn't consult `verbose_mode`, despite the name suggesting otherwise.

## Repro

```bash
# foreground --web
conductor --silent run examples/simple-qa.yaml --web 2> /tmp/stderr.txt
# Expected: empty (apart from the JSON result on stdout)
# Actual: 'Dashboard: http://127.0.0.1:XXXXX' on stderr

# replay
conductor --silent replay /path/to/log.jsonl 2> /tmp/stderr.txt
# Expected: empty
# Actual: '▶ Replay dashboard: http://127.0.0.1:XXXXX' and 'Press Ctrl+C to exit' on stderr
```

## Impact

Same as #201 — stderr-only, no JSON contamination on stdout, but scripted/CI consumers that suppress stderr expecting silence will still see output bleed through.

## Suggested fix

Gate each `console.print` (or `_verbose_console.print`) call by `is_verbose()`, matching the pattern landed in #203:

```python
from conductor.cli.app import is_verbose

if is_verbose():
    console.print(f"[bold cyan]Dashboard:[/bold cyan] {dashboard.url}")
```

For the `run.py` sites, consider either making `_verbose_console` itself silent-aware (preferred, fixes the class of bug at the source — the name implies it already is) or adding explicit `is_verbose()` gates at the call sites. The former is cleaner but has wider blast radius and should be done carefully since `_verbose_console` is used throughout `run.py` for progress logging that's already gated elsewhere — auditing each call site is recommended.

For `replay`, the simple gate suffices since the URL is the primary purpose of the command (a silent caller is unusual but should still be honored).

## Acceptance criteria

- `conductor --silent run <wf> --web 2>&1 >/dev/null` produces no stderr from conductor (post-engine output may still occur if the workflow itself writes; that's separate).
- `conductor --silent resume <wf> --web 2>&1 >/dev/null` ditto.
- `conductor --silent replay <log>` produces no stderr.
- Regression tests follow the pattern from #203 (`test_silent_web_bg_suppresses_dashboard_output` and friends).

jrob5756:
Follow-up to #211, closes #209.

## Problem

PR #211 gated the three foreground dashboard URL prints behind `is_verbose()` but left several other stderr leaks in place. Under `--silent` ("No progress output. Only JSON result on stdout."):

* `app.py:_run_replay` still printed `Press Ctrl+C to exit` and `Replay stopped` unconditionally.
* `run.py` had ~12 `_verbose_console.print` call sites that bypassed `is_verbose()` — warnings (dashboard failed to start, log-file open failure, workflow-hash mismatch), `Press Esc to interrupt`, `Event log written to ...`, `Log written to ...`, and `_print_resume_instructions` (printed on failure).

## Fix

**Fix at the source.** Subclass `Console` for `_verbose_console` so every `.print(...)` is a no-op when `is_verbose()` is False. This:
* Aligns the implementation with the long-misleading name (`_verbose_console` is now actually verbose-aware).
* Removes the per-call-site audit burden — adding new `_verbose_console.print(...)` calls in `run.py` will now respect `--silent` automatically.
* Is a no-op for the already-gated helpers (`verbose_log`, `verbose_log_agent_*`, the four PR #211 dashboard prints) — their internal `is_verbose()` checks become belt-and-suspenders.

The app-wide `console` in `app.py` is **intentionally NOT** made silent-aware because it carries real error messages (`print_error` etc.). The two remaining replay prints in `_run_replay` are gated per-call instead.

## Acceptance criterion

```
$ conductor --silent replay /tmp/log.jsonl 2>/tmp/stderr  &
$ wc -c /tmp/stderr
0 /tmp/stderr
```

Verified end-to-end. Pre-fix, the same invocation leaked `Press Ctrl+C to exit` to stderr.

## Tests

* `TestSilentAwareConsole` (3 tests) — verifies the subclass mechanism and that the module-level `_verbose_console` instance uses it. The `test_module_verbose_console_is_silent_aware` test guards against a future refactor accidentally swapping back to a plain `Console` (the original regression).
* `TestReplaySilentCompliance` (2 tests) — verifies `conductor --silent replay` produces no stderr, and that without `--silent` the messages still appear (guard against an over-eager gate).

Mocks `ReplayDashboard.start/stop` and patches `asyncio.Event` to raise `CancelledError`; otherwise the tests would hang on the `await asyncio.Event().wait()` the replay command uses to keep the dashboard alive.

All 397 existing `tests/test_cli` tests pass.

## Rubber-duck pass

Consulted before implementation. Key findings adopted:
* Test plan would have hung on infinite dashboard waits — solved via the `asyncio.Event` patch above.
* `Press Esc` only fires for non-`--web` interactive runs, so a `--silent run --web` test couldn't exercise it — the silent-aware subclass tests at the unit level cover all 12 `run.py` call sites at once instead.
* Confirmed no existing tests assert these messages appear in non-silent mode (no tests broken).

### DIFF
--- a/src/conductor/cli/app.py
+++ b/src/conductor/cli/app.py
@@ -993,7 +993,7 @@ def replay(
         await dashboard.start()
         if is_verbose():
             console.print(f"\n[bold green]▶ Replay dashboard:[/] {dashboard.url}\n")
-        console.print("[dim]Press Ctrl+C to exit[/dim]\n")
+            console.print("[dim]Press Ctrl+C to exit[/dim]\n")
 
         try:
             await asyncio.Event().wait()
@@ -1005,7 +1005,8 @@ def replay(
     try:
         asyncio.run(_run_replay())
     except KeyboardInterrupt:
-        console.print("\n[dim]Replay stopped.[/dim]")
+        if is_verbose():
+            console.print("\n[dim]Replay stopped.[/dim]")
 
 
 @app.command()
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -30,8 +30,28 @@ from conductor.providers.registry import ProviderRegistry
 if TYPE_CHECKING:
     from conductor.events import WorkflowEvent
 
-# Verbose console for logging (stderr)
-_verbose_console = Console(stderr=True, highlight=False)
+
+# Verbose console for logging (stderr).
+#
+# This subclass enforces the global --silent contract ("No progress output.
+# Only JSON result on stdout.") at the source: every ``_verbose_console.print``
+# call in this module becomes a no-op when ``is_verbose()`` is False, so
+# individual call sites do not each have to remember to gate themselves
+# (see issue #209, follow-up to #203/#211). File logging is unaffected —
+# ``_file_console`` is a separate Console wired up by ``init_file_logging``.
+class _SilentAwareConsole(Console):
+    """``Console`` that honors ``--silent`` at the print level."""
+
+    def print(self, *args: Any, **kwargs: Any) -> None:
+        # Lazy import to avoid the cli.run -> cli.app import cycle at module
+        # load time, and to pick up the current ContextVar value per-call.
+        from conductor.cli.app import is_verbose
+
+        if is_verbose():
+            super().print(*args, **kwargs)
+
+
+_verbose_console = _SilentAwareConsole(stderr=True, highlight=False)
 
 # File console for file logging (None when not active)
 _file_console: Console | None = None
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -34,24 +34,41 @@ if TYPE_CHECKING:
 # Verbose console for logging (stderr).
 #
 # This subclass enforces the global --silent contract ("No progress output.
-# Only JSON result on stdout.") at the source: every ``_verbose_console.print``
-# call in this module becomes a no-op when ``is_verbose()`` is False, so
+# Only JSON result on stdout.") at the source: every ``.print()`` call on a
+# ``_SilentAwareConsole`` becomes a no-op when ``is_verbose()`` is False, so
 # individual call sites do not each have to remember to gate themselves
 # (see issue #209, follow-up to #203/#211). File logging is unaffected —
 # ``_file_console`` is a separate Console wired up by ``init_file_logging``.
+#
+# Scope: only ``.print()`` is gated. Other Rich ``Console`` output methods
+# (``.log()``, ``.rule()``, ``.status()``, ``.print_json()``, etc.) would
+# silently bypass ``--silent`` if used on this instance. All current call
+# sites in this module use only ``.print``; if you introduce a new one,
+# either route it through ``.print`` or extend this subclass.
 class _SilentAwareConsole(Console):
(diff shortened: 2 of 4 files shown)


jrob5756:
Addressed PR review feedback (commit bcc29a5):

**Critical**
- ✅ Covered the `KeyboardInterrupt` branch at `app.py:1007-1009` — the previous tests only exercised the inner `CancelledError` path. Added `test_silent_replay_suppresses_keyboardinterrupt_message` + verbose counterpart via `patch("asyncio.run", side_effect=KeyboardInterrupt())`. Also more refactor-robust than the inner-path mock.

**Important**
- ✅ Locked `_SilentAwareConsole.__init__` to force `stderr=True` (prevents a future second instance from corrupting the stdout JSON contract).
- ✅ Tests now assert on `result.stderr` (not the combined `result.output`) — matches the test names and catches a hypothetical stdout-leak regression.

**Suggestions**
- ✅ Added `test_quiet_replay_prints_dashboard_messages` to lock in the `app.py:204` contract that `--quiet` (MINIMAL) does NOT suppress.
- ✅ Class-level comment on `_SilentAwareConsole` warns that only `.print()` is gated (`.rule`/`.log`/`.status`/`.print_json` would silently bypass — latent footgun for future contributors).

**Verification**
- `conductor --silent replay <log>`: 0 bytes on stderr ✓
- `conductor --quiet replay <log>`: dashboard URL + Ctrl+C hint visible ✓
- 400 `tests/test_cli` tests pass (was 397; +3 net new). Lint/format/typecheck clean.

**Not adopted (with reason)**
- Inline `# silent:` comments at `app.py:996`/`:1008` — comment-noise vs. marginal future-proofing.
- Removing the redundant outer `is_verbose()` guards in `run.py` helpers — out of scope; 16+ existing tests in `test_logging.py` patch `_verbose_console` with a plain `Console` and rely on the outer guards.
- Overriding `.log`/`.rule`/etc. on `_SilentAwareConsole` — YAGNI; the warning comment is sufficient defense (and a future test failure would surface the gap immediately).
- "Tighten 'in this module' phrasing" — cosmetic.

PR #227, opened 2026-05-21
RFC: first-class on_error routing (brainstorm)

PolyphonyRequiem:
RFC adding an engineering brief that proposes teaching conductor's `routes:` table about errors.

**Read order:** the TL;DR + thesis + top decisions + critical open questions at the top of the doc are the only parts that need eyes. Everything below the `# Deep dives` fold is the design at full depth — skim or skip.

## What it proposes

A node raises a typed error envelope; outgoing routes match on either success or error; four route actions (`to` / `retry` / `halt` / `propagate`) cover the full control-flow surface; sub-workflow errors propagate by default; provider transport exhaustion surfaces as a routable kind.

``yaml
routes:
  - to: next_node                                  # success path (unchanged)
  - on_error: external.git.drift
    retry: { max: 5, backoff: exponential, initial_seconds: 2 }
  - on_error: external.git.drift
    to: drift_recovery_gate                        # post-exhaustion fallback
  - on_error: provider.exhausted
    halt: { message: "model down: {{ error.message }}" }
  - on_error: "*"
    propagate: true                                # uncaught -> parent workflow
``

## Why

Polyphony has ~40 `human_gate` nodes across its 15 workflows whose sole purpose is to absorb script-level failures (`{success: false}` lands here, human picks retry / skip / abort). They exist because conductor has no other primitive that can catch a node-level failure. With this brief shipped, most of those collapse to one or two route lines on the failing node.

## What's in this PR

Just the brainstorm doc — `docs/projects/error-routing/on-error-routing.brainstorm.md`. No schema changes, no code.

## What I'd like back

Not a yes/no on the whole thing. Specifically:

1. A read on the three **critical open questions** flagged at the top of the brief (post-retry-exhaustion semantics, route-actions vs. smaller alternative, reserved-kind namespace).
2. A read on the thesis: is *"routes table as universal control flow"* the right north star for conductor, or am I projecting?
3. If yes-in-principle, your preference on phasing.

Submitted from polyphony's architecture/rationalization track (AB#3257, parent epic AB#3253).

### DIFF
--- a/docs/projects/error-routing/on-error-routing.brainstorm.md
+++ b/docs/projects/error-routing/on-error-routing.brainstorm.md
@@ -1,0 +1,786 @@
+# First-Class `on_error` Routing
+
+**For:** conductor engineering (hi Jason)
+**From:** polyphony — a workflow-heavy downstream consumer (15 workflows, 296 nodes)
+**Status:** Brainstorm / RFC. Ready for design after a few decisions land.
+**Polyphony tracking:** AB#3257 (parent epic: AB#3253)
+
+---
+
+## TL;DR
+
+Teach conductor's `routes:` table about errors.
+
+A node raises a typed error envelope; outgoing routes match on either
+success or error; four route actions (`to`, `retry`, `halt`, `propagate`)
+cover the full control-flow surface; sub-workflow errors propagate by
+default. Replaces the per-author `{success: bool, error: "..."}` +
+branch + `human_gate` idiom that workflow authors fall back to today.
+
+```yaml
+routes:
+  - to: next_node                                  # success path (unchanged)
+  - on_error: external.git.drift
+    retry: { max: 5, backoff: exponential, initial_seconds: 2 }
+  - on_error: external.git.drift
+    to: drift_recovery_gate                        # post-exhaustion fallback
+  - on_error: provider.exhausted
+    halt: { message: "model down: {{ error.message }}" }
+  - on_error: "*"
+    propagate: true                                # uncaught -> parent workflow
+```
+
+Everything in the deep dives below this fold expands one of those lines.
+
+---
+
+## Thesis
+
+> **Routes are conductor's expressive heart. Make them complete.**
+
+Today `routes:` is "where do I go next when this node succeeds?" Promote
+it to "where does control flow next, regardless of what happened?" — and
+the rest of the failure-handling vocabulary (retry, halt, propagate)
+falls into the same table as route actions instead of needing four
+separate schema concepts.
+
+A smaller alternative (sibling fields for retry, routes unchanged) is
+described in the deep dive; same end-state behavior, two surfaces
+instead of one. This brief recommends the unified shape.
+
+---
+
+## What this enables (the payoff)
+
+Polyphony has ~40 `human_gate` nodes across its workflows whose sole
+purpose is to absorb script-level failures (`{success: false}` lands
+here, human picks retry / skip / abort). They exist because conductor
+has no other primitive that can catch a node-level failure.
+
+With this brief shipped:
+
+- Most of those 40 gates collapse to one or two route lines on the
+  failing node. The handful that genuinely want a human-in-the-loop
+  decision stay, but as targets of an `on_error` route — not as
+  hand-rolled discriminator branches.
+- Sub-workflow failures stop silently wedging apex runs. Today a child
+  workflow halting at one of these gates appears to the parent as
+  "didn't return"; with propagation, the parent's routes table sees it.
+- Provider outages stop being silent halts. `on_error: provider.exhausted`
+  lets workflows degrade gracefully — switch model, wait, escalate.
+
+The point isn't error handling per se. It's that conductor gains *one*
+universal control-flow primitive in the place authors already look for
+control flow.
+
+---
+
+## Top decisions baked into this proposal
+
+These are the choices the brief makes opinionatedly. Push back on any of
+them and the design adjusts cleanly.
+
+1. **`routes:` is the surface, not a new `on_failure:` block.** Errors
+   become a matcher on the existing route table (`on_error: <kind>`).
+   The alternative — a parallel block like `on_failure: [{when, action}]`
+   — is described in the deep dive and explicitly not recommended.
+
+2. **Four route actions, exactly one per entry: `to` / `retry` / `halt` /
+   `propagate`.** Today `to:` is implicit; promote it. The other three
+   subsume what would otherwise need separate sibling schemas. This is
+   the bigger swing; the smaller alternative keeps just `to:` and moves
+   retry to a sibling field. Both shapes detailed in the deep dive.
+
+3. **Typed error envelope is the load-bearing primitive.**
+   `{conductor_error: true, kind, message, details}` — flat dotted-string
+   kinds, equality match only in v1. Scripts write to a runtime-supplied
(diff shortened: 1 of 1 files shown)


PolyphonyRequiem:
Pushed a substantive update to D1 (the script-side error contract). Top-level shape is unchanged; the change is about engine reach.

**What changed and why**

A reviewer (Daniel) pushed back: conductor's script executor is engine-agnostic (`asyncio.create_subprocess_exec`, no language assumptions), and the original draft leaned on a pwsh `Write-ConductorError` cmdlet as the headline ergonomic surface. That was an accident of the polyphony codebase being pwsh-heavy, not a design choice — and it pushed the cost of adoption onto every other engine.

The fix is to make the contract language-neutral, with prior art:

- **Contract:** conductor sets `\` to a temp-file path; the script writes the envelope JSON to that path and exits 0. Three lines in pwsh, bash, python, node, or dotnet (side-by-side examples in D1.2).
- **Prior art:** GitHub Actions explicitly migrated from stdout sentinels (`::set-output::`) to env-var-files (`\`, `\`, `\`) for these exact reasons (stdout pollution, parser fragility, security). We're walking the same path. Cross-platform behavior is proven at scale.
- **Helpers ship as sugar, not as the contract.** `helpers/error/` directory with one file per common engine, all optional. The contract is what you can implement in 60 seconds from the spec.

**Also in this push**

- **D1.6 'Where do kinds come from?'** — an authorship table making it explicit that conductor never infers kinds from exit codes or stderr patterns. Without intentional authorship at the failure site, all you can know is `internal.script_error`.
- **D1.7 optional `raises: [kind]` declaration on nodes** — opt-in self-doc that powers a lint pass (validator can flag `on_error: x.z` against a node that declares `raises: [x.y]`) and a runtime undeclared-kind check (kinds raised but not declared become `internal.undeclared_kind`). Strictly optional; existing nodes need no change.
- **D1.8 wrapping third-party CLIs** — concrete bash and pwsh examples of translating `git fetch` / `npm install` exit codes into typed envelopes at the workflow author's boundary.
- New non-goal: **no inference of kinds from runtime signals.**
- Touch points updated: `tempfile.mkstemp(delete=False)` for the path so Windows file-locking doesn't bite; conductor reads only after subprocess exit so no read/write overlap.
- Phase 1 acceptance criteria expanded: cross-engine tests (pwsh-on-Windows, bash-on-Linux, python on both); `raises:` lint and runtime checks.

**Net effect:** the contract is now provably the simplest thing that could work for any script engine, the prior art is named, and the ergonomic helpers are positioned correctly (optional, not load-bearing).

Commit: 02eace8

PolyphonyRequiem:
## Design update: narrowed Phase 1 + proposed roadmap

@jrob5756 — after rebasing the implementation against current `main` and working through the review feedback, I substantially narrowed the first implementation slice. The updated implementation is in draft PR #229: https://github.com/microsoft/conductor/pull/229

### What changed from the RFC's original Phase 1

The original Phase 1 attempted to establish most of the eventual framework at once: script and agent raises, synthesized kinds for nonzero exits, runtime `raises:` enforcement, new CLI failure output/exit codes, bundled helpers, and groundwork for propagation.

The revised Phase 1 proves one complete vertical slice instead: **a top-level script intentionally publishes a typed domain failure, the existing routes table selects a deterministic recovery path, and checkpoint/resume state remains correct.**

Key decisions:

- **Script-only for now.** `on_error` and `raises` are accepted only on top-level `type: script` steps. Provider agents, groups, and sub-workflows fail validation rather than accepting handlers that cannot run yet.
- **One language-neutral transport.** Conductor injects a private `CONDUCTOR_ERROR_OUT` file path; the script writes `{kind, message, details}` as UTF-8 JSON. No language-specific helper libraries ship in the wheel.
- **No inference from process failure.** A missing file means no typed failure. Ordinary nonzero exits remain ordinary script output and existing `exit_code` routes are unchanged.
- **File presence is authoritative.** A valid file routes the typed failure regardless of exit code. Malformed/unreadable envelopes fail closed as `internal.script_error_transport`.
- **Engine namespaces are protected.** Scripts cannot publish `internal.*`, `provider.*`, `subworkflow.*`, or `retry.*` kinds.
- **Deterministic route buckets.** Routes without `on_error` handle success. Routes with `on_error` handle typed failure and support an exact kind, a list, or final `true` catch-all.
- **`raises:` is documentation/lint only.** It does not opt into synthesis, constrain runtime kinds, or rewrite undeclared kinds. Catch-all handlers see the original kind.
- **Partial failure output remains usable.** Typed routing occurs before validating the script's success `output:` schema, so a handler can inspect partial stdout/stderr alongside `<step>.error`.
- **Checkpoint precedence is explicit.** Handled failures commit output + envelope without creating a failure checkpoint. If the handler then fails, its checkpoint includes the prior envelope. Unhandled failures checkpoint the original script before committing that failed execution.
- **Existing control-flow ownership is unchanged.** Provider retries remain provider-owned; failed explicit terminate remains terminal/non-checkpointed; `SubworkflowTerminatedError` keeps its current parent behavior.

### Does this Phase 1 deliver user value by itself?

I think **yes, for a focused but real class of workflows**.

Today, script authors often encode failures as ordinary data (`success: false`, magic exit codes, stderr parsing) and then add generic routing or human gates whose only job is to interpret that data. Phase 1 gives those workflows a first-class way to distinguish a domain failure from process status and route directly to recovery while retaining partial diagnostics.

Concrete value:

- A script can distinguish `external.git.drift` from authentication failure or an unknown failure without teaching Conductor language-specific parsing rules.
- Workflows can replace some failure-as-data plumbing and “catch this script failure” human gates with exact or catch-all routes.
- Existing script workflows gain no behavioral changes until they intentionally write the envelope file.
- Recovery paths become visible and statically validated in the same routes table as successful control flow.
- A failed recovery resumes at the recovery handler with the original typed failure still available, rather than rerunning or losing the failure context unpredictably.

The value is intentionally narrower than the full RFC. Phase 1 does **not** yet solve provider exhaustion, retry policy in the route table, or propagation through sub-workflows/groups. The question is whether the script-only slice is useful and coherent enough to merge as a foundation rather than waiting for all of those surfaces.

### Proposed roadmap after Phase 1

**Phase 2 — propagation and additional failure producers**

- Define the sub-workflow envelope boundary and default propagation behavior.
- Define parallel/for-each aggregation and failure-mode interaction.
- Surface selected provider terminal failures through the same envelope, including provider exhaustion.
- Decide whether model-agent failures need an explicit discriminator or should initially be limited to engine/provider-authored kinds.

**Phase 3 — route actions and policy**

- Add retry as an explicit route action, including exhaustion/fallback ordering.
- Add explicit `halt` and `propagate` actions where they improve on today’s implicit unmatched behavior.
- Extend observability only after the state-transition semantics are stable.

I would keep the single envelope and route-bucket semantics stable across phases; later work should add failure producers and route actions rather than redefine the Phase 1 transport.

### Input requested

1. Is script-only typed routing independently valuable enough to merge, or should #229 remain blocked on another producer/boundary?
2. Does optional documentation-only `raises:` strike the right balance, or should it be omitted entirely until it can support stronger tooling?
3. For Phase 2, which is more important first: sub-workflow/group propagation, provider terminal failures, or retry actions?
4. Are the checkpoint and terminate precedence rules above consistent with how you expect Conductor workflows to recover?
5. Any concern with reserving engine namespaces while leaving author-owned kinds otherwise open?

PR #229 is intentionally back in draft while we get a read on those decisions. Its branch is rebased, the narrowed design is documented, and CI is green.

Issue #221, opened 2026-05-21
Add a `type: set` step for binding computed values into the context


jrob5756:
## Summary

Add a step type whose only job is to evaluate one or more Jinja2 expressions and bind the results as a named entry in the workflow context. No LLM call, no subprocess, no I/O.

## Motivation

Workflows often need a tiny bit of state derivation between agents:

- Combine inputs: `repo_slug = "{{ workflow.input.org }}/{{ workflow.input.repo }}"`
- Compute a flag from prior output: `is_breaking = research.output.severity in ["high", "critical"]`
- Derive a default: `branch_name = workflow.input.branch or "main"`
- Normalize a value once so every downstream prompt uses the same form
- Stash a constant referenced from many later prompts

Today the options are all awkward:

1. **Inline the expression in every consuming agent's prompt** — duplicated logic, drifts, painful to change.
2. **`type: script` with `echo`** — subprocess overhead for what should be a single line of templating.
3. **Put it in an LLM prompt** — wastes a model call on something deterministic.

A tiny "set" step makes this explicit, deterministic, free, and visible in the dashboard.

## Proposed shape

Single binding:

```yaml
agents:
  - name: compute_target
    type: set
    value: "{{ workflow.input.org }}/{{ workflow.input.repo }}"
    # output available as: compute_target.output (a string)
```

Multiple named bindings:

```yaml
agents:
  - name: derive_flags
    type: set
    values:
      is_breaking: "{{ research.output.severity in ['high', 'critical'] }}"
      target_branch: "{{ workflow.input.branch or 'main' }}"
      effective_model: "{{ workflow.input.model or 'claude-sonnet-4-5' }}"
    # outputs available as: derive_flags.output.is_breaking, .target_branch, .effective_model
```

Exactly one of `value:` / `values:` must be present.

### Behavior

- Expressions render against the full current context with the standard Jinja2 environment.
- Result type is detected: looks like JSON / list / dict / bool / number → parsed; otherwise string. Optional `output_type:` to force a specific type when detection is ambiguous.
- The step's output is queryable like any other agent's output (`step_name.output` or `step_name.output.field`).
- Routes can be attached just like any other step (e.g. set a flag, then route on it).
- Effectively free — emits standard `agent_started` / `agent_completed` events so the dashboard timeline stays consistent.

## Why now

- Tiny implementation, very high "look how nice this YAML reads now" payoff.
[...]

PR #232, opened 2026-05-26
fix: external workflow friction - minimal evidence-anchored fixes

lucioctinoco:
# fix: external workflow friction — minimal evidence-anchored fixes

**Branch:** `fix/external-workflow-friction` → `main`
**Type:** Bug fixes (3) + documentation
**Risk:** Low — two narrow bug fixes, one CLI pre-fork validation, one new docs section. No schema or API changes.
**Plan:** [docs/projects/usability-features/external-workflow-friction.plan.md](docs/projects/usability-features/external-workflow-friction.plan.md)
**Brainstorm:** [docs/projects/usability-features/external-workflow-friction.brainstorm.md](docs/projects/usability-features/external-workflow-friction.brainstorm.md)

## Background

A single external contributor running two real-world workflows against Conductor v0.1.16 hit seven failed runs before reaching a successful end-to-end execution. The companion brainstorm identified nine candidate issues. Pre-design code review confirmed **four** of those issues are real, root-caused, and have a clear minimal fix; the remaining five either could not be reproduced from the cited evidence, propose speculative configurability, or address unobserved failure modes.

This PR ships three of the four (the fourth, scripts-inside-parallel validation, was already on main per pre-flight check). Five issues are explicitly deferred in the plan §6 with thresholds-to-reopen.

## Commits

| Commit | Issue | Plan Item |
|---|---|---|
| `505caee fix(parser): use greedy fence regex for JSON with embedded backticks` | Brainstorm #1 | Item 1 (FR-1, FR-2, FR-3) |
| `69f2dd5 fix(cli): abort --web-bg before fork when workflow has human_gate` | Brainstorm #8 | Item 4 (FR-6) |
| `4e5c07b docs: omit-output guidance, --web-bg gate constraint, brainstorm + plan` | Brainstorm #2 + docs gaps | Item 3 (FR-5) + plan/brainstorm in-tree |

## Issue 1: Fence regex strips JSON containing triple-backticks

**Failure mode:** `parse_json_output` (executor/output.py) and `_extract_json` (providers/copilot.py) both used a non-greedy fenced-block regex:

```
r"```(?:json)?\s*\n?(.*?)\n?```"
```

When an agent emitted JSON whose string values contained triple-backtick substrings (common for Markdown- or shell-snippet-bearing payloads), the first inner ` ``` ` closed the match prematurely. The extracted substring was invalid JSON, triggering parse-recovery loops and burning tokens — the exact failure observed in the brainstorm timeline.

**Fix:** switch both call sites to a greedy capture with `re.DOTALL`:

```
r"```(?:json)?\s*\n(.*)\n```"
```

With `DOTALL` the greedy `.*` terminates at the last triple-backtick in the string, which is the correct boundary. The existing first-`{`/`[` heuristic and brace-match fallback remain as final attempts; the canonical `json.loads` at the end is the unchanged failure point for genuinely malformed JSON.

**Test:** `tests/test_executor/test_output.py::test_parse_json_with_triple_backticks_inside_string` fails on `main` and passes after this commit. The malformed-input test is retained as a regression guard for the unchanged error message.

## Issue 4: --web-bg crashes silently when workflow contains human_gate

**Failure mode:** `conductor run --web-bg` (and `resume --web-bg`) forked a detached background process. When the workflow reached a `human_gate` step, `Prompt.ask()` read from the closed stdin and the child crashed with `EOFError`. The parent only saw `"Background process exited immediately with code 1"` — nothing pointed at the actual incompatibility, the `--skip-gates` workaround, or the foreground `--web` alternative.

**Fix:** `_abort_web_bg_if_human_gate()` helper in `cli/app.py` loads the workflow, walks `config.agents`, and if any `type: human_gate` is present (and `--skip-gates` is not set) aborts with this guidance message before `launch_background()` forks anything:

```
--web-bg is incompatible with workflows that contain human_gate steps
because the detached process has no stdin to prompt on.

Options:
  1. Use --web (foreground) instead of --web-bg
  2. Add --skip-gates to auto-accept the first option
  3. Remove human_gate steps from the workflow
  4. Wait for CLI gate-resolution support (planned follow-up)
```

Call sites added to both `run` and `resume` per the run/resume parity rule in `AGENTS.md`.

**Tests:** `tests/test_cli/test_web_flags.py::TestWebBgHumanGateValidation`:

- `test_web_bg_with_human_gate_aborts_before_fork`
- `test_web_bg_with_human_gate_and_skip_gates_proceeds`
- `test_resume_web_bg_with_human_gate_aborts_before_fork`

All three mock `launch_background`, run in-process via `CliRunner`, and produce no subprocess. Deterministic.

## Issue 2 (docs-only): when to declare `output:` vs omit it

**Failure mode:** the external contributor declared `output:` on a synthesizer agent that produced 80 KB of nested JSON. This injected a schema instruction that the model partially complied with, fell into parse-recovery, and burned cost. The brainstorm proposed adding an `output_mode` field; the plan §2 NG1 rejected that as API bloat — the "raw" behavior already exists as "omit `output:`". The fix is docs.

**Change:** new "Choosing whether to declare `output:`" section in `docs/workflow-syntax.md` describing the trade-off in one sentence, the two clear cases (small structured JSON → declare; prose or large JSON → omit and read `<agent>.output.result`), and the YAML for both. Cross-linked from the `output:` reference subsection.

## Additional documentation

- **`docs/cli-reference.md` `--web-bg` section** — now documents the `human_gate` incompatibility and the four supported options matching the new pre-fork validation. Closes a gap noticed while implementing Issue 4.
- **`CHANGELOG.md` [Unreleased]** — entries under `Fixed` and `Documentation` for each change.
- **Plan and brainstorm in-tree at `docs/projects/usability-features/`** — kept so future contributors can answer "why was issue X not done?" without spelunking PR history. The plan §6 table lists each deferred issue with the threshold that would justify reopening it.

## What this PR does NOT do

Plan §2 NG1-NG7 (also §6) enumerate the five brainstorm issues explicitly out of scope:

- **Issue #2 `output_mode` field** — synonym for "absence of `output:`"; docs fix (Item 3) likely sufficient.
- **Issue #3 Windows path normalization** — no Python-only reproduction; likely shell/YAML environmental.
- **Issue #4 env-var regex rewrite** — inspection shows the current regex already accepts colons in defaults; reported failure is likely YAML quoting or PowerShell variable expansion.
- **Issue #5 dashboard keepalive, CLI gate command, dashboard auth, webhooks** — out of scope by user decision; revisit in a follow-up plan.
- **Issue #6 configurable retry budget** — no observed demand.
- **Issue #9 `command:` vs `args:` parity** — author labels anecdotal.

Each entry in the plan table includes a re-open threshold so the next person who hits one of these knows what evidence would justify reopening.

## Verification

- `make check` (ruff + ruff format + ty) — passes.
- `uv run pytest tests/test_executor/test_output.py tests/test_cli/test_web_flags.py -q` — 41 / 41 pass.
- Full suite passes (modulo the 11 pre-existing failures on `main` unrelated to this branch).

## Reviewer guidance

- **Item 1 is the highest-value change** — the brainstorm timeline shows it was the proximate cause of the multi-hour debugging session. Review the regex carefully; the test specifically guards the "string field containing ` ``` `" case that the old regex failed on.
- **Item 4 is pre-fork on purpose** — crashing the child after fork loses observability; failing fast at load is cheaper and clearer. Both `run` and `resume` get the check by necessity (run/resume parity rule in AGENTS.md), not as gold-plating.
- **The docs/plan files are intentionally verbose** — they enumerate explicitly-deferred work with reopen thresholds so this PR doesn't quietly become "the time someone shipped fixes 1, 3, 4 and #5 was never reconsidered."

### DIFF
--- a/src/conductor/cli/app.py
+++ b/src/conductor/cli/app.py
@@ -156,6 +156,39 @@ def print_error(error: Exception) -> None:
         console.print(panel)
 
 
+def _abort_web_bg_if_human_gate(workflow_path: Path, *, skip_gates: bool) -> None:
+    """Reject ``--web-bg`` when the workflow has a ``human_gate`` agent.
+
+    Without this check, ``--web-bg`` forks a detached child whose stdin is
+    redirected to ``DEVNULL``; ``Prompt.ask`` then raises ``EOFError`` and
+    the parent only reports ``"Background process exited immediately"``,
+    which never mentions ``human_gate`` or ``--skip-gates``. Failing fast
+    in the parent process produces a single visible error on the user's
+    terminal. ``--skip-gates`` is a documented escape hatch and is honored.
+    """
+    if skip_gates:
+        return
+    try:
+        from conductor.config.loader import load_config
+
+        config = load_config(workflow_path)
+    except Exception:  # noqa: BLE001 — defer real validation to the loader path
+        # If config fails to load, let the normal run path surface the error.
+        return
+    if not any(getattr(a, "type", None) == "human_gate" for a in config.agents):
+        return
+    raise typer.BadParameter(
+        "--web-bg is incompatible with workflows that contain human_gate steps "
+        "because the detached process has no stdin to prompt on.\n"
+        "\n"
+        "Options:\n"
+        "  1. Use --web (foreground) instead of --web-bg\n"
+        "  2. Add --skip-gates to auto-accept the first option\n"
+        "  3. Remove human_gate steps from the workflow\n"
+        "  4. Wait for CLI gate-resolution support (planned follow-up)"
+    )
+
+
 def version_callback(value: bool) -> None:
     """Display version information and exit."""
     if value:
@@ -420,6 +453,7 @@ def run(
 
     # Handle --web-bg: fork a background process and exit immediately
     if web_bg:
+        _abort_web_bg_if_human_gate(workflow_path, skip_gates=skip_gates)
         from conductor.cli.bg_runner import launch_background
 
         try:
@@ -855,6 +889,8 @@ def resume(
 
     # Handle --web-bg: fork a background process and exit immediately
     if web_bg:
+        if resolved_workflow is not None:
+            _abort_web_bg_if_human_gate(resolved_workflow, skip_gates=skip_gates)
         from conductor.cli.bg_runner import launch_background_resume
 
         try:
--- a/src/conductor/cli/app.py
+++ b/src/conductor/cli/app.py
@@ -177,9 +177,13 @@ def _abort_web_bg_if_human_gate(workflow_path: Path, *, skip_gates: bool) -> Non
         return
     if not any(getattr(a, "type", None) == "human_gate" for a in config.agents):
         return
-    raise typer.BadParameter(
-        "--web-bg is incompatible with workflows that contain human_gate steps "
-        "because the detached process has no stdin to prompt on.\n"
+    # Emit via plain typer.echo (not typer.BadParameter) so the message renders
+    # verbatim — BadParameter is rendered as a Rich panel whose text wrapping
+    # can split long flag names like ``--skip-gates`` across border lines in
+    # narrow terminals (e.g. CI runners), hiding the remediation hint.
+    message = (
+        "Error: --web-bg is incompatible with workflows that contain human_gate "
+        "steps because the detached process has no stdin to prompt on.\n"
         "\n"
         "Options:\n"
         "  1. Use --web (foreground) instead of --web-bg\n"
@@ -187,6 +191,8 @@ def _abort_web_bg_if_human_gate(workflow_path: Path, *, skip_gates: bool) -> Non
         "  3. Remove human_gate steps from the workflow\n"
         "  4. Wait for CLI gate-resolution support (planned follow-up)"
     )
+    typer.echo(message, err=True)
+    raise typer.Exit(code=2)
 
 
 def version_callback(value: bool) -> None:
--- a/src/conductor/cli/app.py
+++ b/src/conductor/cli/app.py
@@ -175,7 +175,10 @@ def _abort_web_bg_if_human_gate(workflow_path: Path, *, skip_gates: bool) -> Non
     except Exception:  # noqa: BLE001 — defer real validation to the loader path
         # If config fails to load, let the normal run path surface the error.
         return
-    if not any(getattr(a, "type", None) == "human_gate" for a in config.agents):
+    has_gate = any(getattr(a, "type", None) == "human_gate" for a in config.agents) or any(
+        getattr(getattr(fe, "agent", None), "type", None) == "human_gate" for fe in config.for_each
+    )
+    if not has_gate:
         return
     # Emit via plain typer.echo (not typer.BadParameter) so the message renders
     # verbatim — BadParameter is rendered as a Rich panel whose text wrapping
@@ -895,8 +898,25 @@ def resume(
 
     # Handle --web-bg: fork a background process and exit immediately
     if web_bg:
-        if resolved_workflow is not None:
-            _abort_web_bg_if_human_gate(resolved_workflow, skip_gates=skip_gates)
+        # When the user resumes via --from <checkpoint> alone (no workflow
+        # argument), resolved_workflow is None but the checkpoint records the
+        # original workflow path. Read it so the human_gate guard still fires
+        # and the user gets a single visible error instead of a silent crash
+        # in the detached child.
+        gate_check_workflow: Path | None = resolved_workflow
+        if gate_check_workflow is None and resolved_checkpoint is not None:
+            try:
+                ckpt_data = json.loads(resolved_checkpoint.read_text(encoding="utf-8"))
+                ckpt_workflow = ckpt_data.get("workflow_path")
+                if isinstance(ckpt_workflow, str):
+                    candidate = Path(ckpt_workflow)
+                    if candidate.exists():
+                        gate_check_workflow = candidate
+            except (OSError, json.JSONDecodeError):
+                # Checkpoint unreadable — let the normal resume path surface it.
+                pass
+        if gate_check_workflow is not None:
+            _abort_web_bg_if_human_gate(gate_check_workflow, skip_gates=skip_gates)
         from conductor.cli.bg_runner import launch_background_resume
 
         try:
--- a/src/conductor/executor/output.py
+++ b/src/conductor/executor/output.py
@@ -116,12 +116,28 @@ def parse_json_output(raw_response: str) -> dict[str, Any]:
 
     text = raw_response.strip()
 
-    # Try to extract JSON from markdown code blocks.
-    # Greedy capture so the regex closes at the LAST ``` in the response,
-    # not the first inner ``` (which may appear inside a JSON string field).
-    json_block_match = re.search(r"```(?:json)?\s*\n?(.*)\n?```", text, re.DOTALL)
-    if json_block_match:
-        text = json_block_match.group(1).strip()
+    # Try to extract JSON from markdown code blocks. Two-stage strategy:
+    # 1. Non-greedy findall + try-parse each candidate (first valid wins).
+    #    Handles the common case of multiple fenced blocks in one response
+    #    (e.g. "initial answer ... revised answer") where the first complete
+    #    JSON block is the authoritative one.
+    # 2. Greedy single capture as fallback. Handles the case where the JSON
+    #    contains literal ``` inside a string field, which breaks non-greedy
+    #    matching at the inner fence but is recovered by closing at the LAST
+    #    fence in the response.
+    candidates = re.findall(r"```(?:json)?\s*\n?(.*?)\n?```", text, re.DOTALL)
+    for candidate in candidates:
+        stripped = candidate.strip()
+        try:
+            result = json.loads(stripped)
+            if isinstance(result, dict):
+                return result
+            return {"result": result}
+        except json.JSONDecodeError:
+            continue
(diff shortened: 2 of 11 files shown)


src/conductor/executor/output.py
```diff
@@ -116,8 +116,10 @@ def parse_json_output(raw_response: str) -> dict[str, Any]:
 
     text = raw_response.strip()
 
-    # Try to extract JSON from markdown code blocks
-    json_block_match = re.search(r"```(?:json)?\s*\n?(.*?)\n?```", text, re.DOTALL)
+    # Try to extract JSON from markdown code blocks.
+    # Greedy capture so the regex closes at the LAST ``` in the response,
+    # not the first inner ``` (which may appear inside a JSON string field).
+    json_block_match = re.search(r"```(?:json)?\s*\n?(.*)\n?```", text, re.DOTALL)
```
jrob5756: **Behavior change worth a regression test:** going from non-greedy `(.*?)` to greedy `(.*)` correctly fixes the embedded-backticks case in the new test, but it also changes the behavior for responses with **multiple** fenced blocks. Repro:

```python
raw = '```json\n{"a": 1}\n```\n\nupdated answer:\n\n```json\n{"a": 2}\n```'
# OLD captures '{"a": 1}'                                                   → json.loads OK
# NEW captures '{"a": 1}\n```\n\nupdated answer:\n\n```json\n{"a": 2}'      → json.loads "Extra data"
```

The fallback at lines 125–134 (find first `{`/`[`) does **not** rescue this case because the greedy capture already starts with `{`, so we go straight to `json.loads(text)` on line 136 and raise `ValidationError`.

Realistic frequency is low (model emitting two JSON blocks in one response, e.g. "initial answer ... revised answer"), and the bug being fixed is more common, so I'm fine shipping this trade-off. But please:

1. **Add a regression test** that pins the chosen behavior on the multi-block case — either "first block wins" (use `re.findall` over the non-greedy pattern and try each) or "last fence is authoritative" (current PR behavior). Without that, the next person tightening the regex won't know which way is intentional.
2. **Or** switch to `re.findall` + per-candidate try-parse — handles both the backticks-in-string case (last candidate matches the full content) and the multi-block case (first valid JSON wins). Slightly more code, no behavior trade-off.

Not blocking — but worth a one-line decision in the commit or a test that says "yes, this is intentional."

src/conductor/providers/copilot.py
```diff
@@ -1098,8 +1098,10 @@ def _extract_json(self, content: str) -> dict[str, Any]:
         # Try to find JSON in code blocks
         import re
 
-        # Look for ```json ... ``` blocks
-        json_match = re.search(r"```(?:json)?\s*\n?(.*?)\n?```", content, re.DOTALL)
+        # Look for ```json ... ``` blocks (greedy so the regex closes at the
+        # LAST ``` in the content, not the first inner ``` which may appear
+        # inside a JSON string field).
+        json_match = re.search(r"```(?:json)?\s*\n?(.*)\n?```", content, re.DOTALL)
```
jrob5756, line 1275: Same multi-fence trade-off as `executor/output.py` line 122 — see that comment for the repro and suggestion. Worth keeping the two regexes consistent (they currently are), and worth a parallel regression test in `tests/test_providers/` if you go with the `re.findall` approach.

src/conductor/cli/app.py
```diff
@@ -155,6 +155,45 @@ def print_error(error: Exception) -> None:
...
+    if skip_gates:
+        return
+    try:
+        from conductor.config.loader import load_config
+
+        config = load_config(workflow_path)
+    except Exception:  # noqa: BLE001 — defer real validation to the loader path
+        # If config fails to load, let the normal run path surface the error.
+        return
+    if not any(getattr(a, "type", None) == "human_gate" for a in config.agents):
```
jrob5756: **Coverage gap (non-blocking):** this walks `config.agents` only. A workflow with `human_gate` solely as a `for_each.agent` (the inline agent inside a `for_each` group) is allowed by the validator — I confirmed with `conductor validate` on a synthetic workflow — and would still silent-crash under `--web-bg` with the original `EOFError`. 

Parallel groups are already covered because the validator at `config/validator.py:483` (PE-2.7) rejects `human_gate` in parallel groups, so the top-level walk is sufficient there.

Suggested minimal fix:
```python
has_gate = any(getattr(a, "type", None) == "human_gate" for a in config.agents) or any(
    getattr(fe.agent, "type", None) == "human_gate" for fe in config.for_each
)
if not has_gate:
    return
```

src/conductor/cli/app.py
```diff
@@ -833,6 +873,8 @@ def resume(
 
     # Handle --web-bg: fork a background process and exit immediately
     if web_bg:
+        if resolved_workflow is not None:
```
jrob5756: **Coverage gap (non-blocking):** when the user runs `conductor resume --web-bg` without an explicit workflow argument (relying on the checkpoint to recover the workflow path), `resolved_workflow is None` and the check is skipped — they can still hit the original silent crash. The full fix would be to derive the workflow path from the checkpoint here and re-run the check.

jrob5756 (commented):
Nice, tight, evidence-anchored PR. Three small fixes + docs, with an explicit deferred-issue table — that's exactly the shape I want for friction-driven follow-ups.

**What I verified locally on the PR branch:**
- `ruff check` and `ruff format --check` — clean
- `pytest tests/test_executor tests/test_cli tests/test_providers -m "not performance"` — 954 passed, 3 skipped
- The 3 new `--web-bg` + `human_gate` tests are deterministic (mock `launch_background`, no subprocess)
- `parse_json_output` on the PR's own backticks-in-string case — fixed; OLD truncated, NEW parses
- `ty` typecheck has one diagnostic that is pre-existing on `main`, unrelated to this PR
- The branch is behind `main` (missing `examples/wait-smoke.yaml`, `terminate.yaml`, `set-step.yaml`, `wait-step.yaml`, `examples/README.md` added in #221/#224); a rebase before merge will pick those up, no code conflict expected

**Main finding — the greedy regex has a real (small) regression I'd like a regression test for**, see inline comment on `output.py`. The other inline comments are coverage gaps in `_abort_web_bg_if_human_gate` that I think are acceptable to ship but worth acknowledging in code or a follow-up.

lucioctinoco:
Thanks for the review! Pushed 41d18d4 addressing all four inline comments:

**1. `output.py` line 122 + `copilot.py` line 1104 — multi-fence regression**

Went with your suggestion #2 (`re.findall` + per-candidate try-parse) since it has no behavior trade-off. Two-stage strategy:
- First: non-greedy `re.findall` over fenced blocks, try-parse each in order, **first valid wins** — handles the multi-block case from your repro.
- Fallback: greedy single capture — handles the backticks-in-string case (where non-greedy splits the JSON at the inner fence and no individual candidate parses).

Both parsers kept in parity. Added regression tests in both `test_output.py` and `test_copilot.py` that pin first-valid-wins on your exact repro input.

**2. `app.py` line 177 — `_abort_web_bg_if_human_gate` coverage gap (for_each)**

Applied your suggested fix verbatim — the check now walks both `config.agents` and `config.for_each[*].agent`. Parallel groups remain excluded because `config/validator.py:483` (PE-2.7) already rejects `human_gate` there.

**3. `app.py` line 876 — resume coverage gap (checkpoint-only path)**

When the user runs `conductor resume --from <ckpt> --web-bg` without a workflow argument, the gate guard was previously skipped. Now reads `workflow_path` from the checkpoint JSON and runs the same check. Falls through silently if the checkpoint is unreadable so the normal resume path still surfaces the real error.

Regression tests added for both #2 and #3 in `tests/test_cli/test_web_flags.py`.

**Verification:** `pytest tests/test_executor tests/test_cli tests/test_providers -m "not performance"` → 959 passed, 3 skipped (954 prior + 5 new). `ruff check` and `ruff format --check` clean.

lucioctinoco:
Thanks for the review! Pushed 41d18d4 addressing all four inline comments:

**1. `output.py` line 122 + `copilot.py` line 1104 — multi-fence regression**

Went with your suggestion #2 (`re.findall` + per-candidate try-parse) since it has no behavior trade-off. Two-stage strategy:
- First: non-greedy `re.findall` over fenced blocks, try-parse each in order, **first valid wins** — handles the multi-block case from your repro.
- Fallback: greedy single capture — handles the backticks-in-string case (where non-greedy splits the JSON at the inner fence and no individual candidate parses).

Both parsers kept in parity. Added regression tests in both `test_output.py` and `test_copilot.py` that pin first-valid-wins on your exact repro input.

**2. `app.py` line 177 — `_abort_web_bg_if_human_gate` coverage gap (for_each)**

Applied your suggested fix verbatim — the check now walks both `config.agents` and `config.for_each[*].agent`. Parallel groups remain excluded because `config/validator.py:483` (PE-2.7) already rejects `human_gate` there.

**3. `app.py` line 876 — resume coverage gap (checkpoint-only path)**

When the user runs `conductor resume --from <ckpt> --web-bg` without a workflow argument, the gate guard was previously skipped. Now reads `workflow_path` from the checkpoint JSON and runs the same check. Falls through silently if the checkpoint is unreadable so the normal resume path still surfaces the real error.

Regression tests added for both #2 and #3 in `tests/test_cli/test_web_flags.py`.

**Verification:** `pytest tests/test_executor tests/test_cli tests/test_providers -m "not performance"` → 959 passed, 3 skipped (954 prior + 5 new). `ruff check` and `ruff format --check` clean.

jrob5756 (approved):
All review findings addressed in commit 41d18d40 (now 9284c07 post-rebase). Two-stage parser strategy with regression tests for both backticks-in-string and multi-fence-first-wins. for_each and resume-from-checkpoint gaps both closed with regression tests. Rebased onto main, full test suite green (3136 passed), all CI checks green. 🚢

PR #234, opened 2026-05-28
feat: external-workflow-friction EPICs 1-4 (output_mode, max_parse_recovery_attempts, gate-respond CLI, Windows paths)

lucioctinoco:
## Summary

Follow-up to #232. This PR delivers EPICs 1–4 from the external-workflow-friction plan ([`docs/projects/usability-features/external-workflow-friction-v2.plan.md`](docs/projects/usability-features/external-workflow-friction-v2.plan.md)) — improvements that surfaced while running real-world workflows but were out of scope for the minimal evidence-anchored fixes in #232.

Rebased cleanly onto `main` after #232 landed.

## EPIC 1 — Cross-provider parity & `output_mode`

- New schema field `output_mode: raw | envelope` on agent configs. Default `envelope` keeps current behavior; `raw` returns the model's response verbatim (useful for prompts that already constrain the structure).
- Mutual-exclusion check: `output_mode: raw` cannot coexist with an `output:` schema block.
- Field is rejected on non-prompt agent types (`script`, `human_gate`, `workflow`, `wait`, `set`, `terminate`).
- Cross-provider parity fixes: aligned Claude and Copilot providers around `RetryConfig` wiring and corrected test assertions that had drifted between providers.
- Fixed parse-exhaustion retry so the engine reports `is_retryable=False` once all parse-recovery attempts are spent (instead of looping or surfacing a misleading retryable error).
- New tests: `tests/test_config/test_output_mode.py`, `tests/test_providers/test_output_mode.py`.

## EPIC 2 — Configurable `max_parse_recovery_attempts`

- Exposes `max_parse_recovery_attempts` in the YAML `retry:` policy. Previously hard-coded; now per-agent tunable for workflows that legitimately need more (or fewer) recovery rounds when models emit malformed structured output.

## EPIC 3 — `conductor gate-respond` CLI

- New subcommand for resolving `human_gate` agents from the terminal without opening the dashboard. Useful for scripted/CI flows and for the `--web-bg` case where the dashboard is the only other resolver.
- Hardened malformed-JSON handling in the underlying gate API; improved error messages when the dashboard is unreachable or no gate is waiting.
- Docs updated: [`docs/cli-reference.md`](docs/cli-reference.md), [`CHANGELOG.md`](CHANGELOG.md).

## EPIC 4 — Windows path normalization in script executor

- The script executor now normalizes path-shaped values when running on Windows so that `tmp/foo` / `tmp\foo` resolve identically, and forward-slash paths supplied via `args:` or `env:` no longer break downstream tools that expect native separators.

## Test status

- Lint: `uv run ruff check src tests` ✅
- Format: `uv run ruff format --check src tests` ✅
- Tests: `uv run pytest -m "not performance"` — **3192 passed**, 14 skipped. 11 failures are pre-existing on `main` (Windows-only TOML hex-escape and path-separator assertions in `test_registry/test_integration.py` and one `test_event_log` assertion) and reproduce identically on a clean `origin/main` checkout — not introduced by this PR.

## Plan reference

[`docs/projects/usability-features/external-workflow-friction-v2.plan.md`](docs/projects/usability-features/external-workflow-friction-v2.plan.md)


src/conductor/executor/script.py
```diff
@@ -84,6 +86,18 @@ async def execute(
...
+        # and a Windows path missing its .exe suffix resolves correctly. Relative
+        # paths containing a separator are left untouched so they keep resolving
+        # against ``working_dir``. Resolution is non-destructive: when ``which``
+        # cannot resolve the command we fall back to the rendered value and let
+        # the FileNotFoundError handler below produce a clear error.
+        has_separator = os.sep in rendered_command or (
+            os.altsep is not None and os.altsep in rendered_command
+        )
+        if os.path.isabs(rendered_command) or not has_separator:
+            rendered_command = shutil.which(rendered_command) or rendered_command
```
jrob5756: This is the blocking issue I called out in the summary. `shutil.which(rendered_command)` resolves against the parent `os.environ["PATH"]`, but the subprocess (line ~125) is launched with a merged env where `agent.env` can override `PATH`. I reproduced this: a `script` step with `env: {PATH: /childbin}` ran `/childbin/toolx` before this PR but now runs `/parentbin/toolx` — a different binary, silently. Resolution also happens before `env` is built, and it runs on all platforms, not just Windows.

The simplest fix is to build `env` first, then resolve with the same PATH:
```python
rendered_command = shutil.which(rendered_command, path=env.get("PATH")) or rendered_command
```

src/conductor/providers/claude.py
```diff
@@ -1867,16 +1898,23 @@ async def _execute_with_parse_recovery(
...
         logger.error(
-            f"Parse recovery exhausted after {self._max_parse_recovery_attempts} attempts. "
+            f"Parse recovery exhausted after {effective_max_recovery} attempts. "
             f"History: {'; '.join(recovery_history)}"
         )
+        # Note: is_retryable=False is set for correctness and documentation
+        # clarity. In practice, Claude's _is_retryable_error() already
+        # returns False for ProviderError (not an Anthropic SDK type), so
+        # the outer retry loop won't retry regardless. The attribute
+        # ensures consistent behavior if the retry dispatch ever changes.
```
jrob5756: This comment is now wrong. It says `is_retryable=False` is for "documentation clarity" and the outer loop "won't retry regardless," but this same PR added `if isinstance(exception, ProviderError): return exception.is_retryable` to `_is_retryable_error()` around line 726. The flag is now load-bearing — it directly controls whether retries happen. Reword to something like "`_is_retryable_error()` now honors `ProviderError.is_retryable`, which is False here."

src/conductor/web/server.py
```diff
@@ -233,6 +239,84 @@ async def resume_agent() -> JSONResponse:
...
+            "additional_input": str?}``
+
+            When the ``CONDUCTOR_GATE_TOKEN`` environment variable is set,
+            the request must carry a matching token in the
+            ``Authorization: Bearer <token>`` header.
+            """
+            try:
+                body = await request.json()
+            except Exception:
+                return JSONResponse({"error": "Invalid JSON body"}, status_code=422)
```
jrob5756, line 269: The bare `except Exception` here means `ClientDisconnect` and `UnicodeDecodeError` both surface as "Invalid JSON body" with no logging, which makes real failures harder to diagnose. Narrowing to `except (json.JSONDecodeError, UnicodeDecodeError)` would keep the 422 path for actual JSON problems and let other errors bubble up where they belong.

CHANGELOG.md
```diff
@@ -120,6 +120,42 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  budgets. Accepts integer 0–10; `0` disables all recovery attempts and lets
+  the first parse failure propagate immediately. Threaded through both the
+  Copilot and Claude providers.
+- New `POST /api/gate-respond` and `GET /api/gate-status` HTTP API endpoints
+  on the web dashboard server. `GET /api/gate-status` returns whether a
+  `human_gate` agent is currently waiting, and which agent name it is.
+  `POST /api/gate-respond` resolves the parked gate by injecting a
+  `GateResponse` into the engine's queue. Both endpoints respect an optional
+  `CONDUCTOR_GATE_TOKEN` secret for auth; when a token is configured on the
+  server any request without a matching `token` field is rejected with HTTP 403.
```
jrob5756: Two things to correct here. Only `POST /api/gate-respond` enforces the token — `GET /api/gate-status` is unauthenticated, so "Both endpoints" isn't quite right. And the auth is via the `Authorization: Bearer` header, not a `token` field in the request body (the test for `gate_respond` explicitly rejects body-supplied tokens).

CHANGELOG.md
```diff
@@ -120,6 +120,42 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- New `conductor gate-respond` CLI command for resolving a parked human gate
+  from the command line without opening a browser. Accepts `--port`, `--choice`,
+  `--agent` (auto-discovered via `/api/gate-status` when omitted), `--input`,
+  and `--token` / `CONDUCTOR_GATE_TOKEN` env var. Designed for SSH or headless
+  environments where the web dashboard UI is unreachable.
+
+### Changed
+- **Breaking (Claude provider):** `ClaudeProvider._extract_text_response` now
```
jrob5756: Wrong method name — the actual method is `_extract_text_content`, not `_extract_text_response`.

jrob5756 (commented):
## Review Summary

Looked at this across code quality, tests, error handling, type design, comments, dead code, and simplification. All 354 tests pass, `ty check` is clean, and overall the parity work is solid. One blocking issue and a handful of other things worth addressing.

### Blocking

**`shutil.which` ignores the agent's `env.PATH`, silently running the wrong binary.** `executor/script.py:96-100`

Command resolution uses `shutil.which(rendered_command)`, which resolves against the **parent** `os.environ["PATH"]`. But the subprocess (line ~125) is launched with a merged env where `agent.env` can override `PATH`. I reproduced this end-to-end: a `script` step with `env: {PATH: /childbin}` ran `/childbin/toolx` pre-PR but now runs `/parentbin/toolx` — wrong binary, silently. A documented per-step PATH/toolchain-selection pattern is broken. Note also that `shutil.which` runs on all platforms here, not Windows-only as the EPIC framing implies.

Fix: build `env` first, then `shutil.which(rendered_command, path=env.get("PATH")) or rendered_command`.

### Worth addressing

- **Stale `is_retryable` comment that's now contradictory.** `claude.py:1904-1908`. The comment says the flag is for "documentation clarity" and the loop "won't retry regardless," but this same PR added `if isinstance(exception, ProviderError): return exception.is_retryable` to `_is_retryable_error()` (line ~726). The flag is now load-bearing — it directly controls retry behavior. Worth updating the comment so the next person doesn't think it's safe to flip.

- **`CONDUCTOR_GATE_TOKEN` only protects the HTTP gate endpoint, not the WebSocket gate path.** `server.py:283-289` vs `:410-411`. The existing `/ws` `gate_response` handler enqueues with no token check and no waiting-state validation, so anyone who can reach `/ws` can still resolve gates and bypass the token entirely. The `127.0.0.1` default bind mitigates real-world exposure, but the asymmetry is surprising — worth either closing it or documenting that the token is HTTP-only.

- **CHANGELOG has a few inaccuracies.** Line 150 references `_extract_text_response`; the actual method is `_extract_text_content`. Lines 140-142 say "Both endpoints" enforce the token, but only `POST /api/gate-respond` does — `GET /api/gate-status` is unauthenticated. Same lines call it a "token field," but auth is via the `Authorization: Bearer` header (the body-token path is explicitly rejected, per the test). Also: there's no CHANGELOG entry at all for the new script-resolution behavior.

- **Bare `except Exception` swallows and mislabels.** `server.py:260-262`. Wrapping `request.json()` in a bare except means `ClientDisconnect` and `UnicodeDecodeError` both surface as "Invalid JSON body" with no logging. Narrow to `(json.JSONDecodeError, UnicodeDecodeError)` so genuinely different failures aren't conflated.

- **`_extract_text_content` `{"text"}` → `{"result"}` is breaking for all no-schema Claude agents.** `claude.py:2224`. This affects both the main no-schema path and the mid-interrupt partial path, not just the new raw mode — any existing workflow referencing `{{ agent.output.text }}` will break. The change itself is correct (it aligns Claude with Copilot and the engine's documented contract), and it is called out as Breaking in the CHANGELOG. Just worth making sure the migration note stays prominent.

- **Test gaps.** No end-to-end gate round-trip (HTTP submit → engine consumes → workflow continues); `_gate_waiting_agent` set-state is never asserted (every API test sets it manually); the 403-vs-422 ordering — a real security property — isn't pinned. Two Copilot `output_mode` tests (`test_output_mode.py:37, :85`) are effectively tautological because `mock_handler` short-circuits before the output_mode logic runs. And `test_args_not_resolved` (`test_script.py:396`) should assert `mock_which.call_count == 1` to actually pin command-only resolution.

### Smaller suggestions

- **Latent parity bug for `output: {}`.** Claude uses `agent.output is not None` (`claude.py:932`); Copilot uses truthiness (`copilot.py:688`). An empty-dict schema is treated as a schema by Claude and as no-schema by Copilot — same input, different behavior. Extracting `AgentDef.effective_output_schema()` and calling it from both providers would fix this and de-duplicate the rule.

- **Gate double-submit can leave a stale response queued.** `server.py:292-318`. The single-request check→enqueue is atomic (good), but two concurrent POSTs (or a POST racing a WS submit) can both pass the waiting check and enqueue. `wait_for_gate_response` consumes one and clears `_gate_waiting_agent`; the other stays in the queue. Given loop-back patterns, if the same `agent_name` gate is reached again later, the leftover auto-resolves it. Consider draining or guarding the queue on resolution.

- **Claude resolves `max_parse_recovery_attempts` twice.** The fallback in `_execute_api_call_with_recovery` (`claude.py:1786-1792`) and the `self._max_parse_recovery_attempts` attribute (`:191`) are effectively dead — `_resolve_retry_config` already produces a non-None value before this method is called. One resolution point is enough.

- **Minor:** redundant `has_schema` / `has_output_schema` aliasing at `claude.py:935, 957`; and the inline comment in `script.py:89-95` describes PATHEXT/normalization behavior that doesn't quite match the actual command-only resolution (and won't make sense once the env-PATH issue above is fixed).

### What's working well

- 354 tests pass in ~9.4s; `ty check` clean; no dead code introduced.
- The parse-exhaustion `is_retryable=False` work is genuine cross-provider parity (Copilot via `except ProviderError` in its retry loop, Claude via `_is_retryable_error` honoring the flag — different mechanisms, same outcome).
- Good negative and parity test coverage; sensible security guards on the new endpoint (constant-time token compare, explicit body-token rejection test, full 403/409/422/200 status handling).
- `max_parse_recovery_attempts` bounds (`ge=0, le=10`) with `0` correctly disabling recovery.

### Suggested next steps

The blocker is the one I'd want to see fixed before merge — it's a silent correctness regression on a documented pattern. Everything else can be batched: comment/CHANGELOG cleanup, the bare-except narrowing, the WS/HTTP auth gap, and the test additions. The smaller suggestions are nice-to-haves.

lucioctinoco:
Thanks for the thorough review, @jrob5756 — all points addressed in 6ad1ac8. Summary:

**Blocker**
- **`script.py` command resolution** — `shutil.which` now resolves against the subprocess's own `PATH` (`shutil.which(rendered_command, path=env.get("PATH")) or rendered_command`), and the env block is built *before* resolution so an `agent.env` `PATH` override picks the same binary the child will actually run. No more silent wrong-binary mismatch.

**Security / correctness**
- **WS gate auth gap** — extracted `_gate_token_ok()` (constant-time Bearer check) and `_validate_gate_target()`, and applied both on the `/ws` `gate_response` path, not just the HTTP endpoint. Rejections are logged.
- **Gate double-submit race** — duplicate queued responses are now drained **on resolution** (after the matching message is consumed), per your suggestion, instead of guarding at entry.
- **Narrowed bare `except`** on `request.json()` to `(json.JSONDecodeError, UnicodeDecodeError)`.

**Provider parity**
- Added `AgentDef.effective_output_schema()` and routed both Copilot and Claude through it, so an empty-dict output schema is treated identically by both providers.
- Claude: collapsed parse-recovery to a single resolution point, dropped the redundant `has_schema` alias, and reworded the `is_retryable=False` comment now that `_is_retryable_error()` honors the flag directly.

**Docs**
- CHANGELOG: corrected the gate-auth description (only `POST /api/gate-respond` requires the `Authorization: Bearer` header; `GET /api/gate-status` is unauthenticated), fixed `_extract_text_content`, and added the script command-resolution entry.

**Tests** (closing the gaps you flagged)
- e2e HTTP gate round-trip, waiting-state validation, and 403-before-422 ordering.
- De-tautologized the Copilot `output_mode` tests so they actually drive the SDK path; pinned the `which` `call_count`.

Full local gate is green (ruff, ruff format, ty, pytest 3202 passed). Remaining suite failures are pre-existing Windows/environmental (registry TOML escaping, real-installer temp-dir cleanup, a live-model integration test) and unrelated to this change.

jrob5756 (approved):
LGTM

PR #235, opened 2026-05-28
Feat/hermes provider

tima-110:
## Summary

 Adds the NousResearch `hermes-agent` library as a third optional provider alongside Copilot and Claude. Hermes currently has limited workflow capabilities, and this extends it's capabilities.   Hermes is a multi-agent, provider agnostic AI assistant, with its own built-in tool ecosystem — no MCP configuration required.

  The library is an **optional dependency** — Conductor works without it. Users who want
  the hermes provider install it separately (`pip install hermes-agent`).

  ## Changes

  ### New provider (`src/conductor/providers/hermes.py`)
  - `HermesProvider` wraps the synchronous `AIAgent.run_conversation()` call via
    `asyncio.run_in_executor` so it fits Conductor's async execution model
  - Forwards `system_prompt` as `system_message=` to `run_conversation()`
  - Maps `agent.tools` to `AIAgent(enabled_toolsets=...)`:
    `null` → omit (all ~33 tools), `[]` → empty allowlist (0 tools), `["web"]` → named toolsets
  - Forwards `max_tokens`, `temperature`, `base_url`, and `api_key` from provider config to `AIAgent`
  - Supports `max_session_seconds` as an `asyncio.wait_for` timeout
  - Supports interrupt signals (Esc/Ctrl+G) via `asyncio.wait` racing
  - Supports structured `runtime.provider: { name: hermes, base_url: ..., api_key: ... }` for
    routing through custom endpoints (OpenRouter, litellm gateways, corporate proxies)

  ### Factory (`src/conductor/providers/factory.py`)
  - New `case "hermes"` branch wires all runtime config through to `HermesProvider`

  ### Schema (`src/conductor/config/schema.py`)
  - `ProviderSettings._check_field_compatibility` relaxed to allow `base_url`/`api_key`
    for `name="hermes"` (was copilot-only)

  ### Tests
  - 28 unit tests covering init, connection validation, execute (plain text, structured
    output, model resolution, iteration limits, token counts, event callbacks, interrupt,
    timeout, system_prompt forwarding, tools mapping, param forwarding, error messages)
  - 4 new `ProviderSettings` tests for hermes `base_url`/`api_key` acceptance and
    regression guard that `name="claude"` still rejects structured config

  ### Docs and skills
  - `docs/providers/hermes.md` — full provider reference (runtime config, toolset control,
    model routing, structured output, limitations, troubleshooting)
  - `docs/providers/comparison.md` — updated to three-way comparison
  - `README.md`, `docs/configuration.md`, `docs/cli-reference.md`,
    `docs/workflow-syntax.md` — hermes added to all provider enumerations
  - Skill references updated (`yaml-schema.md`, `execution.md`) for accuracy

  ## Limitations (documented)

  - No MCP server support — hermes has its own tool ecosystem
  - No streaming events — synchronous library, responses arrive all-at-once
  - No session resume — stateless, no persistent session to checkpoint
  - `reasoning.effort` accepted for interface parity but has no effect
  - Structured output via prompt injection (less reliable than native APIs)

  ## Test plan

  - [ ] `pytest tests/test_providers/test_hermes.py` — 28 tests pass
  - [ ] `pytest tests/test_config/test_provider_settings.py` — all pass
  - [ ] `pytest tests/ -q` — full suite, no regressions
  - [ ] `conductor validate examples/hermes-simple.yaml` — schema validates
  - [ ] End-to-end smoke test (requires `pip install hermes-agent` and API key):
    `conductor run examples/hermes-simple.yaml --input question="What is Python?"`


tima-110:

src/conductor/providers/hermes.py
```diff
@@ -0,0 +1,315 @@
...
+logger = logging.getLogger(__name__)
+
+# JSON instruction appended to the prompt when a structured output schema is declared.
+_JSON_INSTRUCTION = (
+    "\n\nRespond ONLY with a valid JSON object. "
+    "Do not include any explanation, markdown, or text outside the JSON object."
+)
+
+
+class HermesProvider(AgentProvider):
```
jrob5756, line 57: **BLOCKER — Missing `CAPABILITIES` declaration (rebase blocker)**

After rebase onto current `main`, `__init_subclass__` in `AgentProvider` (added by commit ca73f56) will refuse to import this class until a class-level `CAPABILITIES: ProviderCapabilities` attribute is declared. See the §2 section of the main review for the suggested declaration and the required registration in `capabilities.py:_PROVIDER_CLASS_PATHS`.

Without these two changes, `import conductor.providers.hermes` raises `TypeError` and `conductor validate` fails for every workflow naming `provider: hermes`.

src/conductor/providers/hermes.py
```diff
@@ -0,0 +1,315 @@
...
+        if self._base_url:
+            agent_kwargs["base_url"] = self._base_url
+        if self._api_key:
+            agent_kwargs["api_key"] = self._api_key
+        # Map Conductor tools field to hermes enabled_toolsets.
+        # tools=None → omit (hermes default: all tools)
+        # tools=[]   → enabled_toolsets=[] (empty allowlist = 0 tools)
+        # tools=[..] → enabled_toolsets=[..] (hermes toolset names)
+        if agent.tools is not None:
+            agent_kwargs["enabled_toolsets"] = agent.tools
```
jrob5756: **BLOCKER — Security regression: Conductor `tools:` allowlist silently reinterpreted as hermes toolset names**

`AgentDef.tools` is documented as a Conductor tool allowlist (`None` = all workflow tools, `[]` = none, `[list]` = subset of workflow tool names). This passthrough forwards it as-is to hermes-agent's `enabled_toolsets` — a **completely different namespace** (`web`, `code_interpreter`, etc.).

Concrete failure modes:
- `tools: ["read_file", "write_file"]` (Conductor-style names from a workflow that runs against Copilot/Claude) becomes an invalid hermes toolset list — either silently empty or a runtime error with names the author never typed.
- `tools: null` (= "all workflow tools" in Conductor semantics) becomes "all of hermes's ~33 default toolsets" — possibly granting capabilities the author never reviewed.

This is exactly what `ProviderCapabilities.workflow_tools_passthrough` exists to prevent — its docstring explicitly calls out: *"silently dropping the allowlist is a security regression."*

**Fix** (paired with `workflow_tools_passthrough=False` in CAPABILITIES, per the main review): drop this passthrough. The validator will then reject any agent that sets `tools:` against `provider: hermes`, forcing authors to be explicit. If hermes-specific toolset selection is needed later, expose it via an explicit hermes-only field (e.g. `agent.provider_options.hermes_toolsets`) so the vocabularies cannot be confused.

src/conductor/providers/hermes.py
```diff
@@ -0,0 +1,315 @@
...
+            if interrupt_task is not None and interrupt_task in done:
+                raise ProviderError(
+                    f"Agent '{agent.name}' was interrupted by user request",
+                )
+            raise ProviderError(
+                f"Agent '{agent.name}' exceeded maximum session duration "
+                f"of {resolved_timeout:.0f}s",
+            )
+
+        result = call_task.result()
```
jrob5756: **BLOCKER — Raw third-party exceptions escape unwrapped**

`call_task.result()` re-raises whatever the executor thread raised, completely bypassing the `ProviderError` mapping the class docstring promises (`"Library failures, API errors, or unexpected result states → ProviderError"`).

Things that leak through today:
- `openai.AuthenticationError` / `httpx.ConnectError` — user sees the raw SDK error with no hint it came from hermes
- `KeyError` on a bad model name
- `ValueError` from `AIAgent.__init__` (bad `base_url`, unknown kwarg if hermes-agent version drifts)
- `json.JSONDecodeError` from inside hermes if it parses tool output internally

**Fix:**

```python
try:
    result = call_task.result()
except asyncio.CancelledError:
    raise
except Exception as e:
    raise ProviderError(
        f"Hermes agent execution failed (model='{resolved_model}'): {e}",
        suggestion="Check the hermes-agent logs and verify base_url/api_key.",
    ) from e
```

The `from e` preserves the original traceback for debugging/Sentry. Pair with `is_retryable=` classification (see comment on lines 225-233).

src/conductor/providers/hermes.py
```diff
@@ -0,0 +1,315 @@
...
+            # Either timeout or interrupt fired first
+            call_task.cancel()
+            if interrupt_task is not None and interrupt_task in done:
+                raise ProviderError(
+                    f"Agent '{agent.name}' was interrupted by user request",
+                )
+            raise ProviderError(
+                f"Agent '{agent.name}' exceeded maximum session duration "
+                f"of {resolved_timeout:.0f}s",
+            )
```
jrob5756, line 378: **BLOCKER — Two issues here**

**(a) `call_task.cancel()` cannot actually stop the executor thread.** `Future.cancel()` on a `run_in_executor` future returns `False` once the thread has started executing. The wrapping coroutine raises `ProviderError`, but the underlying `run_conversation` call keeps running — burning tokens and potentially executing destructive tool calls for up to `max_iterations=90` after the user thought the agent was stopped.

This means `interrupt=True` in CAPABILITIES would be technically dishonest. Two acceptable resolutions:

1. **Declare `interrupt=False`** in CAPABILITIES (recommended — honest about behavior) and add a `logger.warning(...)` here noting the thread continues.
2. Investigate whether hermes-agent exposes a cooperative cancellation hook (event flag, generator API) and use it inside `_run_sync`.

At minimum, update the interrupt `ProviderError` message:

```python
raise ProviderError(
    f"Agent '{agent.name}' was interrupted; underlying hermes call may "
    f"continue in the background until natural completion.",
    is_retryable=False,
)
```

**(b) Neither `ProviderError` sets `is_retryable`.** It defaults to `False`, so a transient timeout aborts the workflow instead of retrying with a fresh `AIAgent`. The same issue exists at lines 240 and 247.

Suggested classifications:
- Interrupt → `is_retryable=False` (user explicitly stopped)
- Timeout → `is_retryable=True` (next attempt might be faster)
- `result.get("failed")` (line 238) → `is_retryable=False` (library-level error, not transient)
- `final_response is None` (line 245) → `is_retryable=True` (degraded mid-stream response)

src/conductor/providers/hermes.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+def _fire(callback: EventCallback | None, event: str, data: dict[str, Any]) -> None:
+    """Call event_callback safely, swallowing any exception."""
+    if callback is None:
+        return
+    try:
+        callback(event, data)
+    except Exception:
+        logger.debug("Error in event_callback for %s", event, exc_info=True)
```
jrob5756: **BLOCKER — Callback exceptions silently swallowed at DEBUG level**

If the event sink is broken (schema mismatch, dashboard down, serialization error on a new field), the operator sees nothing — the default log level hides DEBUG. The agent appears to run normally; observability is silently dead.

This pattern matches `copilot.py` / `claude.py`, but those providers emit far more events so a silent drop is more visible in aggregate. For Hermes, with only two event types fired per agent run, a silent drop is total observability loss for the entire agent.

Suggested fix (minimum):
```python
except Exception:
    logger.warning("Error in event_callback for %s", event, exc_info=True)
```

Better: re-raise on first failure (let the user know something is broken) and only suppress after an opt-in flag. Either way, `debug` is the wrong level here.

src/conductor/providers/factory.py
```diff
@@ -106,10 +107,32 @@ async def create_provider(
...
+            provider = HermesProvider(
+                model=default_model,
+                max_tokens=max_tokens,
+                temperature=temperature,
+                base_url=hermes_base_url,
+                api_key=hermes_api_key,
+                max_agent_iterations=max_agent_iterations,
+                max_session_seconds=max_session_seconds,
+                default_reasoning_effort=default_reasoning_effort,
+            )
```
jrob5756, line 165: **BLOCKER — Zero test coverage for this branch**

Neither `tests/test_providers/test_factory.py` nor `tests/test_providers/test_hermes.py` exercises any code path in this `case "hermes":` branch. The branch contains non-trivial logic that needs explicit tests:

1. `HERMES_SDK_AVAILABLE=False` → `ProviderError("requires the hermes-agent package")` with the `pip install hermes-agent` suggestion.
2. `provider_settings is not None and provider_settings.name == "hermes"` → `base_url` / `api_key.get_secret_value()` extraction.
3. `provider_settings is not None and provider_settings.name == "copilot"` (or other non-hermes) → `hermes_base_url`/`hermes_api_key` stay None.
4. All eight constructor params (`model`, `max_tokens`, `temperature`, `base_url`, `api_key`, `max_agent_iterations`, `max_session_seconds`, `default_reasoning_effort`) forwarded to `HermesProvider`.

The PR description's end-to-end claim — *"Structured `runtime.provider: { name: hermes, base_url, api_key }` works end-to-end"* — is not actually verified end-to-end. `test_hermes.py:475-507` asserts the kwargs reach `AIAgent`, and `test_provider_settings.py:185-203` asserts the schema accepts the fields, but **nothing connects the two**. A future refactor that drops the `.get_secret_value()` call at line 121 would ship a `SecretStr` object to `AIAgent` undetected.

Please add tests covering each numbered case above to `tests/test_providers/test_factory.py`.

src/conductor/providers/hermes.py
```diff
@@ -0,0 +1,315 @@
...
+    def __init__(
+        self,
+        model: str | None = None,
+        max_tokens: int | None = None,
+        temperature: float | None = None,
+        base_url: str | None = None,
+        api_key: str | None = None,
+        max_agent_iterations: int | None = None,
+        max_session_seconds: float | None = None,
+        default_reasoning_effort: ReasoningEffort | None = None,
```
jrob5756, line 103: **SHOULD-FIX — `default_reasoning_effort` accepted then completely discarded**

The docstring says *"is accepted for interface parity but has no effect"*, which is correct on the wire — but the parameter is **not even stored** on `self`. Compare lines 102-108 where every other parameter is stored as `self._default_*`. A future maintainer adding a `__repr__` or debug-logging block would assume the value is at least retained for inspection. It isn't.

Two acceptable resolutions:

**Option 1** (preferred — visible no-op): Store it and emit a one-time WARNING when used:

```python
self._default_reasoning_effort = default_reasoning_effort
if default_reasoning_effort is not None:
    logger.warning(
        "HermesProvider received reasoning_effort=%s but hermes-agent does "
        "not expose this control. Setting will be ignored.",
        default_reasoning_effort,
    )
```

**Option 2**: Drop the parameter entirely and document the constructor-signature divergence from other providers prominently in the class docstring.

Either way, the current "silent discard with a buried docstring note" pattern will frustrate users debugging *"my reasoning_effort setting did nothing"*.

(Paired with `reasoning_effort=None` in CAPABILITIES, the validator will eventually reject workflows that declare `reasoning.effort` against this provider — but `default_reasoning_effort` at the runtime level can still be set, and that path needs to be visible.)

src/conductor/providers/hermes.py
```diff
@@ -0,0 +1,315 @@
...
+            "skip_memory": True,
+        }
+        if resolved_model:
+            agent_kwargs["model"] = resolved_model
+        if resolved_max_iter is not None:
+            agent_kwargs["max_iterations"] = resolved_max_iter
+        if self._default_max_tokens is not None:
+            agent_kwargs["max_tokens"] = self._default_max_tokens
+        if self._default_temperature is not None:
+            agent_kwargs["temperature"] = self._default_temperature
```
jrob5756: **SHOULD-FIX — Per-agent `max_tokens` / `temperature` overrides silently dropped (parity break)**

These lines only consult `self._default_max_tokens` / `self._default_temperature`, never `agent.max_tokens` / `agent.temperature`. Compare lines 146-155 above, which correctly resolve `agent.max_agent_iterations` and `agent.max_session_seconds` with the same `agent.X if agent.X is not None else self._default_X` pattern. And compare `claude.py` / `copilot.py`, both of which resolve per-agent overrides first.

A workflow author writing:

```yaml
agents:
  - name: precise_analyzer
    temperature: 0.0      # silently ignored
    max_tokens: 16384     # silently ignored
```

…will see their overrides discarded with no warning. Fix:

```python
resolved_max_tokens = agent.max_tokens if agent.max_tokens is not None else self._default_max_tokens
resolved_temperature = agent.temperature if agent.temperature is not None else self._default_temperature
if resolved_max_tokens is not None:
    agent_kwargs["max_tokens"] = resolved_max_tokens
if resolved_temperature is not None:
    agent_kwargs["temperature"] = resolved_temperature
```

src/conductor/providers/hermes.py
```diff
@@ -0,0 +1,315 @@
...
+        if self._api_key:
+            agent_kwargs["api_key"] = self._api_key
+        # Map Conductor tools field to hermes enabled_toolsets.
+        # tools=None → omit (hermes default: all tools)
+        # tools=[]   → enabled_toolsets=[] (empty allowlist = 0 tools)
+        # tools=[..] → enabled_toolsets=[..] (hermes toolset names)
+        if agent.tools is not None:
+            agent_kwargs["enabled_toolsets"] = agent.tools
+
+        loop = asyncio.get_event_loop()
```
jrob5756: **SHOULD-FIX — `asyncio.get_event_loop()` is deprecated inside coroutines**

Since 3.10 this emits `DeprecationWarning` if there is no running loop, and is slated for removal in 3.14. Since this function is `async def`, a running loop is guaranteed — use the explicit accessor:

```python
loop = asyncio.get_running_loop()
```

(Also caught by some default lint configs, so this will eventually be a CI failure.)

src/conductor/providers/hermes.py
```diff
@@ -0,0 +1,315 @@
...
+        # Parse structured output or wrap as plain text
+        if agent.output:
+            content = parse_json_output(final_response)
+            validate_output(content, agent.output)
+        else:
+            content = {"text": final_response}
+
+        # Populate token counts from the result dict when available
+        input_tokens: int | None = result.get("input_tokens") or result.get("prompt_tokens")
+        output_tokens: int | None = result.get("output_tokens") or result.get("completion_tokens")
```
jrob5756: **SHOULD-FIX — `or` fallback chain collapses legitimate zero into the wrong value**

`result.get("input_tokens") or result.get("prompt_tokens")` treats `0` as falsy. If hermes-agent ever returns `{"input_tokens": 0, "prompt_tokens": 5}` (e.g. a cache-hit case where input tokens are already counted elsewhere), this silently picks `5` instead of `0`.

Defensive fix:

```python
input_tokens = result.get("input_tokens")
if input_tokens is None:
    input_tokens = result.get("prompt_tokens")
output_tokens = result.get("output_tokens")
if output_tokens is None:
    output_tokens = result.get("completion_tokens")
```

Bonus: this also makes the intent (`None`-fallback, not falsy-fallback) self-documenting.

src/conductor/providers/hermes.py
```diff
@@ -0,0 +1,315 @@
...
+    async def validate_connection(self) -> bool:
+        """Verify the hermes-agent library is importable and functional.
+
+        Performs a lightweight check by importing the library. Does not
+        make any API calls — hermes uses the caller's ambient API keys.
+
+        Returns:
+            True if the hermes-agent library is available, False otherwise.
+        """
+        return HERMES_SDK_AVAILABLE
```
jrob5756, line 466: **SHOULD-FIX — Docstring misrepresents what this method does**

The docstring says *"Performs a lightweight check by importing the library"* — but the import happens at module load (lines 30-36) and is reflected in the module-level `HERMES_SDK_AVAILABLE` constant. By the time `validate_connection()` is called, `__init__` has already raised `ProviderError` if the SDK is missing, so this method is functionally `return True`. No "check" occurs inside the method body.

Consequence: a bad `base_url` (typo, dead endpoint, wrong port) is not caught at `validate` time — it surfaces only when the first agent runs, possibly minutes into a long workflow.

Two paths forward:

**Option A** (low-effort, correct docstring):
```python
async def validate_connection(self) -> bool:
    """Confirms the hermes-agent SDK is importable.

    The import is performed at module load and reflected in the
    module-level ``HERMES_SDK_AVAILABLE`` constant; this method
    simply returns that value. Does NOT verify the configured
    ``base_url``/``api_key`` are reachable or valid — credential
    and endpoint failures surface only at first agent execution.
    """
    return HERMES_SDK_AVAILABLE
```

**Option B** (real probe): Do a minimal API call (1-token completion or `httpx.head(base_url)` if set) so misconfigured endpoints fail at `conductor validate` instead of mid-workflow.

src/conductor/providers/hermes.py
```diff
@@ -0,0 +1,315 @@
...
+    from conductor.config.schema import AgentDef
+
+# Try to import the hermes-agent SDK
+try:
+    from run_agent import AIAgent
+
+    HERMES_SDK_AVAILABLE = True
+except ImportError:
+    HERMES_SDK_AVAILABLE = False
+    AIAgent = None  # type: ignore[misc, assignment]
```
jrob5756: **SHOULD-FIX — Two concerns with the import block**

**(a) `run_agent` is an extremely generic top-level package name.** It will conflict with any other PyPI / local package named `run_agent`, and a user who runs `pip install hermes-agent` and gets `ModuleNotFoundError: No module named 'run_agent'` will be confused (the error and the documented install command don't share a name). If hermes exposes a qualified path (`hermes_agent.run_agent.AIAgent` or similar), prefer it. At minimum, add a comment here noting that `run_agent` is the actual top-level module shipped by the `hermes-agent` package.

**(b) Broad `except ImportError` masks unrelated failures.** If hermes-agent itself has a missing dependency (e.g. an internal `import openai` fails because of a version conflict), the user sees the misleading *"Hermes provider requires the hermes-agent package"* error instead of the real root cause. Tighten:

```python
try:
    from run_agent import AIAgent
    HERMES_SDK_AVAILABLE = True
except ModuleNotFoundError as e:
    if e.name in ("run_agent", "hermes_agent"):
        HERMES_SDK_AVAILABLE = False
        AIAgent = None  # type: ignore[misc, assignment]
    else:
        raise  # surface real dependency problems
```

src/conductor/config/schema.py
```diff
@@ -1373,10 +1374,11 @@ def _check_field_compatibility(self) -> ProviderSettings:
...
             if self.base_url is not None or self.api_key is not None:
-                raise ValueError(
-                    f"Structured provider config (base_url/api_key) for name='{self.name}' "
-                    "is not yet implemented; use environment variables for the underlying SDK."
-                )
+                if self.name not in ("copilot", "hermes"):
+                    raise ValueError(
+                        f"Structured provider config (base_url/api_key) for name='{self.name}' "
+                        "is not yet implemented; use environment variables for the underlying SDK."
+                    )
```
jrob5756: **SHOULD-FIX — Dead `"copilot"` check, and an allowlist that doesn't scale**

Two issues here:

**(a) The inner `"copilot"` membership is unreachable.** This entire block lives inside `if self.name != "copilot":` (line 1369), so `self.name` is already known not to be `"copilot"`. The condition `self.name not in ("copilot", "hermes")` is effectively `self.name != "hermes"`. Simplify:

```python
if self.base_url is not None or self.api_key is not None:
    if self.name != "hermes":
        raise ValueError(
            f"Structured provider config (base_url/api_key) for name='{self.name}' "
            "is not yet implemented; use environment variables for the underlying SDK."
        )
```

**(b) String allowlist won't scale.** Hard-coding `("copilot", "hermes")` here means every new provider that supports structured routing requires a schema edit. After rebase, the cleaner pattern is to add a capability field (e.g. `supports_structured_routing: bool`) to `ProviderCapabilities` and look it up:

```python
from conductor.providers.capabilities import get_capabilities

if (self.base_url is not None or self.api_key is not None):
    if not get_capabilities(self.name).supports_structured_routing:
        raise ValueError(...)
```

This keeps the universal schema validator provider-agnostic and moves the per-provider truth back where it belongs. Could be a follow-up PR if too much for this one — but the simplified version (a) should land now.

docs/providers/hermes.md
```diff
@@ -0,0 +1,273 @@
+# Hermes Provider Documentation
+
+The Hermes provider enables Conductor workflows to use the [NousResearch hermes-agent](https://github.com/NousResearch/hermes-agent) library. Hermes is an agentic AI framework that manages its own tool ecosystem and supports models from multiple providers via OpenRouter-style model identifiers.
```
jrob5756, line 10: **SHOULD-FIX — Missing experimental disclaimer**

After rebase, this provider will be in the "experimental" tier per the framework added in `ca73f56`. Add an admonition at the top of the page linking to `docs/providers/experimental.md` so users understand the stability contract before they invest in building workflows around it.

Suggested:

```markdown
# Hermes Provider Documentation

> ⚠️ **Experimental Provider** — Hermes delegates the agentic loop to an
> upstream SDK and has documented parity carve-outs (no MCP, no streaming
> events, no reasoning-effort control, prompt-injection structured output).
> See [Experimental Providers](experimental.md) for the full stability
> policy and promotion criteria.

The Hermes provider enables ...
```

The same disclaimer (or a similar reference) should also appear in:
- `docs/providers/comparison.md` — add a "Tier" row to the Quick Comparison table
- `README.md` — the Hermes mention in the features section
- `docs/configuration.md` — the Hermes Provider subsection
- `plugins/conductor/skills/conductor/SKILL.md` and the three skill references — these feed Conductor's authoring agents

(See the main review for the full list.)

jrob5756 (commented):
Thanks for the. work on this PR. I ran a multi-aspect review (code quality, tests, comments/docs, silent failures, type design, dead code) and the implementation shows good architectural instincts — the sync/async bridge is clean, per-call `AIAgent` instantiation correctly avoids session-state leakage, and the tri-state `tools` contract is well-documented.

That said, the PR predates a significant framework change on main and has several issues that need to land before merge. Inline comments below cover the specifics; this top comment focuses on the two structural action items.

---

### 1. Rebase onto current `main` (required)

This PR was opened against `main` at `085b7a51` but main has since advanced by 3 commits:

| Commit | Why it matters |
|---|---|
| `5a5b550` `feat(providers): add claude-agent-sdk provider (#104)` | New experimental provider lands |
| `351ce40` `fix(human_gate): nest prompt_for collected values under additional_input (#237)` | Unrelated |
| `ca73f56` `feat(providers): experimental provider tier + claude-agent-sdk follow-ups (#241) (#242)` | **Adds `ProviderCapabilities` framework — see §2** |

Suggested rebase:

```bash
git fetch origin main
git checkout feat/hermes-provider
git rebase origin/main
# Resolve conflicts (likely in factory.py and schema.py around the new "hermes" branch)
git push --force-with-lease
```

After the rebase, `__init_subclass__` in `AgentProvider` will refuse to import `HermesProvider` until a `CAPABILITIES` attribute is declared (§2 below). Run `uv run pytest tests/test_providers/test_hermes.py -x` immediately after rebase to confirm.

---

### 2. Declare Hermes as an "experimental" provider

After ca73f56, every concrete `AgentProvider` subclass **must** declare a class-level `CAPABILITIES: ProviderCapabilities` attribute. The framework formalizes carve-outs for providers that delegate part of the agentic loop to an upstream SDK — and Hermes is exactly the kind of provider it was built for. See `docs/providers/experimental.md` on main for the full policy and `src/conductor/providers/claude_agent_sdk.py` for a reference declaration.

**Two changes required after rebase:**

#### a) Register Hermes in the capability resolver

Add to `src/conductor/providers/capabilities.py` in `_PROVIDER_CLASS_PATHS`:

```python
_PROVIDER_CLASS_PATHS: Final[dict[str, str]] = {
    "copilot": "conductor.providers.copilot:CopilotProvider",
    "claude": "conductor.providers.claude:ClaudeProvider",
    "claude-agent-sdk": "conductor.providers.claude_agent_sdk:ClaudeAgentSdkProvider",
    "hermes": "conductor.providers.hermes:HermesProvider",  # ← add this
}
```

#### b) Add the `CAPABILITIES` declaration to `HermesProvider`

The carve-outs map directly from the limitations you already documented in the PR description and `docs/providers/hermes.md`. Suggested values based on the current implementation:

```python
from conductor.providers.capabilities import ProviderCapabilities

class HermesProvider(AgentProvider):
    CAPABILITIES = ProviderCapabilities(
        tier="experimental",
        # No translation from Conductor's mcp_servers config to hermes — hermes
        # manages its own tool ecosystem with no MCP equivalent.
        mcp_tools=False,
        # `agent.tools` and hermes `enabled_toolsets` are different vocabularies
        # (Conductor allowlist vs hermes-internal toolset names). Declaring
        # False forces the validator to reject silent reinterpretation —
        # see the inline comment on the passthrough at hermes.py:186-187.
        workflow_tools_passthrough=False,
        # Synchronous `run_conversation()`; events fire only after completion.
        streaming_events=False,
        # No agent_reasoning events emitted.
        agent_reasoning_events=False,
        # `default_reasoning_effort` accepted for factory parity but discarded.
        reasoning_effort=None,
        # `_JSON_INSTRUCTION` appended to the prompt; not native schema enforcement.
        structured_output="prompt_injection",
        # Interrupt cancels the wrapping Future, but `run_in_executor` cannot
        # actually stop the thread — the LLM call continues to natural completion
        # in the background. Declaring False is more honest than claiming True;
        # if you'd rather declare True, document the "best-effort" semantics
        # prominently in the docstring and the experimental banner.
        interrupt=False,
        # `asyncio.wait(timeout=...)` enforces the wall-clock cap (with the same
        # caveat as `interrupt` — the thread continues burning tokens).
        max_session_seconds=True,
        # Stateless; no session_id to persist.
        checkpoint_resume=False,
        # Token fields are optional in the hermes result dict and frequently
        # absent. Set to False unless you tighten the result handling to require them.
        usage_tracking=False,
        # Per-call `AIAgent(**kwargs)` instantiation — verify hermes-agent has
        # no module-level shared mutable state before keeping this True.
        concurrent_safe=True,
        upstream_pin="hermes-agent",
        maintainer="(community contribution)",
    )
```

#### c) Add the experimental banner to user-facing docs

Several doc surfaces currently present Hermes as a peer to Copilot/Claude with no stability disclaimer. After rebase, add an "Experimental" admonition (linking to `docs/providers/experimental.md`) to:

- `docs/providers/hermes.md` (top of page)
- `docs/providers/comparison.md` (add a "Tier" row to the Quick Comparison table)
- `README.md` (the Hermes mention in the features section)
- `docs/configuration.md` (the Hermes Provider subsection)
- `plugins/conductor/skills/conductor/SKILL.md` and the three skill references — these are consumed by the Conductor authoring agents, so an experimental marker here prevents agents from treating Hermes as production-grade.

---

### Summary of inline comments

**Blockers** (must fix before merge):
1. Missing `CAPABILITIES` declaration (rebase blocker — see §2 above)
2. `agent.tools` → `enabled_toolsets` silent vocabulary reinterpretation
3. Raw third-party exceptions escape unwrapped from `_run_sync`
4. Interrupt cancellation does not actually stop the executor thread
5. `_fire` swallows callback exceptions at DEBUG (invisible by default)
6. Hermes factory branch has zero test coverage

**Should-fix** (before merge):
7. `default_reasoning_effort` accepted then dropped (not even stored on `self`)
8. Per-agent `max_tokens` / `temperature` overrides silently dropped (parity break)
9. `asyncio.get_event_loop()` deprecated inside coroutines
10. `validate_connection()` docstring misrepresents what the method does
11. Dead `"copilot"` check in nested schema guard
12. Token-count `or` fallback collapses legitimate zero
13. `from run_agent import AIAgent` — generic package name; broad `ImportError` masks unrelated import failures

Smaller suggestions and test-hardening recommendations are deferred to a post-merge follow-up.

tima-110:
> Thanks for the. work on this PR. I ran a multi-aspect review (code quality, tests, comments/docs, silent failures, type design, dead code) and the implementation shows good architectural instincts — the sync/async bridge is clean, per-call `AIAgent` instantiation correctly avoids session-state leakage, and the tri-state `tools` contract is well-documented.
> 
> That said, the PR predates a significant framework change on main and has several issues that need to land before merge. Inline comments below cover the specifics; this top comment focuses on the two structural action items.
> 

Jason,
Thank you for the detailed review.  I will get started on this.
T

src/conductor/providers/hermes.py
tima-110: Blocker Resolved

  1. CAPABILITIES declaration at line 72 of hermes.py with a full ProviderCapabilities object
  2. Registration in _PROVIDER_CLASS_PATHS at line 207 of capabilities.py mapping "hermes" to "conductor.providers.hermes:HermesProvider".

src/conductor/providers/hermes.py
tima-110: Fixed in 32bac18. call_task.result() is now wrapped (lines 356–364):
  ▎
  ▎ try:
  ▎     result = call_task.result()
  ▎ except asyncio.CancelledError:
  ▎     raise
  ▎ except Exception as e:
  ▎     raise ProviderError(
  ▎         f"Hermes agent execution failed (model='{resolved_model}'): {e}",
  ▎         suggestion="Check the hermes-agent logs and verify base_url/api_key.",
  ▎     ) from e
  ▎
  ▎ All enumerated leak paths (openai.AuthenticationError, httpx.ConnectError, KeyError, ValueError, json.JSONDecodeError) are now caught, wrapped in ProviderError with model context and actionable suggestion, and chained via from e for traceback preservation. The
  ▎ interrupt/timeout block above (lines 339–354) also provides is_retryable classification as requested.

src/conductor/providers/hermes.py
tima-110: Fixed in 32bac18. _fire log level raised from DEBUG to WARNING (line 548):
 
  except Exception:
      logger.warning("Error in event_callback for %s", event, exc_info=True)
 
  Broken event sinks (schema mismatches, serialization errors, dashboard disconnects) are now visible at default log levels with full traceback via exc_info=True. Operators will see the failure immediately rather than experiencing silent total observability loss.

src/conductor/providers/hermes.py
tima-110: Resolved. workflow_tools_passthrough is now False in the CAPABILITIES declaration. The provider follows the same pattern as claude_agent_sdk.py:

  - tools=None (omitted) → uses hermes defaults (all available toolsets), or the provider-level hermes_toolsets if configured
  - tools=[] → explicitly disables all tools (enabled_toolsets=[])
  - tools=[non-empty] → raises ProviderError explaining the vocabulary mismatch

  For authors who need to restrict Hermes toolsets, a new hermes_toolsets field on ProviderSettings provides the knob:

  runtime:
    provider:
      name: hermes
      hermes_toolsets: [filesystem, web]

  The import succeeds cleanly and conductor validate now catches per-agent tools: misuse at validation time rather than silently passing through incorrect tool names.

src/conductor/providers/hermes.py
tima-110: ## BLOCKER — Raw third-party exceptions escape unwrapped

**Fixed** in `32bac18`. `call_task.result()` is now wrapped (lines 356–364):

```python
try:
    result = call_task.result()
except asyncio.CancelledError:
    raise
except Exception as e:
    raise ProviderError(
        f"Hermes agent execution failed (model='{resolved_model}'): {e}",
        suggestion="Check the hermes-agent logs and verify base_url/api_key.",
    ) from e
```

All enumerated leak paths (`openai.AuthenticationError`, `httpx.ConnectError`, `KeyError`, `ValueError`, `json.JSONDecodeError`) are now caught, wrapped in `ProviderError` with model context and actionable suggestion, and chained via `from e` for traceback preservation. The interrupt/timeout block above (lines 339–354) also provides `is_retryable` classification as requested.

---

## BLOCKER — Callback exceptions silently swallowed at DEBUG level

**Fixed** in `32bac18`. `_fire` log level raised from `DEBUG` to `WARNING` (line 548):

```python
except Exception:
    logger.warning("Error in event_callback for %s", event, exc_info=True)
```

Broken event sinks (schema mismatches, serialization errors, dashboard disconnects) are now visible at default log levels with full traceback via `exc_info=True`. Operators will see the failure immediately rather than experiencing silent total observability loss.

---

## BLOCKER — Interrupt/timeout issues (two parts)

**Fixed** across commits `32bac18` and `2e5cb49`.

**(a) Interrupt cannot actually stop the executor thread** — Resolved by implementing cooperative interrupt via `hermes_agent.interrupt()` (`2e5cb49`). The fix keeps a reference to the `AIAgent` instance created inside `_run_sync` (line 302), and when interrupt/timeout fires, calls `hermes_agent_ref[0].interrupt()` (line 343) before cancelling the future. `AIAgent.interrupt()` is a thread-safe method that sets `_interrupt_requested`, which the hermes conversation loop checks between each API call and tool iteration — the agent stops cooperatively rather than running to natural completion. `interrupt=True` in CAPABILITIES is now accurate. This is the reviewer's option (2): hermes-agent does expose a cooperative cancellation hook, and we use it.

**(b) Neither ProviderError sets `is_retryable`** — Fixed in `32bac18`, all four paths now classified:

| Path | `is_retryable` | Rationale |
|------|----------------|-----------|
| Interrupt | `False` (line 348) | User explicitly stopped |
| Timeout | `True` (line 353) | Next attempt might be faster |
| `result.get("failed")` | `False` (line 369) | Library-level error, not transient |
| `final_response is None` | `False` (line 376) | Degraded response, not a transient network issue |

> **Note:** the `final_response is None` classification differs from the reviewer's suggestion of `is_retryable=True`. This is intentional — with cooperative interrupt now in place, a missing final response indicates the model genuinely produced no output (not a mid-stream network drop), so retrying is unlikely to help.

src/conductor/providers/hermes.py
tima-110: Resolved — `default_reasoning_effort` is stored on `self` (line 145) and actively wired through to hermes via `reasoning_config` in the execute path (lines 234-236). This isn't a no-op: when set, it produces `agent_kwargs["reasoning_config"] = {"effort": effort}` which
  is forwarded to the `AIAgent` constructor. The docstring was also updated to reflect this ("Forwarded to hermes via `reasoning_config`"). Neither Option 1 nor Option 2 from the review applies since the value is genuinely used, not discarded.

src/conductor/providers/factory.py
tima-110: **Fixed.** Original `32bac18` added 3 tests covering SDK-unavailable, basic creation, and partial config forwarding. Missing cases now added in `tests/test_providers/test_factory.py`:

| Case | Test |
|------|------|
| 1. `HERMES_SDK_AVAILABLE=False` → `ProviderError` | `test_factory_raises_when_sdk_not_available` |
| 2. `provider_settings.name == "hermes"` → `base_url` / `api_key.get_secret_value()` / `hermes_home` / `hermes_toolsets` extraction | `test_factory_extracts_provider_settings_for_hermes` |
| 3. `provider_settings.name != "hermes"` → all hermes-specific fields stay `None` | `test_factory_ignores_non_hermes_provider_settings` |
| 4. All constructor params forwarded (`model`, `max_tokens`, `temperature`, `max_agent_iterations`, `max_session_seconds`, `default_reasoning_effort`) | `test_factory_passes_all_config_to_hermes` |

The `SecretStr` → `get_secret_value()` path is now explicitly tested: `test_factory_extracts_provider_settings_for_hermes` passes `api_key=SecretStr("sk-test-key")` in `ProviderSettings` and asserts the provider receives the unwrapped `"sk-test-key"` string. A regression that drops `.get_secret_value()` will fail this test.

All 54 tests pass (19 factory + 35 hermes, 5 skipped).

src/conductor/providers/hermes.py
tima-110: Resolved — but not via the per-agent override path suggested.

  My research suggests that `AgentDef` does not expose `temperature` or `max_tokens` fields (it uses `extra="forbid"`, so specifying them on an agent would fail Pydantic validation before
  reaching any provider). These are runtime-level settings only.

  However, investigating this did surface a bug: `AIAgent.__init__` does NOT accept `temperature` as a direct keyword argument (it would raise `TypeError`). It accepts `max_tokens` directly, but
  temperature must go through `request_overrides` — a dict that gets applied at the transport layer.

  Fixed in 5529a1c: temperature is now routed via `agent_kwargs.setdefault("request_overrides", {})["temperature"]`, which is how hermes handles temperature internally (via `request_overrides →
  chat_completions transport → api_kwargs["temperature"]`).

  No per-agent override support was added because the schema doesn't define per-agent `temperature`/`max_tokens` fields — that would be a separate feature request requiring schema changes across all
  providers.

src/conductor/providers/hermes.py
tima-110: Already resolved — the code uses `asyncio.get_running_loop()` (hermes.py:288). No instances of `asyncio.get_event_loop()` exist in the source tree.

src/conductor/providers/hermes.py
tima-110: Already resolved — the code uses explicit `None` checks, not `or` fallback (hermes.py:400-405):

  ```python
  input_tokens: int | None = result.get("input_tokens")
  if input_tokens is None:
      input_tokens = result.get("prompt_tokens")
  output_tokens: int | None = result.get("output_tokens")
  if output_tokens is None:
      output_tokens = result.get("completion_tokens")

  A legitimate 0 is preserved correctly.

src/conductor/providers/hermes.py
tima-110: ⏺ Both concerns are already addressed:

  1. (a) Comment at line 32-35 documents that run_agent is the actual top-level module shipped by hermes-agent.
  2. (b) Catches ModuleNotFoundError (not broad ImportError), checks _e.name matches run_agent, and re-raises otherwise.

  Already resolved — both concerns are addressed in the current code (hermes.py:32-45):

  (a) A comment documents that `hermes-agent` ships its public API under the top-level module name `run_agent`:
  ```python
  # The hermes-agent package ships its public API under the top-level module
  # name "run_agent" (not "hermes_agent").

  (b) The except clause catches only ModuleNotFoundError (not broad ImportError) and checks _e.name before suppressing:
  except ModuleNotFoundError as _e:
      if _e.name is not None and _e.name.split(".")[0] == "run_agent":
          HERMES_SDK_AVAILABLE = False
          AIAgent = None
      else:
          raise

  Unrelated dependency failures (e.g. missing openai inside hermes) propagate with their real traceback.
  ```

docs/providers/hermes.md
tima-110: disclaimer added

src/conductor/providers/hermes.py
tima-110: Already resolved (Option A) — the docstring was corrected to accurately describe the method's behavior (hermes.py:437-444):

  ```python
  async def validate_connection(self) -> bool:
      """Confirms the hermes-agent SDK is importable.

      The import is performed at module load and reflected in the
      module-level ``HERMES_SDK_AVAILABLE`` constant; this method
      simply returns that value. Does NOT verify the configured
      ``base_url``/``api_key`` are reachable — credential and endpoint
      failures surface only at first agent execution.
      """
      return HERMES_SDK_AVAILABLE

  Option B (real probe) was not implemented — hermes doesn't expose a lightweight health-check API, and a 1-token completion would incur cost/latency at validate time. This matches the Claude Agent SDK
  provider's pattern where connection validation is also deferred to first execution.
  ```

src/conductor/config/schema.py
tima-110: Already resolved — the code (schema.py:1391-1403) is already in the simplified form:

  ```python
  if self.name != "copilot":
      # ... copilot-only fields check ...
      if self.base_url is not None or self.api_key is not None:
          if self.name != "hermes":
              raise ValueError(...)
```
  There's no self.name not in ("copilot", "hermes") pattern — the outer if self.name != "copilot": guard already excludes copilot, and the inner if self.name != "hermes": is the only additional check
  needed. The logic is correct as-is.

  Regarding (b) — adding supports_structured_routing to ProviderCapabilities to avoid the string check: agreed this would be cleaner as we add more providers, but today only copilot and hermes support it,
  so the current explicit checks are adequate. Tracked as a follow-up for when a third provider needs structured routing.

**Comment by tima-110** [comment]
@jrob5756 The changes have been made.  I have peformed some functional testing using a working hermes agent environment, and I think things are working.

**Comment by tima-110** [comment]
@jrob5756. sorry forgot to request the review.  Let me know if i need to rebaseline

File: src/conductor/providers/hermes.py [hunk 15]
```diff
@@ -0,0 +1,625 @@
...
+
+    CAPABILITIES = ProviderCapabilities(
+        tier="experimental",
+        mcp_tools=False,
+        workflow_tools_passthrough=False,
+        streaming_events=True,
+        agent_reasoning_events=True,
+        reasoning_effort=("low", "medium", "high", "xhigh"),
+        structured_output="prompt_injection",
+        interrupt=True,
```
jrob5756: interrupt=True is the one capability the experimental tier treats as a hard contract, and the implementation doesn't match it. The descriptor defines interrupt=True as "can return partial output mid-execution," but the interrupt path raises ProviderError (around line 368) and never returns a partial AgentOutput. That is exactly the interrupt=False carve-out behavior described in experimental.md. Either return the partial hermes result on interrupt, or declare the weaker value:

```suggestion
        interrupt=False,
```

src/conductor/providers/hermes.py
```diff
@@ -0,0 +1,625 @@
...
+    ) -> dict[str, Any]:
+        """Parse response as JSON, retrying via conversation if parsing fails."""
+        last_error: str | None = None
+
+        for attempt in range(_MAX_PARSE_RECOVERY_ATTEMPTS + 1):
+            try:
+                content = parse_json_output(response)
+                validate_output(content, agent.output)  # type: ignore[arg-type]
+                return content
+            except (json.JSONDecodeError, ValueError, ValidationError) as e:
```
jrob5756, line 493: This except only catches parse errors, so a library/API/network error raised by the recovery run_conversation just below (an upstream APIError, a connection reset) propagates straight out of execute() unwrapped. That reopens the "raw exceptions escape" hole from the first review: it was fixed on the primary call path (the try/except around call_task.result()) but not on this second path. Wrap the recovery run_conversation and its result.get access and re-raise anything non-parse as ProviderError, so the class docstring contract ("Library failures, API errors ... -> ProviderError") holds here too.

src/conductor/providers/hermes.py
```diff
@@ -0,0 +1,625 @@
...
+                # Use the messages from the failed run as history
+                history = messages if messages else conversation_history
+
+                recovery_kwargs = {
+                    k: v
+                    for k, v in agent_kwargs.items()
+                    if k not in ("stream_delta_callback", "reasoning_callback")
+                }
+                hermes_agent = AIAgent(**recovery_kwargs)  # ty: ignore[call-non-callable]
+                recovery_result = hermes_agent.run_conversation(
```
jrob5756: This recovery call runs synchronously on the event loop. The primary call correctly goes through loop.run_in_executor, but _parse_with_recovery is invoked directly from async execute(), so each recovery attempt blocks the loop for a full model round-trip, up to three times. With concurrent_safe=True advertised and ParallelGroup usage, that stalls every other agent plus interrupt monitoring. Route this through run_in_executor like the primary path. Secondary: the recovery path also skips the hermes_home ContextVar override and ignores interrupt_signal/timeout, so a recovery run can load the wrong profile and can't be cancelled.

jrob5756 (commented):
Thanks for the updates! I have a few more items.

As a heads up, so this doesn't read as re-litigating round one: these are in code that landed after the previous review :) . The recovery loop, the cooperative interrupt, the streaming callbacks, and the session persistence were all added in the remediation commits, so they're new surface area introduced by the fixes rather than old findings resurfacing. 

The three items to address are inline below. The lowest-risk route to green is probably to right-size the descriptor (interrupt=False) and either run the recovery call through run_in_executor with ProviderError wrapping, or drop the recovery loop in favor of single-shot structured output.

Thanks again!

tima-110:
> Thanks for the updates! I have a few more items.
> 
> As a heads up, so this doesn't read as re-litigating round one: these are in code that landed after the previous review :) . The recovery loop, the cooperative interrupt, the streaming callbacks, and the session persistence were all added in the remediation commits, so they're new surface area introduced by the fixes rather than old findings resurfacing.
> 
> The three items to address are inline below. The lowest-risk route to green is probably to right-size the descriptor (interrupt=False) and either run the recovery call through run_in_executor with ProviderError wrapping, or drop the recovery loop in favor of single-shot structured output.
> 
> Thanks again!

Hi JRob ... no worries!  didn't think this was re-litigation.  Alot was done in the last round, so I expected the feedback :).  Appreciate the support.  Will get started on these.

Tim

src/conductor/providers/hermes.py
tima-110: Fixed. The recovery run_conversation call is now wrapped — parse errors re-raise normally (so the retry loop continues), but any library/API/network error is caught and wrapped in ProviderError. Tests pass.

src/conductor/providers/hermes.py
tima-110: ▎ Fixed — _parse_with_recovery is now async and routes each recovery call through loop.run_in_executor, matching the primary call path. The recovery closure also applies the hermes_home ContextVar override (same pattern as _run_sync) and wraps non-parse exceptions in ProviderError. Full interrupt/timeout racing on individual recovery attempts was omitted since they're short JSON-fixup calls and the executor approach already unblocks the loop for other agents and interrupt monitoring.

src/conductor/providers/hermes.py
tima-110: Fixed — changed to interrupt=False. The current implementation raises ProviderError on interrupt rather than returning partial output, which matches the interrupt=False carve-out behavior. If we later add proper partial-result handling (awaiting hermes's response after interrupt() and returning AgentOutput(partial=True)), we can promote this back to True.

PR #237, opened 2026-06-01
fix(human_gate): nest prompt_for collected values under additional_input

ldavidgomez:
## Bug

When a `human_gate` option uses `prompt_for` to collect free-text input, the collected value is stored in `GateResult.additional_input`, but the workflow context output previously flattened that value into the gate output root.

This meant downstream templates could not access the value through the intended nested form:

```jinja
{{ gate.output.additional_input.field }}
```

Previously, the value was exposed through the flattened path:

```jinja
{{ gate.output.field }}
```

That flat shape is unsafe because it can collide with reserved gate output keys. For example, `prompt_for: selected` could overwrite the selected option value.

## Root cause

`engine/workflow.py` stored the gate result using a dict spread:

```python
self.context.store(
    agent.name,
    {
        "selected": gate_result.selected_option.value,
        **gate_result.additional_input,
    },
)
```

This placed collected values at `gate.output.<field>` rather than `gate.output.additional_input.<field>`.

It also introduced a silent data-corruption risk: if `prompt_for` used the field name `selected`, the spread could overwrite the chosen option value with the user's raw text.

The `gate_resolved` event and `_wait_for_web_gate()` already use `additional_input` as a nested key, so the flat spread in workflow context storage was inconsistent with the rest of the implementation.

## Fix

Replace the flat spread with an explicit nested key:

```python
self.context.store(
    agent.name,
    {
        "selected": gate_result.selected_option.value,
        "additional_input": gate_result.additional_input,
    },
)
```

Gates without `prompt_for` continue to produce:

```json
{"additional_input": {}}
```

## Compatibility note

This changes the previously observable flat output shape for `prompt_for` values, for example:

```jinja
{{ gate.output.answer }}
```

That flat shape appears to be undocumented and unsafe because it can collide with reserved output keys such as `selected`.

This may affect workflows that were relying on the undocumented flattened form `gate.output.<prompt_for_field>`.

The canonical downstream reference after this fix is:

```jinja
{{ gate.output.additional_input.answer }}
```

## Tests added

| Test | What it covers |
|---|---|
| `test_human_gate_stores_additional_input_nested_in_context` | `output.additional_input.answer` is populated after `prompt_for` input |
| `test_human_gate_no_prompt_for_has_empty_additional_input` | Gates without `prompt_for` still produce `additional_input: {}` |
| `test_human_gate_prompt_for_named_selected_does_not_corrupt_selected` | `prompt_for: selected` does not overwrite the chosen option value |

## Tests run

```text
uv run pytest tests/test_engine/test_workflow.py -k "human_gate" -v
7 passed

uv run pytest tests/test_gates/ -v
48 passed
```

Total relevant validation:

```text
55 passed, 0 failed

tests/test_gates/test_dialog.py       18 passed
tests/test_gates/test_human.py        30 passed
tests/test_engine/test_workflow.py     7 passed
```

## Smoke result

CLI smoke confirmed:

```text
Gate output dump: {'output': {'selected': 'provide_answer', 'additional_input': {'answer': 'README.md'}}}

Selected: provide_answer
Answer: README.md
```

Confirmed template access:

```jinja
{{ ask_human.output.additional_input.answer }}
```

renders:

```text
README.md
```

## Known limitation

In `context: explicit`, the current input projection logic does not support nested shorthand declarations such as:

```yaml
input:
  - ask_human.additional_input.answer
```

The working explicit-mode declaration is to include the parent key:

```yaml
input:
  - ask_human.selected
  - ask_human.additional_input
```

and then let the template traverse into it:

```jinja
{{ ask_human.output.additional_input.answer }}
```

This is a generic context input-projection limitation, not specific to `human_gate` or `prompt_for`, and is intentionally left out of this PR to keep the change focused.


ldavidgomez:

PR #238, opened 2026-06-01
feat(instructions): honor scoped applyTo via CWD-implicit overlap (#231)

### LINKED ISSUE #231: Feature: Honor scoped `applyTo` in `.github/instructions/` via CWD-derived workspace scope
## Problem summary

Conductor v0.1.17's `--workspace-instructions` discovers `.github/instructions/**/*.instructions.md` but filters them through `_is_always_on_instructions_file`, which loads ONLY files with `applyTo: "**"` (`config/instructions.py:145-166`). Files with **any** scoped `applyTo` glob — `**/*.cs`, `tests/**`, `src/**`, `services/foo/**`, etc. — are silently skipped.

GitHub Copilot's documented `applyTo` semantic is *per-path scoping*: a file with `applyTo: "**/*.cs"` applies whenever any C# file is involved; one with `applyTo: "services/foo/**"` applies to that service's tree. These are correct uses of the convention, not misuse. Conductor's all-or-nothing filter doesn't honor them — it treats anything other than the literal `"**"` as "skip unconditionally," which excludes the bulk of real-world per-area instructions.

The gap applies to **any** repo that uses `applyTo` for scoping — single-service flat repos with per-language scoping, library repos splitting `src/` vs `examples/` vs `docs/`, and monorepos with per-service scoping.

### Where the filter lives

`src/conductor/config/instructions.py`:

```python
# L177
CONVENTIONS: list[Convention] = [
    ConventionFile("AGENTS.md"),
    ConventionFile(".github/copilot-instructions.md"),
    ConventionFile("CLAUDE.md"),
    ConventionDirectory(
        path=".github/instructions",
        pattern="*.instructions.md",
        include_file=_is_always_on_instructions_file,   # ← the filter
        recursive=True,
    ),
]

# L145
def _is_always_on_instructions_file(path: Path) -> bool:
    fm = _parse_frontmatter(path)
    if fm is None:
        return False
    return fm.get("applyTo") == "**"   # L166 — exact-equality "**" only
```

CWD is already available at the right seam: `cli/run.py:1223` passes `Path.cwd()` as `auto_discover_dir` to `discover_workspace_instructions`. The plumbing for CWD-implicit scoping exists; only the per-file filter needs to learn how to use it.

### Real-world data (Azure Chaos Studio)

Inventory of `.github/instructions/*.instructions.md` files in a real Azure repo. Most globs are language- or area-scoped rather than service-scoped — i.e., the same pattern that appears in single-service repos:

| File | applyTo | Loaded by conductor today? |
|---|---|---|
| `arm-rpc-guidelines.instructions.md` | `services/GW/**;services/BE/**` | ❌ |
| `csharp-coding-standards.instructions.md` | `**/*.cs` | ❌ |
| `eng_ms.instructions.md` | `/docs/eng.ms/*.*; /docs/eng.ms/**/*.*` | ❌ |
| `engineering-standards.instructions.md` | `**/*.cs,**/*.csproj,**/Directory.Packages.props,**/global.json,**/*.bicep,**/*.md` | ❌ |
| `poc-warning.instructions.md` | (uses non-standard `glob:` field, no `applyTo`) | ❌ |
| `portal-ux-patterns.instructions.md` | `**/portal-extension/**,**/portal-ux/**,**/ux-accelerator/**` | ❌ |
| `specs-arm-api.instructions.md` | `docs/specs/arm-api-published/*.*; docs/specs/arm-api-published/**/*.*` | ❌ |
| `testing-standards.instructions.md` | `**/*Tests.cs,**/*Tests/**,**/*UnitTests*/**,**/*HermeticTests*/**` | ❌ |
| `workload-identity.instructions.md` | `**/kubectl/*.yaml,**/kubectl/*.yml,**/identity.bicep,**/ev2/**` | ❌ |

Total: **52 KB / ~13K tokens. 0 / 9 files loaded today.** All 9 contain real engineering guidance (ARM RPC patterns, async/await rules, testing categories, AKS workload identity setup, etc.).

These globs are not misuse of `applyTo` — they correctly express the convention's per-path scoping. Promoting them all to `applyTo: "**"` would falsify the convention's intent and inject (e.g.) AKS workload-identity guidance into agents working on portal-extension code.

## Repro

```bash
mkdir repro && cd repro
git init -q
mkdir -p .github/instructions src tests

echo "Repo-wide rule." > .github/copilot-instructions.md

cat > .github/instructions/csharp.instructions.md <<'EOF'
---
description: C# coding standards
applyTo: "**/*.cs"
---
Use file-scoped namespaces.
EOF

cat > .github/instructions/testing.instructions.md <<'EOF'
---
description: Test-specific patterns
applyTo: "tests/**"
---
Tests must use the Arrange-Act-Assert pattern.
EOF

cat > wf.yaml <<'EOF'
workflow:
  name: repro
  entry_point: agent_a
agents:
  - name: agent_a
    prompt: "Echo your workspace_instructions."
    routes:
      - to: $end
EOF

conductor run --workspace-instructions wf.yaml
```

**Expected:** agent prompt's `<workspace_instructions>` block contains content from `csharp.instructions.md` and `testing.instructions.md`.
**Actual:** block contains only `copilot-instructions.md`. Both scoped files are silently skipped.

## Proposed design

Three pieces. The first is the correctness fix; the second is the minimum generalization to keep the seam clean for future scoped conventions; the third is inspectability so the implicit behavior is debuggable.

### Up-front honest framing

At repo-root CWD — which is the dominant case for programmatic launchers (skyship `Invoke-Ship.ps1`, octane `run-phases.py`, ATS-Copilot pr-reviewer, all of which `cd` to the user's repo root before invoking conductor) — bidirectional overlap means "every scoped file overlaps with the root subtree" → load everything. **At the root, this design is effectively "drop the `applyTo` filter."** That is strictly better than today's silent skip (correctness fix), but it is *not* token narrowing for programmatic callers. Narrowing only kicks in when a human `cd`s into a subdir (`cd services/GW/ && conductor run …`).

This is a deliberate tradeoff: **correctness over token economy for v1.** Programmatic narrowing can be tackled later if/when token bloat at the repo root becomes a measured problem.

### Piece 1 — CWD-implicit scope filter

Replace `_is_always_on_instructions_file`'s `applyTo == "**"` check with bidirectional glob-overlap against the existing `auto_discover_dir` (CWD).

**Algorithm:**

1. From CWD, walk up to find the workspace root (git root, or deepest ancestor with a known convention — same logic conductor already uses for discovery).
2. Compute CWD-relative-to-workspace-root (e.g., `services/GW/`, or `.` when CWD *is* the workspace root).
3. For each discovered instructions file, ask the convention for the scope glob (see Piece 2).
4. **Load the file if any of:**
   - The convention has no scope concept (e.g., `AGENTS.md`) — load unconditionally, matching today's behavior.
   - `extract_scope` returns `None` — equivalent to "always on" (covers both `applyTo: "**"` and unscoped frontmatter).
   - The scope glob's subtree **overlaps** with the CWD-relative subtree, in either direction.

**Bidirectional overlap.** `applyTo: services/GW/**` matches when CWD is `services/GW/src/` (user is *inside* the scope) **and** when CWD is `.` (the scope is *inside* the user's CWD subtree). Strict one-direction glob match would exclude one of these common cases. A "shares-any-literal-prefix-or-could-match-any-file-under-CWD" approximation is sufficient; precise glob intersection is undecidable in general but unnecessary here.

### Piece 2 — Minimal generalization of the existing seam

`ConventionDirectory.include_file: Callable[[Path], bool] | None` (line 77) is the existing pluggable filter slot — exactly the right place to extend. Two minimal changes:

```python
@dataclass(frozen=True)
class ConventionDirectory:
    path: str
    pattern: str
    extract_scope: Callable[[Path], str | None] | None = None  # NEW
    recursive: bool = True
    include_file: Callable[[Path], bool] | None = None  # retained briefly for back-compat
```

The convention author declares **only** how to read their scope field. The universal overlap test lives in core (one source of truth):

```python
# .github/instructions/ registration
ConventionDirectory(
    path=".github/instructions",
    pattern="*.instructions.md",
    extract_scope=_extract_apply_to,
)

def _extract_apply_to(path: Path) -> str | None:
    fm = _parse_frontmatter(path)
    if fm is None:
        return None  # no frontmatter → no scope claim → load
    apply_to = fm.get("applyTo")
    if apply_to is None or apply_to == "**":
        return None  # "always on" semantic → load
    return apply_to
```

`_walk_directory_convention` does the overlap test against `auto_discover_dir` once, generically:

```python
if convention.extract_scope is not None:
    scope = convention.extract_scope(file_path)
    if scope is not None and not _scope_overlaps(scope, cwd_rel_to_root):
        continue
```

When a second scoped convention shows up (e.g., Cursor's `.cursor/rules/*.mdc` with `globs:` frontmatter), it registers its own `extract_scope` — no new abstraction, no core changes, same overlap semantic.

**Not introducing:** a `ConventionHandler` Protocol / registry. With exactly one scoped convention today, the right shape for a registry is undervalidated; extending the existing `ConventionDirectory` slot is the YAGNI-honoring minimum. Revisit if a third scoped convention reveals a shape the dataclass can't accommodate.

### Piece 3 — `--print-loaded-instructions` (v1 deliverable)

Implicit behavior trades discoverability for ergonomics — same command in different directories loads different files. Ship inspectability alongside the implicit filter, not after.

```bash
conductor run --workspace-instructions --print-loaded-instructions wf.yaml
# prints the resolved list of loaded instruction files (with their effective
# scope and the overlap reason) to stderr before launching the workflow
```

Could also fold into existing `--verbose` output instead of a new flag if that's preferred — open to either. The requirement is that "why isn't my instruction loading" is answerable without instrumenting source code.

## What we're explicitly NOT doing

- **No new `--workspace-scope` flag.** The repeatable `--instructions <path>` flag (`cli/app.py:326-329`, `instructions.py:410-424`) already exists as the force-load escape hatch for callers that know exactly what they want. Adding a third knob would be one too many. Programmatic launchers that need finer narrowing than CWD can list specific files via `--instructions`, or `cd` first.
- **No `ConventionHandler` Protocol / registry.** YAGNI — extend the existing `ConventionDirectory` slot until a second scoped convention validates the registry shape.

## Principles satisfied

1. **Convention over configuration.** CWD is the universal CLI primitive every tool (git, grep, npm) already uses. No new concept for users.
2. **Single source of truth.** CWD is where the user "is"; no parallel flag to keep in sync.
3. **Faithful to the convention's own semantics.** `applyTo` is path-glob scoping; evaluating against actual filesystem paths is what the convention describes. The overlap test lives in core (one place), not duplicated across convention authors.
4. **Right architectural seam.** The fix lives in conductor's discovery layer (`config/instructions.py`), upstream of every provider adapter. One change, all providers and all callers benefit.
5. **Graceful degradation.** Conventions without a `extract_scope` keep working unchanged. New conventions plug in with one callable.
6. **Symmetric primitive.** Same mechanism (CWD) handles both ends of the scope spectrum: `cd repo-root` = broad, `cd services/GW/src/` = narrow.
7. **Composable with existing tools.** Anyone wrapping conductor in a script can narrow scope by `cd`-ing first. Standard Unix composition.
8. **YAGNI.** No `--workspace-scope` flag, no `ConventionHandler` registry — both can land later if real cases demand them.
9. **Backward compatibility.** `extract_scope` is a new optional field; `include_file` retained during migration. No existing convention author is broken.

## Honest tensions

1. **Implicit behavior trades discoverability for ergonomics.** Addressed by Piece 3 (`--print-loaded-instructions`) shipping in the same change, not as a follow-up.
2. **At repo-root CWD, this loads all scoped files.** Stated plainly above. Acceptable v1 tradeoff (correctness > token economy); programmatic narrowing can come later as `--workspace-scope` or similar if measured to matter.
3. **`extract_scope -> str | None` implicitly commits to "scope = path glob."** Today's two known scoped conventions (`applyTo`, Cursor's `globs:`) both fit. A wildly different scope semantic (RBAC role, language ID) would need a different mechanism — acceptable per YAGNI.
4. **Glob-overlap semantics need care.** The bidirectional-overlap predicate is more nuanced than strict glob match. Mitigation: an approximation (shared literal prefix + wildcard tolerance) is sufficient; precise intersection is unnecessary and undecidable in general.

## Comparison with the existing `--instructions <path>` flag

| Axis | `--instructions <path>` (today) | This proposal (CWD-implicit + `extract_scope`) |
|---|---|---|
| What you declare | Specific files to load | Nothing (CWD is implicit) |
| Who decides relevance | Caller (must enumerate files) | Conductor (consults each convention's scope predicate) |
| Convention semantics | Bypassed — files loaded unconditionally | Honored — `applyTo` / `globs` / etc. respected via convention's `extract_scope` |
| Per-call cardinality | N flags (one per file) | Zero — CWD-implicit |
| Caller knowledge required | File inventory + per-file relevance | Just `cd` to the right place |

**Complementary, not redundant.** `--instructions` remains as the force-load escape hatch ("ignore the convention, give me these files"). This proposal is the principled mechanism ("honor the convention; here's where I'm working").

## Integration impact: most callers need zero changes

Programmatic launchers across the Microsoft scenarios all already `cd` to the user's repo root before invoking conductor:

| Caller | How it invokes conductor today | Behavior after fix |
|---|---|---|
| Skyship `Invoke-Ship.ps1` (stack / fix-review) | `conductor run` from repo root | ✅ All `applyTo`-matching files load automatically (= correctness fix; not narrowing) |
| Octane `pr-orchestrator` `run-phases.py` | `conductor run` from repo root | ✅ Same |
| Octane `octane-workflow-implement` | `conductor run` from repo root | ✅ Same |
| ATS-Copilot pr-reviewer | `conductor run` from repo root | ✅ Same |
| Manual `conductor run` from SKILL.md docs | User's CWD | ✅ Honors wherever the user is |
| User `cd services/GW && conductor run …` | `services/GW/` | ✅ Narrows to GW automatically — no flag needed |

**No follow-up integration PRs needed in skyship / octane / ATS for the correctness fix.** The `--workspace-instructions` adoption PRs already shipped (or pending review) are exactly the right pre-work; the conductor fix lights them up.

## Affected users

Anyone using `--workspace-instructions` against a repo whose `.github/instructions/` files use scoped `applyTo` globs. Affected categories include single-service flat repos (`**/*.cs`, `tests/**`), library repos (`src/` vs `examples/` vs `docs/`), framework-scoped repos (`**/*.tsx`, `scripts/**`), and monorepos with per-service scoping (the most acute case).

Direct evidence: Azure Chaos Studio (9 files, 0 loaded today; 6/9 globs are language- or area-scoped and would appear identically in non-monorepo repos). Copilot Chat's docs explicitly recommend `applyTo` for per-area scoping in *any* repo, so this pattern is conventional, not anomalous.

franklixuefei:
Today `_is_always_on_instructions_file` only loads `.github/instructions/` files with `applyTo: "**"`, silently dropping every file with a scoped glob (`**/*.cs`, `services/foo/**`, etc.). This is the bug described in issue #231: real Azure repos have 0/9 instruction files loading despite their `applyTo` values being valid per GitHub's documented convention.

This change:

* Adds `ConventionDirectory.extract_scope: Callable[[Path], str | None]`, the convention-author-facing single-callable scope extractor. Returns `None` to opt out, `ALWAYS_ON_SCOPE` (`"**"`) for always-include, or any glob string for scoped evaluation. The bidirectional overlap test lives in conductor core so future scoped conventions (Cursor's `globs:`, Cline's future scope field, etc.) only register their own `extract_scope` — no duplication of the overlap semantic.
* Adds `_scope_overlaps` + `_single_glob_overlaps` (literal-prefix approximation) and the helpers `_normalize_scope_path`, `_apply_convention_filters`, `_discovery_reason`. Multi-glob `applyTo` values separated by `;` or `,` (both observed in real-world data) are supported; brace expansion is documented as unsupported.
* Replaces `_is_always_on_instructions_file` with `_extract_apply_to`. Files without `applyTo` (or with non-string `applyTo`) still opt out — matching today's behavior for those cases.
* Computes `cwd_rel` per walk level (relative to the convention's owner directory, not the workspace root) so nested `services/GW/.github/instructions/foo.md` with `applyTo: "src/**"` resolves `src` relative to `services/GW` — the closest-wins semantic that makes nested conventions meaningful.
* Adds `DiscoveredInstruction` + `discover_workspace_instructions_detailed` so callers can reason about why each file was included without re-parsing the filesystem.
* Adds `--print-loaded-instructions` CLI flag (plus full plumbing through `cli/run.py`, `cli/bg_runner.py`) that dumps the resolved list with scope + reason to stderr — for debugging "why isn't my instruction loading" without instrumenting source code.

Honest framing for the design tradeoff (per the issue body and Brent Rusinow's review feedback): at repo-root CWD (the dominant case for programmatic launchers like skyship / octane / ATS pr-reviewer), the bidirectional overlap loads every scoped file. That is strictly better than today's silent skip (correctness fix), but it is *not* token-bloat narrowing for programmatic callers. Narrowing only kicks in when a human `cd`s into a subdir. Correctness > token economy is the deliberate v1 tradeoff; programmatic narrowing can be revisited later if measured to matter.

Tests: 83 passed in `tests/test_config/test_instructions.py` covering the overlap helper (positive, disjoint, segment-boundary, multi-glob `;`/`,`, leading-`/`, Windows backslash, leading-`**` always-match, empty-separator-only rejection, documented brace-expansion limitation), the new `discover_workspace_instructions_detailed` shape, the `_walk_directory_convention` 3-tuple yield, root-CWD-loads-scoped and subdir-narrows-scoped end-to-end behaviors, owner-relative `cwd_rel` for nested conventions, and bg_runner forwarding of the new flag.

End-to-end repro from the issue body verified: a fresh repo with scoped `applyTo` files plus `.github/copilot-instructions.md` loads all 4 files at root CWD (vs. 1 today) and narrows correctly when invoked from a disjoint subdir.

Closes #231

### DIFF
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -1371,6 +1371,32 @@ async def _execute_with_stop_signal(
     raise ExecutionError(stop_message)
 
 
+def _emit_loaded_instructions_debug(start_dir: Path | None, enabled: bool) -> None:
+    """Discover-and-print the workspace instructions list when the debug flag
+    is enabled and auto-discovery actually ran.
+
+    Extracted from :func:`run_workflow_async` so the branch is directly
+    testable without spinning up the full workflow runner. ``start_dir`` must
+    be the same path that :func:`build_instructions_preamble` was given as
+    ``auto_discover_dir`` — sharing the path is what guarantees the printed
+    list cannot drift from what was actually loaded.
+
+    Args:
+        start_dir: The auto-discovery start directory used by the loader, or
+            ``None`` when auto-discovery did not run. ``None`` short-circuits
+            so the debug print is a no-op (matching the contract that this
+            flag is "meaningful only when --workspace-instructions is set").
+        enabled: Whether ``--print-loaded-instructions`` was passed on the
+            CLI. ``False`` short-circuits.
+    """
+    if not enabled or start_dir is None:
+        return
+    from conductor.config.instructions import discover_workspace_instructions_detailed
+
+    detailed = discover_workspace_instructions_detailed(start_dir)
+    _print_loaded_instructions(detailed)
+
+
 def _print_loaded_instructions(detailed: list[Any]) -> None:
     """Emit a human-readable summary of discovered workspace instruction files
     to stderr.
@@ -1574,13 +1600,7 @@ async def run_workflow_async(
             # stderr for debugging. Only meaningful when auto-discovery ran,
             # and must use the same start_dir as the loader above so the
             # printed list cannot silently drift from what was loaded.
-            if print_loaded_instructions and start_dir is not None:
-                from conductor.config.instructions import (
-                    discover_workspace_instructions_detailed,
-                )
-
-                detailed = discover_workspace_instructions_detailed(start_dir)
-                _print_loaded_instructions(detailed)
+            _emit_loaded_instructions_debug(start_dir, print_loaded_instructions)
 
         # Convert MCP servers from workflow config to SDK format
         mcp_servers = await _build_mcp_servers(config)
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -1551,8 +1551,16 @@ async def run_workflow_async(
         if workspace_instructions or cli_instructions or config.workflow.instructions:
             from conductor.config.instructions import build_instructions_preamble
 
+            # Compute the auto-discovery start dir once and share it between
+            # the loader and the --print-loaded-instructions debug dump. If
+            # these diverged (e.g. the cwd changed between the two calls, or
+            # one was passed a different path), the printed list would no
+            # longer reflect what was actually loaded — defeating the whole
+            # purpose of the debug flag.
+            start_dir = Path.cwd() if workspace_instructions else None
+
             instructions_preamble = build_instructions_preamble(
-                auto_discover_dir=Path.cwd() if workspace_instructions else None,
+                auto_discover_dir=start_dir,
                 yaml_instructions=config.workflow.instructions or None,
                 cli_instruction_paths=cli_instructions,
             )
@@ -1563,13 +1571,15 @@ async def run_workflow_async(
                 )
 
             # --print-loaded-instructions: dump the resolved discovery list to
-            # stderr for debugging. Only meaningful when auto-discovery ran.
-            if print_loaded_instructions and workspace_instructions:
+            # stderr for debugging. Only meaningful when auto-discovery ran,
+            # and must use the same start_dir as the loader above so the
+            # printed list cannot silently drift from what was loaded.
+            if print_loaded_instructions and start_dir is not None:
                 from conductor.config.instructions import (
                     discover_workspace_instructions_detailed,
                 )
 
-                detailed = discover_workspace_instructions_detailed(Path.cwd())
+                detailed = discover_workspace_instructions_detailed(start_dir)
                 _print_loaded_instructions(detailed)
 
         # Convert MCP servers from workflow config to SDK format
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -30,6 +30,7 @@ from conductor.mcp_auth import resolve_mcp_server_auth
 from conductor.providers.registry import ProviderRegistry
 
 if TYPE_CHECKING:
+    from conductor.config.instructions import DiscoveredInstruction
     from conductor.config.schema import ProviderSettings
     from conductor.events import WorkflowEvent
 
@@ -1397,7 +1398,7 @@ def _emit_loaded_instructions_debug(start_dir: Path | None, enabled: bool) -> No
     _print_loaded_instructions(detailed)
 
 
-def _print_loaded_instructions(detailed: list[Any]) -> None:
+def _print_loaded_instructions(detailed: list[DiscoveredInstruction]) -> None:
     """Emit a human-readable summary of discovered workspace instruction files
     to stderr.
 
@@ -1416,25 +1417,26 @@ def _print_loaded_instructions(detailed: list[Any]) -> None:
     plain-print rather than the rich console so it's reliably available in
     background/non-TTY launchers.
 
-    The parameter is typed as ``list[Any]`` rather than
-    ``list[DiscoveredInstruction]`` to avoid a hard import at module load
-    time — the import is deferred to the call site in
-    :func:`run_workflow_async` so the discovery code path isn't loaded when
-    the flag is unused.
+    ``DiscoveredInstruction`` is imported under ``TYPE_CHECKING`` only: this
+    module has ``from __future__ import annotations``, so the annotation is
+    never evaluated at import time, and the lazy import inside
+    :func:`_emit_loaded_instructions_debug` is what actually keeps the
+    discovery code path out of the module graph when the flag is unused.
     """
     import sys
+
+    from conductor.config.instructions import ALWAYS_ON_SCOPE
 
     if not detailed:
         print("[workspace-instructions] 0 files discovered from CWD.", file=sys.stderr)
         return
     print(
-        f"[workspace-instructions] {len(detailed)} file(s) loaded from CWD:",
+        f"[workspace-instructions] {len(detailed)} file(s) discovered from CWD:",
         file=sys.stderr,
     )
     for d in detailed:
-        # Use repr for the path so Windows paths render unambiguously.
         print(f"  {d.path}", file=sys.stderr)
-        scope_part = f"  applyTo={d.scope!r}" if d.scope and d.scope != "**" else ""
+        scope_part = f"  applyTo={d.scope!r}" if d.scope and d.scope != ALWAYS_ON_SCOPE else ""
         print(
             f"    source={d.source}  reason={d.reason}{scope_part}",
             file=sys.stderr,
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -1371,6 +1371,50 @@ async def _execute_with_stop_signal(
     raise ExecutionError(stop_message)
 
 
+def _print_loaded_instructions(detailed: list[Any]) -> None:
+    """Emit a human-readable summary of discovered workspace instruction files
+    to stderr.
+
+    Format:
+
+    .. code-block:: text
+
+        [workspace-instructions] 4 file(s) loaded from CWD:
+          AGENTS.md
+            source=AGENTS.md  reason=file-convention
+          .github/instructions/csharp-coding-standards.instructions.md
+            source=.github/instructions  reason=scope-overlap  applyTo='**/*.cs'
+          ...
+
+    Goes to stderr (not stdout) so it doesn't pollute JSON output. Uses a
+    plain-print rather than the rich console so it's reliably available in
+    background/non-TTY launchers.
+
+    The parameter is typed as ``list[Any]`` rather than
+    ``list[DiscoveredInstruction]`` to avoid a hard import at module load
+    time — the import is deferred to the call site in
+    :func:`run_workflow_async` so the discovery code path isn't loaded when
+    the flag is unused.
+    """
+    import sys
+
+    if not detailed:
+        print("[workspace-instructions] 0 files discovered from CWD.", file=sys.stderr)
+        return
+    print(
+        f"[workspace-instructions] {len(detailed)} file(s) loaded from CWD:",
+        file=sys.stderr,
+    )
+    for d in detailed:
+        # Use repr for the path so Windows paths render unambiguously.
+        print(f"  {d.path}", file=sys.stderr)
+        scope_part = f"  applyTo={d.scope!r}" if d.scope and d.scope != "**" else ""
+        print(
+            f"    source={d.source}  reason={d.reason}{scope_part}",
+            file=sys.stderr,
+        )
+
+
 async def run_workflow_async(
     workflow_path: Path,
     inputs: dict[str, Any],
@@ -1385,6 +1429,7 @@ async def run_workflow_async(
     metadata: dict[str, str] | None = None,
     workspace_instructions: bool = False,
     cli_instructions: list[str] | None = None,
+    print_loaded_instructions: bool = False,
 ) -> dict[str, Any]:
     """Execute a workflow asynchronously.
 
@@ -1401,6 +1446,9 @@ async def run_workflow_async(
         metadata: Optional CLI metadata to merge on top of YAML-declared metadata.
         workspace_instructions: If True, auto-discover workspace instruction files.
         cli_instructions: Optional list of instruction file paths from CLI.
+        print_loaded_instructions: If True, print the resolved instruction file
+            list (with scope and inclusion reason) to stderr before running.
+            No-op unless ``workspace_instructions`` is also True.
 
     Returns:
         The workflow output as a dictionary.
@@ -1513,6 +1561,16 @@ async def run_workflow_async(
                     f"Workspace instructions loaded ({len(instructions_preamble)} chars)",
                     style="cyan",
                 )
+
+            # --print-loaded-instructions: dump the resolved discovery list to
+            # stderr for debugging. Only meaningful when auto-discovery ran.
+            if print_loaded_instructions and workspace_instructions:
+                from conductor.config.instructions import (
+                    discover_workspace_instructions_detailed,
+                )
+
+                detailed = discover_workspace_instructions_detailed(Path.cwd())
+                _print_loaded_instructions(detailed)
 
         # Convert MCP servers from workflow config to SDK format
         mcp_servers = await _build_mcp_servers(config)
--- a/src/conductor/config/instructions.py
+++ b/src/conductor/config/instructions.py
@@ -531,7 +531,11 @@ def discover_workspace_instructions_detailed(start_dir: Path) -> list[Discovered
     """
     start_resolved = start_dir.resolve()
     git_root = _find_git_root(start_dir)
-    stop_at = git_root if git_root is not None else start_resolved
+    # Resolve git_root so equality against the (already-resolved) ``current``
+    # walk pointer holds even when the git root is reached via a symlinked
+    # path. Without this, ``current == stop_at`` can be False at the git
+    # root and the walk overshoots into the parent filesystem.
+    stop_at = git_root.resolve() if git_root is not None else start_resolved
 
     result: list[DiscoveredInstruction] = []
     for convention in CONVENTIONS:
@@ -612,7 +616,15 @@ def _discovery_reason(convention: ConventionDirectory, scope: str | None) -> str
         # Convention with no scope concept (or include_file-only); the file
         # passed the eligibility gate and that's all there is to say.
         return "always-on"
-    if scope == ALWAYS_ON_SCOPE or scope is None:
+    # Invariant: when ``extract_scope`` is set, ``_apply_convention_filters``
+    # returns ``_REJECT`` rather than a ``None`` scope, so a discovered file
+    # always carries a concrete scope string here. Pin this with an assert
+    # so a future regression surfaces loudly instead of being silently
+    # relabeled as "always-on".
+    assert scope is not None, (
+        "extract_scope set but scope is None — _apply_convention_filters invariant violated"
+    )
+    if scope == ALWAYS_ON_SCOPE:
         return "always-on"
     return "scope-overlap"
 
--- a/src/conductor/config/instructions.py
+++ b/src/conductor/config/instructions.py
@@ -27,6 +27,7 @@ import os
 import re
 from collections.abc import Callable, Iterator
 from dataclasses import dataclass
+from enum import Enum, StrEnum, auto
 from pathlib import Path
 
 from ruamel.yaml import YAML
@@ -121,6 +122,19 @@ Convention = ConventionFile | ConventionDirectory
 ALWAYS_ON_SCOPE = "**"
 
 
+class DiscoveryReason(StrEnum):
+    """Stable, machine-readable reason a file was auto-discovered.
+
+    A ``StrEnum`` (mirroring ``RegistryType`` in ``registry/config.py``) so the
+    f-string output and the string equality checks keep working unchanged,
+    while a typo like ``"scope_overlap"`` can no longer construct a record.
+    """
+
+    FILE_CONVENTION = "file-convention"
+    ALWAYS_ON = "always-on"
+    SCOPE_OVERLAP = "scope-overlap"
+
+
 @dataclass(frozen=True)
 class DiscoveredInstruction:
     """A workspace instruction file discovered by ``--workspace-instructions``,
@@ -139,14 +153,14 @@ class DiscoveredInstruction:
     # conventions with no scope concept (single-file conventions; directory
     # conventions without ``extract_scope``).
     scope: str | None
-    # Stable, machine-readable reason for inclusion. One of:
+    # Stable, machine-readable reason for inclusion. See :class:`DiscoveryReason`:
     #
-    # * ``"file-convention"`` — single-file convention, no scope concept.
-    # * ``"always-on"`` — scoped convention but file's scope is
+    # * ``FILE_CONVENTION`` — single-file convention, no scope concept.
+    # * ``ALWAYS_ON`` — scoped convention but file's scope is
     #   :data:`ALWAYS_ON_SCOPE` (or the convention has no ``extract_scope``).
-    # * ``"scope-overlap"`` — scoped convention; file's scope glob overlapped
+    # * ``SCOPE_OVERLAP`` — scoped convention; file's scope glob overlapped
     #   with the user's CWD subtree.
-    reason: str
+    reason: DiscoveryReason
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


franklixuefei:
@jrob5756 kindly requesting your review on this, thanks Jason!

franklixuefei:
@jrob5756 kindly requesting your review on this, thanks Jason!

franklixuefei:
@jrob5756 kindly requesting your review on this, thanks Jason!

jrob5756:
> @jrob5756 kindly requesting your review on this, thanks Jason!

Sorry for the delay here, @franklixuefei . Got really busy on a couple unrelated projects. Will try to wrap this up this week.

src/conductor/config/instructions.py
```diff
@@ -142,28 +210,129 @@ def _parse_frontmatter(path: Path) -> dict[str, object] | None:
...
+      (opt-out) rather than silently joining; that mirrors the convention's
+      stated shape.
     """
     fm = _parse_frontmatter(path)
     if fm is None:
-        return False
-    return fm.get("applyTo") == "**"
+        return None
+    apply_to = fm.get("applyTo")
+    if not isinstance(apply_to, str):
```
jrob5756, line 260: A list-valued `applyTo` (someone writing it as a YAML sequence) hits this `return None` and the file just disappears from discovery. Same for an empty or separator-only value once it reaches `_scope_overlaps`. That was harmless before this PR, since scoped files never loaded anyway, but now a real authoring slip silently drops instructions the user expected. The malformed-YAML branch in `_parse_frontmatter` does log a warning, but that channel goes nowhere in a normal run (no logging handlers are configured).

Could you split "applyTo absent" from "applyTo present but unusable" and surface the latter through `verbose_log` or a `WorkflowEvent`? That keeps the module's own "over-include rather than silently skip" bias honest on the error paths too.

src/conductor/config/instructions.py
```diff
@@ -75,12 +106,49 @@ class ConventionDirectory:
...
+    # conventions without ``extract_scope``).
+    scope: str | None
+    # Stable, machine-readable reason for inclusion. One of:
+    #
+    # * ``"file-convention"`` — single-file convention, no scope concept.
+    # * ``"always-on"`` — scoped convention but file's scope is
+    #   :data:`ALWAYS_ON_SCOPE` (or the convention has no ``extract_scope``).
+    # * ``"scope-overlap"`` — scoped convention; file's scope glob overlapped
+    #   with the user's CWD subtree.
+    reason: str
```
jrob5756: `reason` is really a three-value closed set, but as a bare `str` a typo like `reason="scope_overlap"` constructs happily and slips past the `==` checks in the printer and the tests. The repo already leans on `StrEnum` for non-Pydantic closed sets (e.g. `RegistryType` in `registry/config.py`), and `StrEnum` keeps the f-string output and the string asserts working unchanged. Define a `class DiscoveryReason(StrEnum)` (FILE_CONVENTION / ALWAYS_ON / SCOPE_OVERLAP) alongside and type the field with it:

```suggestion
    reason: DiscoveryReason
```

src/conductor/config/instructions.py
```diff
@@ -259,50 +454,128 @@ def _walk_directory_convention(
...
                 continue
-            yield entry.name, file_path
+            yield entry.name, file_path, accepted_scope
+
+
+# Sentinel returned by :func:`_apply_convention_filters` to signal "this file
+# was rejected by one of the filters; skip it." Distinct from ``None``, which
+# is a legitimate ``scope`` value meaning "no scope concept; the convention
+# has no ``extract_scope`` callable."
+_REJECT = object()
```
jrob5756: Using `object()` here pushes the return type of `_apply_convention_filters` all the way out to `object`, which throws away the real `str | None | <reject>` shape. The `yield`s downstream then feed an `Iterator[tuple[str, Path, str | None]]` that a stricter checker would flag (ty only passes because it doesn't check yield element types yet). A one-member `Enum` sentinel gives you real `is`-narrowing, so after the `is _REJECT` guard the value narrows back to `str | None`:

```python
class _Reject(Enum):
    TOKEN = auto()
```

Not urgent, but it makes the filter chain actually type-check.

src/conductor/config/instructions.py
```diff
@@ -259,50 +454,128 @@ def _walk_directory_convention(
...
+            # project, which is what the convention's `applyTo` globs are
+            # naturally written against.
+            try:
+                cwd_rel = start_resolved.relative_to(current).as_posix()
+            except ValueError:
+                # current is somehow not an ancestor of start_resolved (e.g.,
+                # symlink shenanigans). Fall back to empty CWD which makes
+                # the overlap test permissive — consistent with our
+                # "over-include rather than silently skip" bias.
+                cwd_rel = ""
```
jrob5756, line 624: This `except ValueError` can't fire today: `current` only ever moves up via `.parent` from `start_resolved`, so it's always an ancestor and `relative_to` won't raise. You pinned the analogous can't-happen case in `_discovery_reason` with an `assert`, so it'd be consistent to at least log before falling back here. If a future change ever breaks that invariant, a quiet over-include with no trace is the hard kind of bug to chase. Keep the permissive fallback though, that part's the right call.

src/conductor/cli/run.py
```diff
@@ -1371,6 +1371,76 @@ async def _execute_with_stop_signal(
...
+    """
+    if not enabled or start_dir is None:
+        return
+    from conductor.config.instructions import discover_workspace_instructions_detailed
+
+    detailed = discover_workspace_instructions_detailed(start_dir)
+    _print_loaded_instructions(detailed)
+
+
+def _print_loaded_instructions(detailed: list[Any]) -> None:
```
jrob5756: `list[Any]` drops checking on the `.path` / `.source` / `.scope` / `.reason` reads below, and the stated reason for it doesn't actually hold: this file has `from __future__ import annotations`, so a `DiscoveredInstruction` annotation is never evaluated at import time. The lazy import inside `_emit_loaded_instructions_debug` is what keeps the discovery module out of the graph, and it stays put. Add the type to the existing `TYPE_CHECKING` block up top and tighten the signature:

```suggestion
def _print_loaded_instructions(detailed: list[DiscoveredInstruction]) -> None:
```

src/conductor/cli/run.py
```diff
@@ -1371,6 +1371,76 @@ async def _execute_with_stop_signal(
...
+    :func:`run_workflow_async` so the discovery code path isn't loaded when
+    the flag is unused.
+    """
+    import sys
+
+    if not detailed:
+        print("[workspace-instructions] 0 files discovered from CWD.", file=sys.stderr)
+        return
+    print(
+        f"[workspace-instructions] {len(detailed)} file(s) loaded from CWD:",
```
jrob5756: This says "loaded", but it's really the discovered set. `load_instruction_files` still skips files that are empty or unreadable, so a discovered-but-empty file shows up here while contributing nothing to the preamble. Since the flag exists to debug what does and doesn't get applied, "discovered" is the more honest word, and it matches the empty-state line just below.

```suggestion
        f"[workspace-instructions] {len(detailed)} file(s) discovered from CWD:",
```

src/conductor/cli/run.py
```diff
@@ -1371,6 +1371,76 @@ async def _execute_with_stop_signal(
...
+        file=sys.stderr,
+    )
+    for d in detailed:
+        # Use repr for the path so Windows paths render unambiguously.
+        print(f"  {d.path}", file=sys.stderr)
```
jrob5756: The comment says repr, but the line below prints `{d.path}`, which is plain `str()`. The plain form actually reads better for paths, so I'd just drop the comment. The only `!r` in this block is on `d.scope`, one line down.

```suggestion
        print(f"  {d.path}", file=sys.stderr)
```

src/conductor/cli/run.py
```diff
@@ -1371,6 +1371,76 @@ async def _execute_with_stop_signal(
...
+        scope_part = f"  applyTo={d.scope!r}" if d.scope and d.scope != "**" else ""
```
jrob5756: Tiny one: this compares against the literal `"**"`, but you added `ALWAYS_ON_SCOPE` precisely so the sentinel has one greppable home. Worth referencing the constant here instead (it would need adding to the import from `conductor.config.instructions`).

jrob5756 (commented):
Took a thorough the code changes. The core fix is right and well covered. I rebuilt discovery on a scratch repo and the overlap math holds up: root CWD loads every scoped file, subdirs narrow, and a nested .github/instructions resolves its applyTo against the right owner dir. Tests pass and ruff/ty are clean.

Nothing here blocks merge. The one theme worth a real look is silent drops: a malformed, list-valued, or empty applyTo quietly removes a file from discovery, and the logger.warning fallback is effectively invisible because conductor installs no logging handlers, so those records only reach Python's bare lastResort stderr and never the dashboard or the --web-bg logs. Everything else is type/comment/test polish. Details inline.

src/conductor/config/instructions.py
franklixuefei: Done — added `class DiscoveryReason(StrEnum)` (FILE_CONVENTION / ALWAYS_ON / SCOPE_OVERLAP) and typed the field with it. `_discovery_reason` now returns the enum, and the one literal construction site uses `DiscoveryReason.FILE_CONVENTION`. As you noted, `StrEnum` keeps the f-string output and the `==` checks working unchanged, while a typo'd reason becomes a type error instead of a silent miss. (`69e810b`)

src/conductor/config/instructions.py
franklixuefei: Done — replaced `_REJECT = object()` with a one-member `class _Reject(Enum): TOKEN = auto()` and narrowed `_apply_convention_filters`'s return to `str | None | _Reject`. After the `is _REJECT` guard the value narrows back to `str | None`, so the downstream `yield` chain type-checks honestly instead of leaning on ty not checking yield element types yet. Thanks for the precise framing. (`69e810b`)

src/conductor/config/instructions.py
franklixuefei: Agreed — added a `logger.warning` before the permissive fallback so a future invariant break leaves a trace instead of a quiet over-include. Kept the fallback itself, as you suggested. (`69e810b`)

src/conductor/config/instructions.py
franklixuefei: Thanks — this is the most important one, and you're right that "absent" and "present-but-unusable" deserve different treatment. I split them in `_extract_apply_to` (`69e810b`): a genuinely absent `applyTo` opts out silently (per GitHub's "not applied automatically" spec), while the authoring-slip cases — a non-string (e.g. a YAML list) and an empty/separator-only value — now warn before opting out, each with its own message. Added three tests for the three paths.

On the channel, I kept these on `logger.warning` rather than `verbose_log`, and I want to be transparent about why because it hinges on a detail worth double-checking. Conductor configures no logging handlers, so I expected the "goes nowhere" behavior you flagged — but WARNING-level records actually reach stderr via Python's `logging.lastResort` (the default `_StderrHandler`, level WARNING). I checked it after importing the full CLI + deps: the warning prints to stderr, and both the module logger and the root logger have no handlers, with `logging.lastResort` still the default `_StderrHandler`. So the unusable-`applyTo` case isn't silent in a normal run — it's visible by default.

The tradeoff that kept me there: `verbose_log` is `--verbose`-gated (`is_verbose()`), so routing the warning through it would actually *reduce* visibility — the slip would only surface if the user opted into verbose, which is the opposite of what we want for a silent drop. `logger.warning` is also consistent with the existing malformed-frontmatter warning in `_parse_frontmatter` (same authoring-error class, same channel).

That said, it's your call on house style. If you'd prefer `verbose_log` for consistency with the discovery debug output, or a `WorkflowEvent` so it can also reach the dashboard, I'm glad to switch — the `WorkflowEvent` route is a little more wiring since discovery runs before the emitter is available. Let me know which you'd like.

src/conductor/cli/run.py
franklixuefei: You're right on both counts — `from __future__ import annotations` means the annotation never evaluates at import, and the lazy import inside `_emit_loaded_instructions_debug` is what actually keeps the discovery module out of the graph. Added `DiscoveredInstruction` to the `TYPE_CHECKING` block and tightened the signature to `list[DiscoveredInstruction]`; the `.path` / `.source` / `.scope` / `.reason` reads are now checked. (`69e810b`)

src/conductor/cli/run.py
franklixuefei: Done — switched to "discovered" here. Agreed it's the more honest word given `load_instruction_files` can still skip an empty/unreadable file, and it matches the empty-state line just below. (`69e810b`)

src/conductor/cli/run.py
franklixuefei: Done — dropped the comment and kept the plain `{d.path}`. The only `!r` left is on `d.scope` one line down, as you noted. (`69e810b`)

src/conductor/cli/run.py
franklixuefei: Done — referencing `ALWAYS_ON_SCOPE` here now (added it to the import from `conductor.config.instructions`) so the sentinel keeps its one greppable home. (`69e810b`)

franklixuefei:
> Took a thorough the code changes. The core fix is right and well covered. I rebuilt discovery on a scratch repo and the overlap math holds up: root CWD loads every scoped file, subdirs narrow, and a nested .github/instructions resolves its applyTo against the right owner dir. Tests pass and ruff/ty are clean.
> 
> Nothing here blocks merge. The one theme worth a real look is silent drops: a malformed, list-valued, or empty applyTo quietly removes a file from discovery, and the logger.warning fallback is effectively invisible because conductor installs no logging handlers, so those records only reach Python's bare lastResort stderr and never the dashboard or the --web-bg logs. Everything else is type/comment/test polish. Details inline.

@jrob5756 thanks for your review! Addressed most comment while leaving one comment open for your take. Thanks!

franklixuefei:
@jrob5756 Just a gentle nudge on re-reviewing this after the comments were resolved (except for one of them needing your call). Wanted this to make it to the 0.1.20 release train :-P Thanks!

src/conductor/config/instructions.py
jrob5756: Thanks for the thorough writeup, @franklixuefei — and for splitting the absent vs. present-but-unusable paths, which was the heart of my concern.

Confirmed your `logging.lastResort` point: with no handlers configured, WARNING records do reach stderr by default, so the unusable-`applyTo` case is no longer a silent drop (and it's captured in the `--web-bg` `.bg.stderr.log` too). Agreed that `verbose_log` would be backwards here — it'd hide the warning behind `--verbose`, the opposite of what we want — and `logger.warning` keeps it consistent with the existing `_parse_frontmatter` warning.

Let's go with `logger.warning`. The only surface it misses is the live dashboard / JSONL event stream, and since discovery runs before the emitter exists that's a broader buffer-and-replay change — a reasonable fast-follow if it ever proves necessary, not a blocker here. Resolving.

jrob5756 (approved):
All review threads addressed and resolved. Accepted `logger.warning` as the channel for the unusable-`applyTo` warnings (it reaches stderr by default; dashboard/JSONL surfacing can be a fast-follow). Core fix is solid and well-tested, CI green. LGTM 🚀

PR #239, opened 2026-06-01
fix(context): support nested explicit input projection

ldavidgomez:
## Draft follow-up to #237

This PR is intentionally separate from the `human_gate` `prompt_for` serialization fix. It addresses a generic explicit-context projection gap in `WorkflowContext._add_agent_input`.

## Bug

In `context: explicit`, input declarations support shallow output projection, for example:

```yaml
input:
  - agent.field
  - agent.output.field
```

But nested output paths are not projected correctly:

```yaml
input:
  - agent.foo.bar
  - agent.output.foo.bar
```

These paths either fall through or fail to expose the requested nested value to downstream templates.

## Fix

Update `WorkflowContext._add_agent_input` so explicit input declarations can traverse nested dictionary outputs and project only the requested nested leaf.

This applies generically to any dict-shaped agent output. It is not special-cased to `human_gate`.

## Supported forms

```yaml
input:
  - agent.foo.bar
  - agent.output.foo.bar
```

Both expose the nested value under:

```jinja
{{ agent.output.foo.bar }}
```

## Tests added

| Test | What it covers |
|---|---|
| `test_explicit_mode_nested_field_shorthand` | `agent.foo.bar` exposes `output.foo.bar`; siblings excluded |
| `test_explicit_mode_nested_field_output_prefix` | `agent.output.foo.bar` exposes the same path |
| `test_explicit_mode_nested_field_multiple_declarations` | Two leaves under the same parent do not overwrite each other |
| `test_explicit_mode_nested_field_missing_required_raises` | Missing required nested path raises `KeyError` |
| `test_explicit_mode_nested_field_optional_missing_skipped` | Optional missing nested path is skipped |
| `test_explicit_mode_nested_intermediate_not_dict_raises` | Mid-path scalar raises `KeyError` with a clear message |

## Tests run

```bash
uv run pytest tests/test_engine/test_context.py -k "nested" -v
# 6 passed

uv run pytest tests/test_engine/test_context.py -v
# 90 passed
```

## Scope note (validator/runtime alignment)

Nested explicit projection is now supported for regular dict-shaped agent outputs:
- `agent.output.field.subfield`
- `agent.field.subfield` (shorthand)

Parallel-group projection depth is intentionally unchanged and out of scope:
- supported: `group.outputs.agent.field`
- not supported: `group.outputs.agent.field.subfield`

### DIFF
--- a/src/conductor/engine/context.py
+++ b/src/conductor/engine/context.py
@@ -276,8 +276,8 @@ class WorkflowContext:
         Input reference formats:
         - workflow.input.param_name - References a workflow input
         - agent_name.output - References an agent's entire output
-        - agent_name.output.field - References a specific output field
-        - agent_name.field - Shorthand for agent_name.output.field (deprecated but supported)
+        - agent_name.output.field[.subfield...] - Nested output projection
+        - agent_name.field[.subfield...] - Shorthand nested projection
         - parallel_group.outputs - References all parallel group outputs
         - parallel_group.outputs.agent_name - References a specific parallel agent's output
         - parallel_group.outputs.agent_name.field - Specific field from parallel agent
@@ -347,6 +347,13 @@ class WorkflowContext:
     ) -> None:
         """Add a regular agent output reference to context.
 
+        Supported behaviors:
+        - ``agent_name`` or ``agent_name.output`` copies the full output.
+        - ``agent_name.output.field[.subfield...]`` projects nested fields.
+        - ``agent_name.field[.subfield...]`` projects the same nested fields via shorthand.
+        - Optional refs (``?``) skip missing leaves/intermediate paths.
+        - Projected leaves are deep-copied to avoid mutation aliasing.
+
         Args:
             ctx: The context dictionary to update.
             agent_name: The name of the agent.
@@ -359,6 +366,7 @@ class WorkflowContext:
         agent_output = self.agent_outputs[agent_name]
         is_dict_output = isinstance(agent_output, dict)
 
+        # Seed ctx[agent_name] so optional-skip returns still leave a stub entry.
         # Initialise ``output`` to an empty dict for dict-shaped outputs (so
         # subsequent field-writes have somewhere to land) or to ``None`` for
         # scalar/list outputs (which are assigned whole below).
@@ -389,19 +397,19 @@ class WorkflowContext:
             # Traverse agent_output along path
             value: Any = agent_output
             for i, key in enumerate(path):
-                if not isinstance(value, dict) or key not in value:
-                    if not is_optional:
-                        if not isinstance(value, dict):
-                            raise KeyError(
-                                f"Cannot access field '{key}' on agent '{agent_name}': "
-                                f"intermediate value at '{'.'.join(path[:i])}' is a "
-                                f"{type(value).__name__}, not a dict"
-                            )
-                        raise KeyError(
-                            f"Missing output field '{'.'.join(path)}' from agent '{agent_name}'"
-                        )
+                if isinstance(value, dict) and key in value:
+                    value = value[key]
+                    continue
+                if is_optional:
                     return
-                value = value[key]
+                intermediate_path = ".".join(path[:i]) if i > 0 else agent_name
+                if not isinstance(value, dict):
+                    raise KeyError(
+                        f"Cannot access field '{key}' on agent '{agent_name}': "
+                        f"intermediate value at '{intermediate_path}' is a "
+                        f"{type(value).__name__}, not a dict"
+                    )
+                raise KeyError(f"Missing output field '{'.'.join(path)}' from agent '{agent_name}'")
 
             # Write the resolved leaf value into ctx at the nested output path,
             # creating intermediate dicts as needed. deepcopy matches the
--- a/src/conductor/engine/context.py
+++ b/src/conductor/engine/context.py
@@ -367,48 +367,53 @@ class WorkflowContext:
         elif "output" not in ctx[agent_name]:
             ctx[agent_name]["output"] = {} if is_dict_output else None
 
-        if not remaining_parts:
-            # Just agent_name - copy entire output
+        if not remaining_parts or remaining_parts == ["output"]:
+            # agent_name or agent_name.output — copy entire output
             ctx[agent_name]["output"] = (
                 copy.deepcopy(agent_output) if is_dict_output else (copy.copy(agent_output))
             )
-        elif len(remaining_parts) == 1 and remaining_parts[0] == "output":
-            # agent_name.output - copy entire output
-            ctx[agent_name]["output"] = (
-                copy.deepcopy(agent_output) if is_dict_output else (copy.copy(agent_output))
-            )
-        elif len(remaining_parts) >= 2 and remaining_parts[0] == "output":
-            # agent_name.output.field - copy specific field. Only meaningful
-            # when the stored output is a dict; otherwise the field access
-            # is undefined and (when required) must raise.
-            field_name = remaining_parts[1]
-            if is_dict_output and field_name in agent_output:
-                # Ensure we have a dict to write the field into (the initial
-                # output slot above seeded ``None`` for non-dict outputs).
-                if not isinstance(ctx[agent_name]["output"], dict):
-                    ctx[agent_name]["output"] = {}
-                ctx[agent_name]["output"][field_name] = agent_output[field_name]
-            elif not is_optional:
-                if not is_dict_output:
+        else:
+            # Resolve field path:
+            #   agent_name.output.field[.subfield...] → path = [field, subfield, ...]
+            #   agent_name.field[.subfield...]        → path = [field, subfield, ...]  (shorthand)
+            path = remaining_parts[1:] if remaining_parts[0] == "output" else remaining_parts
+
+            if not is_dict_output:
+                if not is_optional:
                     raise KeyError(
-                        f"Cannot access field '{field_name}' on agent '{agent_name}': "
+                        f"Cannot access field '{path[0]}' on agent '{agent_name}': "
                         f"its output is a {type(agent_output).__name__}, not a dict"
                     )
-                raise KeyError(f"Missing output field '{field_name}' from agent '{agent_name}'")
-        elif len(remaining_parts) == 1 and remaining_parts[0] != "output":
-            # Shorthand format: agent_name.field -> agent_name.output.field
-            field_name = remaining_parts[0]
-            if is_dict_output and field_name in agent_output:
-                if not isinstance(ctx[agent_name]["output"], dict):
-                    ctx[agent_name]["output"] = {}
-                ctx[agent_name]["output"][field_name] = agent_output[field_name]
-            elif not is_optional:
-                if not is_dict_output:
-                    raise KeyError(
-                        f"Cannot access field '{field_name}' on agent '{agent_name}': "
-                        f"its output is a {type(agent_output).__name__}, not a dict"
-                    )
-                raise KeyError(f"Missing output field '{field_name}' from agent '{agent_name}'")
+                return
+
+            # Traverse agent_output along path
+            value: Any = agent_output
+            for i, key in enumerate(path):
+                if not isinstance(value, dict) or key not in value:
+                    if not is_optional:
+                        if not isinstance(value, dict):
+                            raise KeyError(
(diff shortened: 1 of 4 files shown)


src/conductor/engine/context.py
```diff
@@ -367,48 +367,53 @@ def _add_agent_input(
         elif "output" not in ctx[agent_name]:
             ctx[agent_name]["output"] = {} if is_dict_output else None
```
jrob5756, line 376: After the rewrite, this init block exists almost entirely so that the two early-`return` paths (optional-skipped non-dict on line 387, and optional-missing during traversal on line 403) leave a `ctx[agent_name] = {"output": ...}` stub behind. Every write path below overwrites or re-seeds `ctx[agent_name]["output"]` independently — and lines 409-410 re-check the shape before the nested write, which looks like duplication unless you know about this contract.

A one-line "why" comment makes the contract explicit and makes 409-410 read as "ensure dict shape before nested write" rather than "unnecessary recheck":

```python
# Seed ctx[agent_name] so optional-skip returns still leave a stub entry.
# All write paths below overwrite or re-seed ctx[agent_name]["output"] as needed.
```

src/conductor/engine/context.py
```diff
@@ -367,48 +367,53 @@ def _add_agent_input(
...
+                        if not isinstance(value, dict):
+                            raise KeyError(
+                                f"Cannot access field '{key}' on agent '{agent_name}': "
+                                f"intermediate value at '{'.'.join(path[:i])}' is a "
+                                f"{type(value).__name__}, not a dict"
+                            )
+                        raise KeyError(
+                            f"Missing output field '{'.'.join(path)}' from agent '{agent_name}'"
+                        )
+                    return
```
jrob5756, line 404: Two small things on this traversal loop:

**1. Flatten the stacked negations.** The reader has to track `not isinstance ... or not in` → `not is_optional` → `not isinstance` in a 3-level pyramid. Inverting the optional guard first removes a level and gives each error its own clear trigger:

```python
for i, key in enumerate(path):
    if isinstance(value, dict) and key in value:
        value = value[key]
        continue
    if is_optional:
        return
    if not isinstance(value, dict):
        raise KeyError(
            f"Cannot access field '{key}' on agent '{agent_name}': "
            f"intermediate value at '{'.'.join(path[:i])}' is a "
            f"{type(value).__name__}, not a dict"
        )
    raise KeyError(
        f"Missing output field '{'.'.join(path)}' from agent '{agent_name}'"
    )
```

Same behavior, happy path first.

**2. The inner `not isinstance(value, dict)` arm can never fire at `i == 0`** — the outer `not is_dict_output` guard at line 381 has already returned/raised in that case. So `path[:i]` is guaranteed non-empty when this branch fires. If someone later relaxes the outer guard, the error message becomes `intermediate value at '' is a ...`. Either assert `i >= 1` with a comment, or special-case `i == 0` to produce a sensible message. Not blocking — just future-proofing.

tests/test_engine/test_context.py
```diff
@@ -352,6 +352,90 @@ def test_explicit_mode_human_gate_unchanged(self) -> None:
 
         assert agent_ctx["workflow"]["input"] == {}
 
+    def test_explicit_mode_nested_field_shorthand(self) -> None:
+        """Shorthand agent.foo.bar exposes output.foo.bar; sibling keys are excluded."""
+        ctx = WorkflowContext()
+        ctx.store("gate", {"selected": "go", "additional_input": {"answer": "README.md", "other": "x"}})
```
jrob5756: **E501 — line is 104 chars (limit 100).** `make lint` will fail. Either run `uv run ruff format` (which will reflow this automatically) or split the literal:

```python
ctx.store(
    "gate",
    {"selected": "go", "additional_input": {"answer": "README.md", "other": "x"}},
)
```

tests/test_engine/test_context.py
```diff
@@ -352,6 +352,90 @@ def test_explicit_mode_human_gate_unchanged(self) -> None:
...
+        ctx.store("gate", {"additional_input": {}})
+
+        agent_ctx = ctx.build_for_agent(
+            "next",
+            ["gate.additional_input.answer?"],
+            mode="explicit",
+        )
+
+        # No error; the key is simply absent
+        assert "answer" not in agent_ctx.get("gate", {}).get("output", {}).get("additional_input", {})
```
jrob5756: **E501 — line is 102 chars** (CI will fail), and the assertion itself is too forgiving.

The current `.get("gate", {}).get("output", {}).get("additional_input", {})` chain passes under all four of these distinct ctx shapes:
- `agent_ctx` has no `"gate"` key (init code didn't run)
- `agent_ctx["gate"]` has no `"output"` key (seed missing)
- `agent_ctx["gate"]["output"] == {}` (the actual current behavior — traversal returns before any write)
- `agent_ctx["gate"]["output"]["additional_input"] == {}` (a partial intermediate write happened)

Since the current implementation early-returns *before* any write, the precise expected state is `{"gate": {"output": {}}}` — pin that and you catch any regression where intermediate-dict creation drifts before the value-resolution loop:

```python
assert agent_ctx["gate"] == {"output": {}}
```

Fixes both the E501 and the masking assertion in one go.

tests/test_engine/test_context.py
```diff
@@ -352,6 +352,90 @@ def test_explicit_mode_human_gate_unchanged(self) -> None:
...
+        """Traversing through a non-dict intermediate value raises KeyError."""
+        ctx = WorkflowContext()
+        ctx.store("agent1", {"foo": "scalar_not_a_dict"})
+
+        with pytest.raises(KeyError, match="intermediate value"):
+            ctx.build_for_agent(
+                "agent2",
+                ["agent1.foo.bar"],
+                mode="explicit",
+            )
```
jrob5756, line 494: Three test gaps worth filling before merge — the first is the most important.

**1. Regression test for the silent semantic change to `agent.output.field.subfield` (high priority).** Pre-PR, this form took `field_name = remaining_parts[1]` and copied the *entire* `field` dict (siblings included). Post-PR, it projects only the `subfield` leaf. Real behavior change, no test pins it, no mention in the PR body:

```python
def test_explicit_mode_output_prefix_three_parts_projects_leaf_only(self) -> None:
    """Regression: a.output.foo.bar projects only `bar`, not the whole `foo` dict."""
    ctx = WorkflowContext()
    ctx.store("a", {"foo": {"bar": 1, "sibling": 2}})
    rendered = ctx.build_for_agent("c", ["a.output.foo.bar"], mode="explicit")
    assert rendered["a"]["output"] == {"foo": {"bar": 1}}
```

**2. Mixed parent + child declarations (both orders).** The write loop's create-if-absent merge is the subtlest part of the rewrite; neither order is currently tested:

```python
# [a.foo, a.foo.bar] — parent then child, full foo expected
# [a.foo.bar, a.foo] — child then parent, leaf write replaced by full foo
```

**3. Deep nesting (3+ levels) + mid-path error message.** Every new test uses exactly 2 levels (`a.parent.leaf`). The generic loop is never exercised at depth ≥3, and the `path[:i]` index in the error message is never asserted:

```python
def test_explicit_mode_deeply_nested_missing_mid_path_error_message(self) -> None:
    ctx = WorkflowContext()
    ctx.store("a", {"l1": {"l2": {}}})
    with pytest.raises(KeyError, match=r"a\.l1\.l2\.l3\.leaf"):
        ctx.build_for_agent("c", ["a.l1.l2.l3.leaf"], mode="explicit")
```

**Lower-priority hardening** (fine as follow-ups): optional miss at *intermediate* (not just leaf) position; an aliasing test confirming the documented deepcopy invariant; a non-dict leaf (list/scalar) at the end of a nested path; tightening `match="intermediate value"` to include the path prefix and type name so the diagnostic detail can't silently degrade.

jrob5756 (commented):
Thanks for tackling this — the runtime rewrite is genuinely cleaner than the four-way `elif` chain it replaces, sibling-exclusion is well-tested in both shorthand and `output`-prefix forms, and no dead code is introduced (every new branch is reachable from real workflow shapes). The 6 new tests still pass alongside the existing 90.

I ran a multi-agent review (code-reviewer, pr-test-analyzer, comment-analyzer, dead-code-finder, code-simplifier). Aggregated findings below; inline comments cover the items that map to specific lines.

## 🔴 Critical — validator rejects every form this PR enables

The validator's `INPUT_REF_PATTERN` at `src/conductor/config/validator.py:89-95` accepts only `agent.output(.field)?` (a single field). I verified empirically against this branch:

```
$ uv run conductor validate workflow-using-nested-input.yaml
ConfigurationError: Agent 'consumer' has invalid input reference
'gate.output.additional_input.answer'. Expected format: 'agent_name.output',
'agent_name.output.field', ...
```

The runtime now accepts nested forms; validation rejects them. Anyone following the docs' recommendation to `conductor validate` before `conductor run`, or CI's `make validate-examples`, cannot use this feature. There's a pre-existing inconsistency for the bare `a.field` shorthand, but this PR widens the gap by adding more runtime-supported forms.

**Required follow-up in this PR:**
1. Update `INPUT_REF_PATTERN` to accept `agent.output.field(.subfield)*` and `agent.field(.subfield)*`.
2. Update the error message text on `validator.py:323-327` to match.
3. Add validator tests for both forms in `tests/test_config/test_validator.py`.
4. **Second-order concern:** the template-reference advisory machinery at `validator.py:858-862, 1316-1319, 1375-1402` only tracks the *first* field after `.output`. Decide whether nested field-precision warnings should track full paths or deliberately degrade — either is defensible, but pick one explicitly.

## 🟠 Important — docstrings outside the diff

(Can't comment inline because they're outside the changed hunks, but they should be updated in this PR.)

**`_add_explicit_input` docstring at `src/conductor/engine/context.py:270-292`** — its "Input reference formats" bullet list still enumerates only single-level forms. A maintainer reading it will conclude nested refs are unsupported. It also annotates the `agent.field` shorthand as `(deprecated but supported)` — yet this PR doubles down on it by adding the *nested* shorthand as a tested first-class feature. Drop the deprecation note or commit to a real deprecation (`DeprecationWarning` + issue link).

**`_add_agent_input` docstring at `src/conductor/engine/context.py:348-358`** is too generic for what the function now does. After the rewrite it handles four `remaining_parts` shapes (`[]`, `["output"]`, `["output", f1, f2, ...]`, `[f1, f2, ...]`), traverses arbitrarily deep, deep-copies leaves to prevent mutation aliasing, and raises two distinct `KeyError` messages (`intermediate value at ...` vs `Missing output field ...`). Add a `Behavior:` block listing those — tests like `test_explicit_mode_nested_intermediate_not_dict_raises` are currently matching on undocumented implementation detail.

Suggested additions to `_add_explicit_input`:

```
- agent_name.output.field[.subfield...] - References a nested output field
- agent_name.field[.subfield...]        - Shorthand for the above
```

## Out of scope (worth a follow-up issue)

`_add_parallel_group_input` (lines 418+) still caps at `group.outputs.agent.field`. Mention in the PR body so it's clearly a deliberate scope decision.

src/conductor/engine/context.py
ldavidgomez: Implemented in commit 1770382. I added the why-comment above the seed block to make the optional-skip stub contract explicit and to clarify why the later output-shape recheck is still intentional.

src/conductor/engine/context.py
ldavidgomez: Implemented in commit 1770382. I flattened the traversal loop to a happy-path-first structure and added explicit intermediate-path handling so error messages remain sensible and non-empty, including future-proof behavior for early-path failures.

tests/test_engine/test_context.py
ldavidgomez: Implemented in commit 1770382. I split that store literal into multiline form so it passes line-length lint (ruff E501).

tests/test_engine/test_context.py
ldavidgomez: Implemented in commit 1770382. I replaced the permissive chained-get assertion with the precise expected shape assertion (gate output stub) and fixed line length at the same time.

tests/test_engine/test_context.py
ldavidgomez: Implemented in commit 1770382. I added the requested coverage for nested explicit projection semantics, parent/child declaration ordering, deeper nesting, and mid-path error behavior. I also updated validator/runtime alignment tests and kept deeper parallel-group projection explicitly out of scope.

jrob5756 (approved):

PR #252, opened 2026-06-16
feat(context): add per-agent context_tier (long_context / 1M) for Cop…

### LINKED ISSUE #251: Copilot provider: forward a per-agent context_tier (long_context / 1M) to create_session
## Summary

The Copilot SDK exposes a `context_tier` parameter on `create_session` (`ContextTier = Literal["default", "long_context"]`) to select a model's long-context (e.g. 1M-token) window. **Conductor has no way to set it** — neither in the workflow schema (`AgentDef`) nor in the Copilot provider, which only forwards the bare model name. As a result, workflows cannot request the long-context tier for any agent.

## Why this matters now

The long-context window for newer Claude Opus models **moved from a model-name suffix to this separate session parameter**. Previously you could pin a 1M-context model by naming it (e.g. an internal `claude-opus-4.7-1m...` alias). Those `-1m` suffixed aliases no longer resolve:

```
ProviderError: SDK call failed after 3 attempts: Copilot SDK call failed:
JSON-RPC Error -32603: Request session.create failed with message:
Model "claude-opus-4.7-1m-internal" is not available.
```

So a workflow that previously got 1M context via the model name now silently has **no way to get it at all** through conductor — the model resolves at the default (~200K) tier with no knob to change it. Heavy-reasoning agents that ingest large evidence (multi-MB logs, many candidate source files) are the ones that need the 1M window most.

## Repro

1. Pin any agent to `model: claude-opus-4.8` (resolves fine).
2. There is no schema field to also request `context_tier: long_context`.
3. The agent runs at the default context tier; large prompts truncate.

Probing the `-1m` model-name route fails (`NOT AVAILABLE`) for every variant tried (`claude-opus-4.8-1m`, `-1m-internal`, `-1m-preview`).

## Proposed fix

Add an optional `context_tier` to `AgentDef` and forward it from the Copilot provider into `create_session`.

**`conductor/config/schema.py`** — new field on `AgentDef` (right after `model`):

```python
context_tier: Literal["default", "long_context"] | None = None
"""Context-window tier for models that support it (Copilot provider).
Set `context_tier: long_context` to pin a heavy-reasoning agent to the
long-context tier. None sends no value. Only the Copilot provider forwards
this today (maps to the SDK's create_session `context_tier` param)."""
```

**`conductor/providers/copilot.py`** — forward it where `session_kwargs` is built:

```python
agent_context_tier = getattr(agent, "context_tier", None)
if agent_context_tier:
    session_kwargs["context_tier"] = agent_context_tier
```

That's the whole change — the SDK already accepts the kwarg (`create_session(..., context_tier=...)`), conductor just never sets it. Verified working against v0.1.18 with a local patch: `model: claude-opus-4.8` + `context_tier: long_context` validates and runs to completion.

Happy to send a PR if this shape is agreeable. Should `context_tier` also be settable as a `runtime.default_context_tier` (workflow-wide default, per-agent override) mirroring how `reasoning_effort` / `default_reasoning_effort` work? Open to either.

## Environment

- conductor-cli v0.1.18 (latest), installed via `uv tool`
- Copilot SDK: `create_session(model, ..., context_tier: ContextTier | None)` present in `copilot/client.py`

brrusino:
Closes #251.

## Why

The Copilot SDK selects a model's long-context (e.g. 1M-token) window via a
`context_tier` parameter on `create_session`, but conductor had no way to set
it. The long-context window moved from a model-name suffix to this session
parameter, and the old `-1m` model aliases no longer resolve, so workflows that
need the 1M window for heavy-evidence agents silently truncated at the default
(~200K) tier with no knob to change it.

## Approach

Adds a unified `context_tier` knob (`default` | `long_context`), built to mirror
the existing two-level `reasoning` design so the API surface stays consistent:

- **`providers/context_tier.py`** (new): `ContextTier` literal plus a
  `resolve_context_tier(agent, runtime_default)` helper (per-agent value wins,
  else runtime default, else send nothing).
- **`config/schema.py`**: `AgentDef.context_tier` (flat field, sibling to
  `model`) and `RuntimeConfig.default_context_tier` (workflow-wide default).
  Rejected on `script` / `human_gate` / `workflow` agents, which never call a
  model.
- **`providers/copilot.py`**: forwards the resolved tier into both agent and
  dialog `create_session` calls.
- **`providers/factory.py`**: threads `default_context_tier` through the same
  path as `default_reasoning_effort`.
- **`engine/workflow.py`**: surfaces the effective per-agent `context_tier` in
  system metadata for the dashboard and JSONL logs.

## Notes for reviewers

- `context_tier` and `reasoning.effort` are independent and compose: they map to
  two separate `create_session` kwargs, so an agent can carry both.
- This is a Copilot-only capability. Other providers have no equivalent
  parameter and ignore the value (scoped this way in the issue discussion).
- Unlike reasoning effort, there is no advertised capability list to validate
  against, so there is no static validation; the SDK accepts or rejects the
  value at session creation.

## Tests and docs

- Schema tests: field acceptance/rejection, defaults, composition with
  reasoning, and per-agent-type guards.
- Copilot provider tests: per-agent forwarding, runtime default, override
  precedence, absent-key, and reasoning composition.
- Factory test: `default_context_tier` reaches the provider.
- Added `examples/context-tier.yaml` (validates clean) and updated
  `docs/configuration.md`, `docs/workflow-syntax.md`, `examples/README.md`, and
  `CHANGELOG.md`.

### DIFF
--- a/src/conductor/config/schema.py
+++ b/src/conductor/config/schema.py
@@ -19,6 +19,7 @@ from pydantic import (
 )
 
 from conductor.duration import parse_duration
+from conductor.providers.context_tier import ContextTier
 from conductor.providers.reasoning import ReasoningEffort
 
 # Maximum allowed wait-step duration (24 hours). Anything longer almost
@@ -533,6 +534,26 @@ class AgentDef(BaseModel):
     Supports Jinja2 templates: {{ workflow.input.model_name }}
     """
 
+    context_tier: ContextTier | None = None
+    """Context-window tier for models that support it (Copilot provider only).
+
+    Set ``context_tier: long_context`` to pin a heavy-reasoning agent to the
+    model's long-context (e.g. 1M-token) window. ``default`` selects the
+    standard tier; ``None`` sends no value (provider default).
+
+    Falls back to ``runtime.default_context_tier`` when unset. Composes
+    independently with ``reasoning`` — an agent may set both.
+
+    Only the Copilot provider forwards this today (maps to the SDK's
+    ``create_session`` ``context_tier`` param). Other providers ignore it.
+
+    Only applies to provider-backed agents (type='agent' or None).
+
+    Example YAML::
+
+        context_tier: long_context
+    """
+
     input: list[str] = Field(default_factory=list)
     """Context dependencies. Format: 'agent_name.output' or 'workflow.input.param'.
     Suffix with '?' for optional dependencies."""
@@ -892,6 +913,8 @@ class AgentDef(BaseModel):
                 raise ValueError("human_gate agents cannot have 'max_depth'")
             if self.reasoning is not None:
                 raise ValueError("human_gate agents cannot have 'reasoning'")
+            if self.context_tier is not None:
+                raise ValueError("human_gate agents cannot have 'context_tier'")
             if self.timeout_seconds is not None:
                 raise ValueError("human_gate agents cannot have 'timeout_seconds'")
             if self.value is not None:
@@ -931,6 +954,8 @@ class AgentDef(BaseModel):
                 raise ValueError("script agents cannot have 'max_depth'")
             if self.reasoning is not None:
                 raise ValueError("script agents cannot have 'reasoning'")
+            if self.context_tier is not None:
+                raise ValueError("script agents cannot have 'context_tier'")
             if self.timeout_seconds is not None:
                 raise ValueError(
                     "script agents cannot have 'timeout_seconds' "
@@ -1183,6 +1208,8 @@ class AgentDef(BaseModel):
                 )
         if self.type == "workflow" and self.reasoning is not None:
             raise ValueError("workflow agents cannot have 'reasoning'")
+        if self.type == "workflow" and self.context_tier is not None:
+            raise ValueError("workflow agents cannot have 'context_tier'")
 
         # Wait-only fields are forbidden on every other type. ``reason`` is
         # shared with ``type: terminate`` (which has its own required-non-
@@ -1563,6 +1590,17 @@ class RuntimeConfig(BaseModel):
     match the supported prefix list; Copilot consults the SDK's advertised
     ``supported_reasoning_efforts`` (when available) and otherwise allows
     the request through to the SDK.
+    """
+
+    default_context_tier: ContextTier | None = None
+    """Workflow-wide default context-window tier (Copilot provider only).
+
+    Each agent may override with its own ``context_tier``. ``long_context``
+    selects a model's long-context (e.g. 1M-token) window; ``default`` selects
+    the standard tier; ``None`` sends no value.
+
+    Only the Copilot provider forwards this (maps to the SDK's
+    ``create_session`` ``context_tier`` param). Other providers ignore it.
     """
 
 
--- a/src/conductor/config/schema.py
+++ b/src/conductor/config/schema.py
@@ -1041,6 +1041,8 @@ class AgentDef(BaseModel):
                 raise ValueError("wait agents cannot have 'dialog'")
             if self.reasoning is not None:
                 raise ValueError("wait agents cannot have 'reasoning'")
+            if self.context_tier is not None:
+                raise ValueError("wait agents cannot have 'context_tier'")
             if self.timeout_seconds is not None:
                 raise ValueError("wait agents cannot have 'timeout_seconds'")
             if self.output is not None:
@@ -1100,6 +1102,8 @@ class AgentDef(BaseModel):
                 raise ValueError("set agents cannot have 'dialog'")
             if self.reasoning is not None:
                 raise ValueError("set agents cannot have 'reasoning'")
+            if self.context_tier is not None:
+                raise ValueError("set agents cannot have 'context_tier'")
             if self.timeout_seconds is not None:
                 raise ValueError("set agents cannot have 'timeout_seconds'")
             if self.duration is not None:
@@ -1158,6 +1162,8 @@ class AgentDef(BaseModel):
                 raise ValueError("terminate agents cannot have 'dialog'")
             if self.reasoning is not None:
                 raise ValueError("terminate agents cannot have 'reasoning'")
+            if self.context_tier is not None:
+                raise ValueError("terminate agents cannot have 'context_tier'")
             if self.workflow:
                 raise ValueError("terminate agents cannot have 'workflow'")
             if self.input_mapping is not None:
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -584,6 +584,7 @@ class WorkflowEngine:
             self._system_metadata = self._build_system_metadata()
 
         default_effort = self.config.workflow.runtime.default_reasoning_effort
+        default_tier = self.config.workflow.runtime.default_context_tier
         default_provider_name = self.config.workflow.runtime.provider.name
 
         # Resolve the provider per agent (honoring per-agent overrides).
@@ -673,6 +674,9 @@ class WorkflowEngine:
                     "provider_name": _provider_for(a.name),
                     "reasoning_effort": (
                         a.reasoning.effort if a.reasoning is not None else default_effort
+                    ),
+                    "context_tier": (
+                        a.context_tier if a.context_tier is not None else default_tier
                     ),
                 }
                 for a in self.config.agents
--- a/src/conductor/providers/copilot.py
+++ b/src/conductor/providers/copilot.py
@@ -789,6 +789,10 @@ class CopilotProvider(AgentProvider):
                 )
 
             # Resolve context tier: per-agent override wins over runtime default.
+            # Unlike reasoning effort (validated against the model's advertised
+            # supported_reasoning_efforts first), the tier is forwarded as-is:
+            # there is no advertised supported_context_tiers, so the SDK is the
+            # sole authority and validates it at session creation.
             tier = resolve_context_tier(agent, self._default_context_tier)
             if tier is not None:
                 session_kwargs["context_tier"] = tier
--- a/src/conductor/providers/copilot.py
+++ b/src/conductor/providers/copilot.py
@@ -24,6 +24,7 @@ from conductor.providers._event_format import (
 )
 from conductor.providers.base import AgentOutput, AgentProvider, EventCallback, match_model_id
 from conductor.providers.capabilities import ProviderCapabilities
+from conductor.providers.context_tier import ContextTier, resolve_context_tier
 from conductor.providers.reasoning import ReasoningEffort, resolve_reasoning_effort
 
 if TYPE_CHECKING:
@@ -204,6 +205,7 @@ class CopilotProvider(AgentProvider):
         temperature: float | None = None,
         max_agent_iterations: int | None = None,
         default_reasoning_effort: ReasoningEffort | None = None,
+        default_context_tier: ContextTier | None = None,
         provider_settings: ProviderSettings | None = None,
     ) -> None:
         """Initialize the Copilot provider.
@@ -226,6 +228,10 @@ class CopilotProvider(AgentProvider):
                 applied to ``create_session`` when an agent does not specify
                 its own ``reasoning.effort``. One of ``low``, ``medium``,
                 ``high``, ``xhigh``, or ``None`` to send no value.
+            default_context_tier: Workflow-wide default ``context_tier`` applied
+                to ``create_session`` when an agent does not specify its own
+                ``context_tier``. One of ``default``, ``long_context``, or
+                ``None`` to send no value.
             provider_settings: Optional structured provider settings from
                 ``runtime.provider``. When ``has_custom_routing()`` is True,
                 the resolved SDK ``ProviderConfig`` is attached to every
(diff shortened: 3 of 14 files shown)


franklixuefei:
@brrusino — really nice PR. The `context_tier` plumbing here is clean and the resolver/runtime-default/per-agent design is easy to follow.

A thought while you're in here: there's a **sibling staleness** in the reasoning-effort path that's the same class of issue this PR fixes for context tier — would you consider folding it into #252 (or a quick follow-up)?

**The gap:** conductor's `ReasoningEffort` literal still caps at `xhigh`:

- `providers/reasoning.py`: `ReasoningEffort = Literal["low", "medium", "high", "xhigh"]`
- `providers/capabilities.py` / `providers/copilot.py` / `providers/claude.py`: `reasoning_effort=("low", "medium", "high", "xhigh")`

…but the Copilot stack already supports a **`max`** level above `xhigh`:

- The interactive `copilot` CLI exposes it (the model picker's effort levels are Low / Medium / High / Extra High / **Max**, and `~/.copilot/settings.json` persists `"effortLevel": "max"`).
- The SDK doesn't appear to gate it either — `Session.set_model` takes `reasoning_effort: str`, and the wire layer (`rpc.py` / `session_events.py`) round-trips it as a plain `str` (`from_str`, not `parse_enum`), so `"max"` would flow through to the server rather than being rejected client-side.

So today the only thing blocking `reasoning.effort: max` in a workflow is conductor's own `Literal` rejecting it at schema-validation time — exactly the kind of "the knob moved and conductor's enum didn't follow" situation #252 is resolving for `context_tier`.

**Why this is *much* smaller than the context_tier change:** unlike `context_tier` (which legitimately needed a new field + resolver + factory threading + `create_session` forwarding), `max` needs **zero new plumbing** — the entire reasoning-effort path already exists (`ReasoningConfig.effort`, `resolve_reasoning_effort`, the `create_session` / dialog forwarding, and `default_reasoning_effort` factory threading). The *only* thing rejecting `reasoning.effort: max` today is conductor's own `Literal` at validation time. So it's purely an enum-value extension:
- add `"max"` to the `ReasoningEffort` literal + the three capability tuples
- add a `"max"` entry to `EFFORT_TO_BUDGET_TOKENS` in `reasoning.py` for the Claude/Anthropic thinking-budget path (Copilot ignores the budget and just forwards the string)
- a couple of tests asserting `effort: max` validates and reaches `create_session`

**One honest caveat** (you're better placed to confirm than I am): I'm *inferring* `max` is a real server-accepted `reasoning_effort` value from the CLI persisting `effortLevel: max` plus the SDK's str-typed wire layer — but I haven't verified the platform contract end-to-end. If `max` turns out to be a CLI-only alias, this is best skipped. You'd know the real answer.

Totally fine to keep it out of #252 if you'd rather stay tightly scoped to context tier — just flagging it since it's the same class of "the knob moved and conductor's enum didn't follow," and it's a genuinely tiny change. Either way, this PR is good to go. 🚢

franklixuefei (approved):

brrusino:
> @brrusino — really nice PR. The `context_tier` plumbing here is clean and the resolver/runtime-default/per-agent design is easy to follow.
> 
> A thought while you're in here: there's a **sibling staleness** in the reasoning-effort path that's the same class of issue this PR fixes for context tier — would you consider folding it into #252 (or a quick follow-up)?
> 
> **The gap:** conductor's `ReasoningEffort` literal still caps at `xhigh`:
> 
> * `providers/reasoning.py`: `ReasoningEffort = Literal["low", "medium", "high", "xhigh"]`
> * `providers/capabilities.py` / `providers/copilot.py` / `providers/claude.py`: `reasoning_effort=("low", "medium", "high", "xhigh")`
> 
> …but the Copilot stack already supports a **`max`** level above `xhigh`:
> 
> * The interactive `copilot` CLI exposes it (the model picker's effort levels are Low / Medium / High / Extra High / **Max**, and `~/.copilot/settings.json` persists `"effortLevel": "max"`).
> * The SDK doesn't appear to gate it either — `Session.set_model` takes `reasoning_effort: str`, and the wire layer (`rpc.py` / `session_events.py`) round-trips it as a plain `str` (`from_str`, not `parse_enum`), so `"max"` would flow through to the server rather than being rejected client-side.
> 
> So today the only thing blocking `reasoning.effort: max` in a workflow is conductor's own `Literal` rejecting it at schema-validation time — exactly the kind of "the knob moved and conductor's enum didn't follow" situation #252 is resolving for `context_tier`.
> 
> **Why this is _much_ smaller than the context_tier change:** unlike `context_tier` (which legitimately needed a new field + resolver + factory threading + `create_session` forwarding), `max` needs **zero new plumbing** — the entire reasoning-effort path already exists (`ReasoningConfig.effort`, `resolve_reasoning_effort`, the `create_session` / dialog forwarding, and `default_reasoning_effort` factory threading). The _only_ thing rejecting `reasoning.effort: max` today is conductor's own `Literal` at validation time. So it's purely an enum-value extension:
> 
> * add `"max"` to the `ReasoningEffort` literal + the three capability tuples
> * add a `"max"` entry to `EFFORT_TO_BUDGET_TOKENS` in `reasoning.py` for the Claude/Anthropic thinking-budget path (Copilot ignores the budget and just forwards the string)
> * a couple of tests asserting `effort: max` validates and reaches `create_session`
> 
> **One honest caveat** (you're better placed to confirm than I am): I'm _inferring_ `max` is a real server-accepted `reasoning_effort` value from the CLI persisting `effortLevel: max` plus the SDK's str-typed wire layer — but I haven't verified the platform contract end-to-end. If `max` turns out to be a CLI-only alias, this is best skipped. You'd know the real answer.
> 
> Totally fine to keep it out of #252 if you'd rather stay tightly scoped to context tier — just flagging it since it's the same class of "the knob moved and conductor's enum didn't follow," and it's a genuinely tiny change. Either way, this PR is good to go. 🚢

Thanks Frank, really appreciate the careful read. I went and checked the platform contract end-to-end before pulling this in, and your caveat turns out to be the deciding factor.

I queried the live models.list from the Copilot server, and no model advertises max in supportedReasoningEfforts right now - the only values any model lists are low / medium / high / xhigh. So max is a CLI-only concept today (it persists in ~/.copilot/settings.json as effortLevel), not something the per-model capability contract exposes.

The catch is that conductor's _validate_reasoning_effort_for_model validates reasoning.effort against exactly that advertised list. So if I add max to the Literal, it'd pass schema validation but then get rejected at runtime for every model - a knob that always fails, which is worse than rejecting it up front. That's different from context_tier, which actually moved server-side and is live, just unreachable from conductor.

So I'm going to keep #252 scoped to context tier and skip max for now. Happy to revisit the moment a model starts advertising it - at that point it really is the tiny enum extension you described. Thanks again for flagging the sibling staleness, it's the right instinct.

src/conductor/config/schema.py
```diff
@@ -1183,6 +1208,8 @@ def validate_agent_type(self) -> AgentDef:
                 )
         if self.type == "workflow" and self.reasoning is not None:
             raise ValueError("workflow agents cannot have 'reasoning'")
+        if self.type == "workflow" and self.context_tier is not None:
+            raise ValueError("workflow agents cannot have 'context_tier'")
```
jrob5756, line 1218: `context_tier` is rejected here for `workflow`, and just above for `human_gate` and `script`, but the `wait`, `set`, and `terminate` branches have no equivalent guard. `reasoning` is rejected on all six types, so this drops the parity the field was modeled on.

I checked it directly: `AgentDef(name='x', type='wait', duration='1s', context_tier='long_context')` validates with no error today, while the same agent with `reasoning` raises. The value then shows up in `workflow_started` metadata as if it were active.

The fix is to mirror the `reasoning` check in each of the three branches, right after the existing `reasoning` rejection:

```python
# wait branch (after the reasoning check)
if self.context_tier is not None:
    raise ValueError("wait agents cannot have 'context_tier'")
# set branch
if self.context_tier is not None:
    raise ValueError("set agents cannot have 'context_tier'")
# terminate branch
if self.context_tier is not None:
    raise ValueError("terminate agents cannot have 'context_tier'")
```

tests/test_config/test_schema.py
```diff
@@ -1399,6 +1399,105 @@ def test_rejects_invalid_effort(self, effort: object) -> None:
...
+    def test_workflow_with_context_tier_raises(self) -> None:
+        """Test that workflow agents cannot have context_tier."""
+        with pytest.raises(ValidationError) as exc_info:
+            AgentDef(
+                name="w",
+                type="workflow",
+                workflow="./sub.yaml",
+                context_tier="long_context",
+            )
+        assert "workflow agents cannot have 'context_tier'" in str(exc_info.value)
```
jrob5756, line 1472: The rejection tests cover the same three types the schema guards do, so the `wait` / `set` / `terminate` gap is not caught. These three lock in the parity and will fail on the current branch until the guards from the schema comment are added:

```suggestion
        assert "workflow agents cannot have 'context_tier'" in str(exc_info.value)

    def test_wait_with_context_tier_raises(self) -> None:
        """Test that wait agents cannot have context_tier."""
        with pytest.raises(ValidationError) as exc_info:
            AgentDef(name="w", type="wait", duration="1s", context_tier="long_context")
        assert "wait agents cannot have 'context_tier'" in str(exc_info.value)

    def test_set_with_context_tier_raises(self) -> None:
        """Test that set agents cannot have context_tier."""
        with pytest.raises(ValidationError) as exc_info:
            AgentDef(name="s", type="set", value="42", context_tier="long_context")
        assert "set agents cannot have 'context_tier'" in str(exc_info.value)

    def test_terminate_with_context_tier_raises(self) -> None:
        """Test that terminate agents cannot have context_tier."""
        with pytest.raises(ValidationError) as exc_info:
            AgentDef(
                name="t",
                type="terminate",
                status="success",
                reason="done",
                context_tier="long_context",
            )
        assert "terminate agents cannot have 'context_tier'" in str(exc_info.value)
```

src/conductor/engine/workflow.py
```diff
@@ -674,6 +675,9 @@ def _record_provider(name: str) -> None:
                     "reasoning_effort": (
                         a.reasoning.effort if a.reasoning is not None else default_effort
                     ),
+                    "context_tier": (
```
jrob5756, line 678: This per-agent `context_tier` lands in `workflow_started` metadata, but there is no event test for it. `reasoning_effort` has `test_workflow_started_includes_reasoning_effort` and `test_workflow_started_reasoning_effort_unset_is_none` in `tests/test_engine/test_event_emission.py`. Cloning those two would pin the override, runtime-default, and None-when-unset behavior on this line.

src/conductor/providers/copilot.py
```diff
@@ -781,6 +788,17 @@ async def _execute_sdk_call(
                     model,
                 )
 
+            # Resolve context tier: per-agent override wins over runtime default.
+            tier = resolve_context_tier(agent, self._default_context_tier)
+            if tier is not None:
+                session_kwargs["context_tier"] = tier
```
jrob5756, line 798: Minor: the resolved tier is forwarded without a model-capability check, whereas reasoning effort validates against the model first. That is reasonable here since the SDK validates at session creation and there is no advertised `supported_context_tiers`, but a one-line comment noting the asymmetry would save the next reader a trip through the reasoning validator.

jrob5756 (commented):
Nicely scoped change, and it tracks the existing reasoning plumbing closely. The provider and factory forwarding tests are thorough, and the example validates clean.

One thing blocks it for me: context_tier is only rejected on three of the six non-model agent types, so the same setting is silently accepted on wait, set, and terminate. The field's own docstring says it is agent-only, and the matching reasoning mistake already raises, so this reads as an enforcement gap rather than a deliberate carve-out. Inline details below. Everything else is optional.

src/conductor/config/schema.py
brrusino: Good catch, and you're right that this dropped the parity the field was modeled on. Fixed in 2c5e2bb: added the three missing guards (`wait` / `set` / `terminate`) right after each branch's existing `reasoning` rejection, so all six non-model types now reject `context_tier` the same way. Confirmed your repro (`AgentDef(name='x', type='wait', duration='1s', context_tier='long_context')`) now raises.

tests/test_config/test_schema.py
brrusino: Added these three tests verbatim from your suggestion in 2c5e2bb. They lock in the `wait` / `set` / `terminate` rejection parity and pass against the new guards.

src/conductor/engine/workflow.py
brrusino: Done in 2c5e2bb. Cloned the two reasoning_effort tests into `test_workflow_started_includes_context_tier` (explicit override + runtime-default fallback) and `test_workflow_started_context_tier_unset_is_none`, so the override / default / None-when-unset behavior on this line is now pinned.

src/conductor/providers/copilot.py
brrusino: Agreed - added a comment here in 2c5e2bb noting the asymmetry: there's no advertised `supported_context_tiers`, so unlike reasoning effort the tier is forwarded as-is and the SDK is the sole authority, validating it at session creation.

jrob5756 (approved):
LGTM

PR #263, opened 2026-06-19
feat: allow templating per-agent reasoning.effort and context_tier (#262)

### LINKED ISSUE #262: Allow templating of per-agent context_tier / reasoning.effort (resolve {{ workflow.input.* }} before enum validation)
## Summary

Per-agent `context_tier` and `reasoning.effort` are strict enums that are validated **when the workflow YAML is loaded — before `{{ workflow.input.* }}` templates are resolved**. As a result they cannot be parameterized the way `model:` can. A free-form `model: "{{ workflow.input.reviewer_1_model }}"` resolves fine, but `reasoning.effort: "{{ workflow.input.effort }}"` fails at load.

## Repro / evidence

Workflow with a templated effort:

```yaml
agents:
  - name: a1
    model: claude-opus-4.8
    reasoning:
      effort: "{{ workflow.input.eff }}"
    system_prompt: "..."
```

```
$ conductor run wf.yaml --input eff=high
Loading workflow: wf.yaml
ConfigurationError - Configuration validation failed
```

It fails at **"Loading workflow"** - the `--input eff=high` is never applied. `conductor validate` shows the underlying cause directly:

- `agents.0.reasoning.effort` -> `Input should be 'low', 'medium', 'high' or 'xhigh'`
- `agents.0.context_tier` -> `Input should be 'default' or 'long_context'`

(Observed on conductor v0.1.18 for `reasoning.effort`; on v0.1.19 the same applies to `context_tier`, added in #252.)

The literal template string `{{ workflow.input.eff }}` is not a member of the `Literal[...]` enum, so pydantic rejects it before inputs/templates are resolved. `model:` escapes this only because it is typed as a free-form `str`.

## Why it matters

Plugins want to let consumers tune reasoning depth per-repo / per-invocation **the same way they already override `model`** (via `{{ workflow.input.reviewer_1_model }}` + `--input`). Today this requires a wrapper that materializes a patched copy of the workflow with literal enum values substituted before invoking conductor (see the workaround in gim-home/ATS-Copilot PR #33, `resolve-reasoning-config.ps1`). Native support would remove that wrapper.

## Proposed fix (either approach)

1. **Resolve `{{ workflow.input.* }}` (and input defaults) before enum validation**, so a templated value is checked against the enum *after* resolution. This makes every field uniformly templatable. (Preferred.)
2. **Accept `Literal[...] | <unresolved-template-string>`** for `context_tier` / `reasoning.effort`: defer enum validation of an unresolved `{{ ... }}` to run time (validate the resolved value), and have `conductor validate` treat an unresolved template as "defer to runtime" rather than an error.

## Acceptance criteria

- `reasoning.effort: "{{ workflow.input.effort }}"` + `conductor run --input effort=high` runs with `effort=high`.
- Same for `context_tier`.
- `conductor validate` does not reject a templated enum field (it validates the resolved default, or treats an unresolved template as deferred).

## Context

- `context_tier` was introduced in #252.
- Related: `model:` already supports this because it is a free-form string.

franklixuefei:
## Summary

Allows per-agent `reasoning.effort` and `context_tier` to be templated with `{{ workflow.input.* }}` (and `{% %}` statement) Jinja2 expressions, resolved at runtime. Closes #262.

Both fields were strict pydantic `Literal` enums, so a templated value like `effort: "{{ workflow.input.eff }}"` was rejected **at YAML load** — before `--input` / Jinja2 templates are resolved. `model:` already escaped this because it is a free-form `str` resolved at runtime.

```yaml
agents:
  - name: thinker
    reasoning:
      effort: "{{ workflow.input.effort }}"   # now accepted; resolved + validated at runtime
    context_tier: "{{ workflow.input.tier }}"
```

## Approach: defer enum validation to runtime (issue Approach 2)

The issue marks Approach 1 ("resolve templates before enum validation") as preferred, but after reading the code **Approach 2 is the idiomatic, lower-risk fit** — and it mirrors a shipped, load-bearing precedent in the same file: the `wait` step's `duration` (templatable-but-validated via `_validate_wait_duration` + runtime render in `executor/wait.py`).

Why Approach 1 was rejected (grounded in the code):
- **Inputs aren't available at load.** `--input` values + `workflow.input` defaults are merged later in the engine, not in `config/loader.py`.
- **`conductor validate` has no inputs at all**, so a `{{ workflow.input.x }}` could never resolve during validate — you'd *still* need a "defer unresolved template" path.
- **Chicken-and-egg:** `workflow.input` defaults live inside the same YAML that `model_validate` parses.
- Scoping a pre-resolution to "only `workflow.input`, only these enum fields" converges right back to Approach 2's surgery anyway.

## Changes

**Schema** (`config/schema.py`)
- Widen `ReasoningConfig.effort` → `ReasoningEffort | str` + a `@model_validator(mode="after")` that defers `{{`/`{%` templates and validates literals via `get_args`.
- Widen `AgentDef.context_tier` → `ContextTier | str | None` + `_validate_context_tier()` dispatched from the regular-agent branch of `validate_agent_type`. Non-agent step types still reject the field outright (a template string is still "not None").

**Runtime** (`executor/agent.py`)
- After the existing `model` render, render + validate templated `reasoning.effort` / `context_tier` with the full agent context. Every execution path (main / retry / dialog / for_each) routes through `AgentExecutor.execute`, so this single chokepoint covers them all. A `_render_enum_field` helper renders, strips trailing whitespace, validates the resolved literal, and raises a clear `ValidationError` on a bad resolution. The provider then sees a concrete enum — no provider changes.

**Providers** (`providers/reasoning.py`, `providers/context_tier.py`)
- `resolve_*` `cast(...)` the widened field back to the strict enum — by the time they run (provider execute, after the executor resolves the field) the value is concrete and validated.

**Validator** (`config/validator.py`)
- The reasoning.effort capability cross-check defers only the *value-dependent* membership check for templates. The *value-independent* "provider supports no reasoning at all" check (`reasoning_effort=None`) still fires for templated efforts — otherwise a templated effort on a no-reasoning provider would pass `validate` while a literal errors.

## Rigor

Pre-impl + **dual-model** post-impl rubber-duck (Claude Opus 4.8 + GPT-5.5, run in parallel). Both independently flagged the validator capability gap above (fixed). Opus additionally caught a schema/executor asymmetry — the schema deferred only on `{{` while the executor renders `{{` or `{%`, so a `{% if %}` template was rejected at load though the executor would render it; both schema validators now mirror the executor guard. Verified end-to-end: `{% if workflow.input.heavy %}xhigh{% else %}low{% endif %}` resolves to `xhigh`/`low` per input.

## Tests

- **Schema**: templated effort/context_tier deferred (incl. `{% %}` statements); literal-invalid still rejected; human_gate/script/workflow/etc still reject templated fields.
- **Runtime**: render for both fields, static pass-through, no-mutation of the original templated agent, bad-resolution raises, trailing-newline strip.
- **Validator**: templated effort defers the membership check; templated effort on a no-reasoning provider still errors.
- **CLI**: `conductor validate` accepts a templated-enum workflow (exit 0), new fixture `valid_templated_enums.yaml`.

## Verification

- `ruff check` + `ruff format --check`: clean
- `ty check src`: 12 diagnostics, **== baseline** (0 introduced; all pre-existing Windows termios / dialog stubs)
- `pytest`: **3526 passed**; the 13 failures are identical on a clean `upstream/main` checkout (pre-existing Windows-env: registry TOML, event_log non-serializable, script stdin tty) — **0 new failures**.

## Out of scope (deliberate non-goals)

- Workflow-level `runtime.default_reasoning_effort` / `default_context_tier` templating — the AC is per-agent only; these remain strict enums (a templated workflow-level default is rejected with a clear message). Can follow up separately.
- A generalized `Annotated[...]` "templatable enum" type — starting with `| str` + the duration-style validator to match existing code; generalize only if a third enum field needs it.

A minor cosmetic note: the `workflow_started` telemetry event reports the raw (unrendered) per-agent template, since per-`for_each`-item resolutions differ. This dict is telemetry-only (never sent to a provider), so there's no correctness impact — flagging it as a conscious choice.

### DIFF
--- a/src/conductor/config/schema.py
+++ b/src/conductor/config/schema.py
@@ -21,6 +21,7 @@ from pydantic import (
 from conductor.duration import parse_duration
 from conductor.providers.context_tier import ContextTier
 from conductor.providers.reasoning import ReasoningEffort
+from conductor.templating import is_jinja_template
 
 BudgetMode = Literal["audit", "enforce"]
 """How the engine responds when a workflow cost budget is exceeded.
@@ -584,15 +585,21 @@ class ReasoningConfig(BaseModel):
 
     @model_validator(mode="after")
     def _validate_effort(self) -> ReasoningConfig:
-        """Accept literal efforts or defer ``{{ }}`` templates to runtime.
+        """Accept literal efforts or defer ``{{ }}`` / ``{% %}`` templates.
 
-        Mirrors :meth:`AgentDef._validate_wait_duration`: a templated value
-        (containing ``{{`` or ``{%``) skips literal validation here and is
-        rendered + validated at execute time (:mod:`conductor.executor.agent`).
-        A non-templated value must be a valid :data:`ReasoningEffort` literal.
+        A templated value (detected by
+        :func:`~conductor.templating.is_jinja_template`, matching ``{{`` or
+        ``{%``) skips literal validation here and is rendered + validated at
+        execute time (:mod:`conductor.executor.agent`, the same place the
+        ``model`` field is rendered). A non-templated value must be a valid
+        :data:`ReasoningEffort` literal.
+
+        Note: this is a broader check than
+        :meth:`AgentDef._validate_wait_duration`, which intentionally matches
+        only ``{{``.
         """
         value = self.effort
-        if isinstance(value, str) and ("{{" in value or "{%" in value):
+        if is_jinja_template(value):
             return self
         if value not in get_args(ReasoningEffort):
             raise ValueError(
@@ -1507,11 +1514,15 @@ class AgentDef(BaseModel):
     def _validate_context_tier(self) -> None:
         """Validate ``context_tier`` for a regular (provider-backed) agent.
 
-        Mirrors :meth:`_validate_wait_duration`: an unset (``None``) or
-        templated (containing ``{{`` or ``{%``) value defers all literal
-        validation to runtime (rendered + validated in
-        :mod:`conductor.executor.agent`); a non-templated value must be a
-        valid :data:`~conductor.providers.context_tier.ContextTier` literal.
+        An unset (``None``) or templated value (detected by
+        :func:`~conductor.templating.is_jinja_template`, matching ``{{`` or
+        ``{%``) defers all literal validation to runtime (rendered + validated
+        in :mod:`conductor.executor.agent`, alongside ``model``); a
+        non-templated value must be a valid
+        :data:`~conductor.providers.context_tier.ContextTier` literal.
+
+        This differs from :meth:`_validate_wait_duration` on two counts: that
+        method matches only ``{{``, and it does not defer ``None``.
 
         Non-agent step types reject ``context_tier`` outright via their own
         ``is not None`` checks in :meth:`validate_agent_type` (a template
@@ -1519,7 +1530,7 @@ class AgentDef(BaseModel):
         the regular-agent branch.
         """
         value = self.context_tier
-        if value is None or (isinstance(value, str) and ("{{" in value or "{%" in value)):
+        if value is None or is_jinja_template(value):
             return
         if value not in get_args(ContextTier):
             raise ValueError(
--- a/src/conductor/config/schema.py
+++ b/src/conductor/config/schema.py
@@ -6,7 +6,7 @@ workflow YAML configuration files.
 
 from __future__ import annotations
 
-from typing import Any, Literal
+from typing import Any, Literal, get_args
 
 from pydantic import (
     BaseModel,
@@ -563,10 +563,43 @@ class ReasoningConfig(BaseModel):
 
         reasoning:
           effort: high
+
+    Supports Jinja2 templates::
+
+        reasoning:
+          effort: "{{ workflow.input.effort }}"
+
+    A templated ``effort`` is accepted at load time and resolved + validated
+    at runtime (in :mod:`conductor.executor.agent`), mirroring how ``model``
+    and the ``wait`` step's ``duration`` are handled. A *literal* value must
+    be one of :data:`~conductor.providers.reasoning.ReasoningEffort`.
     """
 
-    effort: ReasoningEffort
-    """Reasoning effort level applied to the agent's model calls."""
+    effort: ReasoningEffort | str
+    """Reasoning effort level applied to the agent's model calls.
+
+    Either a literal level (``low`` / ``medium`` / ``high`` / ``xhigh``) or a
+    ``{{ ... }}`` Jinja2 template resolved at runtime.
+    """
+
+    @model_validator(mode="after")
+    def _validate_effort(self) -> ReasoningConfig:
+        """Accept literal efforts or defer ``{{ }}`` templates to runtime.
+
+        Mirrors :meth:`AgentDef._validate_wait_duration`: a templated value
+        (containing ``{{`` or ``{%``) skips literal validation here and is
+        rendered + validated at execute time (:mod:`conductor.executor.agent`).
+        A non-templated value must be a valid :data:`ReasoningEffort` literal.
+        """
+        value = self.effort
+        if isinstance(value, str) and ("{{" in value or "{%" in value):
+            return self
+        if value not in get_args(ReasoningEffort):
+            raise ValueError(
+                f"reasoning.effort must be one of {list(get_args(ReasoningEffort))} "
+                f"or a '{{{{ ... }}}}' template (got {value!r})"
+            )
+        return self
 
 
 class AgentDef(BaseModel):
@@ -642,7 +675,7 @@ class AgentDef(BaseModel):
     Supports Jinja2 templates: {{ workflow.input.model_name }}
     """
 
-    context_tier: ContextTier | None = None
+    context_tier: ContextTier | str | None = None
     """Context-window tier for models that support it (Copilot provider only).
 
     Set ``context_tier: long_context`` to pin a heavy-reasoning agent to the
@@ -657,9 +690,18 @@ class AgentDef(BaseModel):
 
     Only applies to provider-backed agents (type='agent' or None).
 
+    Supports Jinja2 templates: a ``{{ workflow.input.tier }}`` value is
+    accepted at load time and resolved + validated at runtime (mirrors
+    ``model`` and the ``reasoning.effort`` handling). A *literal* value must
+    be one of :data:`~conductor.providers.context_tier.ContextTier`.
+
     Example YAML::
 
         context_tier: long_context
+
+    Templated::
+
+        context_tier: "{{ workflow.input.tier }}"
     """
 
     input: list[str] = Field(default_factory=list)
@@ -1417,6 +1459,10 @@ class AgentDef(BaseModel):
                     f"'{self.type or 'agent'}' agents cannot have 'output_type' "
                     "(only 'set' agents support output_type)"
                 )
+            # #262: regular agents may carry a literal or templated
+            # context_tier; validate the literal here and defer templates to
+            # runtime. (reasoning.effort is validated on ReasoningConfig.)
+            self._validate_context_tier()
         if self.type == "workflow" and self.reasoning is not None:
             raise ValueError("workflow agents cannot have 'reasoning'")
         if self.type == "workflow" and self.context_tier is not None:
@@ -1457,6 +1503,29 @@ class AgentDef(BaseModel):
         if self.output and self.output_mode != "raw":
             return self.output
         return None
+
+    def _validate_context_tier(self) -> None:
+        """Validate ``context_tier`` for a regular (provider-backed) agent.
+
+        Mirrors :meth:`_validate_wait_duration`: an unset (``None``) or
+        templated (containing ``{{`` or ``{%``) value defers all literal
+        validation to runtime (rendered + validated in
+        :mod:`conductor.executor.agent`); a non-templated value must be a
+        valid :data:`~conductor.providers.context_tier.ContextTier` literal.
+
+        Non-agent step types reject ``context_tier`` outright via their own
+        ``is not None`` checks in :meth:`validate_agent_type` (a template
+        string is still "not None"), so this helper is only dispatched from
+        the regular-agent branch.
+        """
+        value = self.context_tier
+        if value is None or (isinstance(value, str) and ("{{" in value or "{%" in value)):
+            return
+        if value not in get_args(ContextTier):
+            raise ValueError(
+                f"context_tier must be one of {list(get_args(ContextTier))} "
+                f"or a '{{{{ ... }}}}' template (got {value!r})"
+            )
 
     def _validate_wait_duration(self) -> None:
         """Validate ``duration`` for a ``wait`` agent.
--- a/src/conductor/executor/agent.py
+++ b/src/conductor/executor/agent.py
@@ -16,6 +16,7 @@ from conductor.executor.template import TemplateRenderer
 from conductor.providers.base import AgentOutput, EventCallback
 from conductor.providers.context_tier import ContextTier
 from conductor.providers.reasoning import ReasoningEffort
+from conductor.templating import is_jinja_template
 
 
 def _verbose_log(message: str, style: str = "dim") -> None:
@@ -134,6 +135,23 @@ class AgentExecutor:
         """
         resolved = self.renderer.render(value, context).strip()
         if resolved not in allowed:
+            if not resolved:
+                # An empty resolution is almost always a conditional template
+                # (``{% if ... %}``) with no matching branch. Fail closed — the
+                # same way a non-empty invalid value (below) and the
+                # provider-side resolver guards do — rather than silently
+                # treating empty as "unset": to fall back to the runtime
+                # default, omit the field or add an else-branch emitting the
+                # desired literal.
+                raise ValidationError(
+                    f"Agent '{agent_name}': {field_name} template resolved to an empty value.",
+                    suggestion=(
+                        f"A conditional template with no matching branch "
+                        f"produced nothing. Emit one of {list(allowed)}, add an "
+                        f"else-branch, or omit {field_name} to use the runtime "
+                        f"default."
+                    ),
+                )
             raise ValidationError(
                 f"Agent '{agent_name}': {field_name} template resolved to "
                 f"{resolved!r}, which is not a valid value.",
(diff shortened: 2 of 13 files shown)


franklixuefei:
@jrob5756 kindly requesting your review on this, thanks Jason!

franklixuefei:
@jrob5756 kindly requesting your review on this, thanks Jason!

src/conductor/providers/reasoning.py
```diff
@@ -87,5 +87,10 @@ def resolve_reasoning_effort(
     set, signalling that no reasoning parameter should be sent to the SDK.
     """
     if agent.reasoning is not None:
-        return agent.reasoning.effort
+        # #262: AgentDef widens ``effort`` to ``ReasoningEffort | str`` so a
+        # ``{{ ... }}`` template survives schema validation. By the time this
+        # resolver runs (provider execute, after AgentExecutor renders +
+        # validates the field) the value is a concrete, validated literal, so
+        # narrowing back to ReasoningEffort is sound.
+        return cast(ReasoningEffort, agent.reasoning.effort)
```
jrob5756: This cast is a runtime no-op, so its safety rests on an invariant set elsewhere: that `AgentExecutor` rendered and validated the field before the provider runs. That holds on every current path, but nothing enforces it. A future caller that reaches a resolver with an unrendered agent would send a raw `{{ ... }}` string straight to the SDK. A guard makes the invariant self-checking, and it lets me fix the comment in passing: `effort` lives on `ReasoningConfig`, not `AgentDef`.

```suggestion
        # #262: ReasoningConfig widens ``effort`` to ``ReasoningEffort | str``
        # so a ``{{ ... }}`` template survives schema load; AgentExecutor
        # renders + validates it before provider execute. Guard the invariant
        # here so an unrendered template raises instead of being cast to the SDK.
        effort = agent.reasoning.effort
        if isinstance(effort, str) and ("{{" in effort or "{%" in effort):
            raise ValueError(f"reasoning.effort reached the provider unresolved: {effort!r}")
        return cast(ReasoningEffort, effort)
```

src/conductor/providers/context_tier.py
```diff
@@ -39,5 +39,10 @@ def resolve_context_tier(
         The resolved ``ContextTier``, or ``None`` to send no value.
     """
     if agent.context_tier is not None:
-        return agent.context_tier
+        # #262: AgentDef widens ``context_tier`` to ``ContextTier | str`` so a
+        # ``{{ ... }}`` template survives schema validation. By the time this
+        # resolver runs (provider execute, after AgentExecutor renders +
+        # validates the field) the value is a concrete, validated literal, so
+        # narrowing back to ContextTier is sound.
+        return cast(ContextTier, agent.context_tier)
```
jrob5756: Same hardening as `reasoning.py`, and `context_tier` is the riskier of the two. Copilot forwards it to the SDK with no validation (the comment at copilot.py:806 says the SDK is the sole authority), so there is no per-model backstop the way reasoning effort has. An unrendered template here would reach the SDK with nothing to catch it.

```suggestion
        # #262: AgentDef widens ``context_tier`` to ``ContextTier | str`` so a
        # ``{{ ... }}`` template survives schema load; AgentExecutor renders +
        # validates it before provider execute. Guard the invariant here so an
        # unrendered template raises instead of being cast to the SDK (Copilot
        # forwards this value to the SDK unvalidated).
        tier = agent.context_tier
        if isinstance(tier, str) and ("{{" in tier or "{%" in tier):
            raise ValueError(f"context_tier reached the provider unresolved: {tier!r}")
        return cast(ContextTier, tier)
```

src/conductor/executor/agent.py
```diff
@@ -154,6 +183,40 @@ async def execute(
...
             agent = agent.model_copy(update={"model": rendered_model})
 
+        # #262: resolve templated reasoning.effort / context_tier the same
+        # way model is handled above. These fields are strict enums that the
+        # schema deliberately accepts as templates (deferring literal
+        # validation to here); render the value with full context, then
+        # validate the resolved literal so the provider sees a concrete enum.
+        # The ``isinstance(..., str)`` guards both detect templates and narrow
+        # the widened ``Enum | str | None`` field types to ``str`` for the
+        # type checker before passing them to ``_render_enum_field``.
```
jrob5756: Small accuracy fix: `ReasoningEffort` and `ContextTier` are `Literal` aliases, not Enums. The code relies on that by calling `get_args`, which does not work on `Enum`. The `Enum | str | None` type expression here does not correspond to a real type.

```suggestion
        # The ``isinstance(..., str)`` guards both detect templates and narrow
        # the widened ``ReasoningEffort | str`` / ``ContextTier | str | None``
        # (Literal-or-template) field types to ``str`` for the type checker
        # before passing them to ``_render_enum_field``.
```

src/conductor/config/schema.py
```diff
@@ -563,10 +563,43 @@ class ReasoningConfig(BaseModel):
...
-    effort: ReasoningEffort
-    """Reasoning effort level applied to the agent's model calls."""
+    @model_validator(mode="after")
+    def _validate_effort(self) -> ReasoningConfig:
+        """Accept literal efforts or defer ``{{ }}`` templates to runtime.
+
+        Mirrors :meth:`AgentDef._validate_wait_duration`: a templated value
+        (containing ``{{`` or ``{%``) skips literal validation here and is
+        rendered + validated at execute time (:mod:`conductor.executor.agent`).
+        A non-templated value must be a valid :data:`ReasoningEffort` literal.
```
jrob5756: This says it mirrors `_validate_wait_duration`, but that method checks only `{{`, not `{%`. The guard that actually matches this `{{`-or-`{%` predicate is the executor's `model` render. Worth rewording so nobody later "syncs" the two methods and quietly drops `{%` handling.

```suggestion
        Follows the defer-to-runtime pattern of the executor's ``model`` guard
        (cf. :meth:`AgentDef._validate_wait_duration`, which defers only
        ``{{``): a templated value (containing ``{{`` or ``{%``) skips literal
        validation here and is rendered + validated at execute time
        (:mod:`conductor.executor.agent`). A non-templated value must be a
        valid :data:`ReasoningEffort` literal.
```

src/conductor/config/schema.py
```diff
@@ -1458,6 +1504,29 @@ def effective_output_schema(self) -> dict[str, OutputField] | None:
...
         return None
 
+    def _validate_context_tier(self) -> None:
+        """Validate ``context_tier`` for a regular (provider-backed) agent.
+
+        Mirrors :meth:`_validate_wait_duration`: an unset (``None``) or
+        templated (containing ``{{`` or ``{%``) value defers all literal
+        validation to runtime (rendered + validated in
+        :mod:`conductor.executor.agent`); a non-templated value must be a
+        valid :data:`~conductor.providers.context_tier.ContextTier` literal.
```
jrob5756: Same as on `_validate_effort`: `_validate_wait_duration` checks only `{{` and never defers `None`, so the mirror claim is off on two counts here. Reword to point at the executor's `model` guard instead.

```suggestion
        Follows the executor's ``model`` defer-to-runtime guard (cf.
        :meth:`_validate_wait_duration`, which defers only ``{{`` and never
        ``None``): an unset (``None``) or templated (containing ``{{`` or
        ``{%``) value defers all literal validation to runtime (rendered +
        validated in :mod:`conductor.executor.agent`); a non-templated value
        must be a valid :data:`~conductor.providers.context_tier.ContextTier`
        literal.
```

jrob5756 (commented):
I ran a full pass over this: code quality, tests, types, comments, and error handling. The implementation is correct. The widen-at-schema, defer, render-in-executor, cast-in-resolver chain holds on every path I traced, and the new tests pass. No blocking issues.

A few non-blocking things, in rough priority:

Harden the two resolver casts (see the inline notes on reasoning.py and context_tier.py). The cast is sound today only because the executor always renders first, and the one direct provider.execute caller (the output validator) builds a synthetic agent without these fields. Nothing enforces that, so a guard is cheap insurance against a future caller.

The ("{{" in v or "{%" in v) check is now duplicated across five sites: both schema validators, the two executor blocks, and the capability cross-check. A shared is_jinja_template() helper keeps them from drifting. One catch worth flagging: it has to return TypeGuard[str], not bool, or the executor loses the narrowing it depends on and ty fails. A leaf module such as conductor/templating.py avoids an import cycle. Leave _validate_wait_duration as-is; its {{-only check is deliberate.

Two test gaps. Nothing renders a {% %} statement template through the executor, which is the only layer that actually renders it; if that guard regressed, the schema and validator tests would still pass. Separately, an empty resolution (a {% if %} with no else) raises instead of falling back to the runtime default, which is a real behavior choice worth pinning with a test.

Pre-existing, not introduced here: a context_tier set on a non-Copilot provider is silently dropped, with no capability check and no warning, unlike reasoning effort which is rejected at validate time. Templating extends that silent drop to the templated path. Probably a follow-up.

src/conductor/providers/reasoning.py
franklixuefei: Done (`a4c698c`) — added the guard and fixed the comment to say `ReasoningConfig` widens `effort` (not `AgentDef`). It now raises if an unrendered template reaches the resolver instead of casting it to the SDK. One small adaptation from the suggestion: instead of the inline `isinstance(effort, str) and ("{{" in effort or "{%" in effort)`, it calls the shared `is_jinja_template()` helper from your next note — same predicate, single home.

src/conductor/providers/context_tier.py
franklixuefei: Done (`a4c698c`) — same guard, via `is_jinja_template()`. The comment calls out that this is the riskier of the two: Copilot forwards the tier to the SDK unvalidated (no advertised `supported_context_tiers`, so the SDK is the sole authority), so a leaked template here has no per-model backstop the way effort does. Thanks for flagging the asymmetry.

src/conductor/executor/agent.py
franklixuefei: Fixed (`a4c698c`) — reworded to `ReasoningEffort | str` / `ContextTier | str | None` and noted explicitly that these are `Literal` aliases (not Enums), which is why the code uses `get_args`. The text now refers to `is_jinja_template`, since the predicate moved into the shared helper.

src/conductor/config/schema.py
franklixuefei: Reworded (`a4c698c`) — it no longer claims to mirror `_validate_wait_duration` (which is `{{`-only). It points at the executor's `model` defer-to-runtime guard as the real analog, references `is_jinja_template` for the `{{`-or-`{%` predicate, and keeps a parenthetical that `_validate_wait_duration` is intentionally `{{`-only — so nobody later "syncs" them and quietly drops `{%`.

src/conductor/config/schema.py
franklixuefei: Reworded (`a4c698c`) — same fix, and it now spells out that `_validate_wait_duration` differs on *two* counts (only `{{`, and it never defers `None`), pointing at the executor's `model` guard as the analog instead.

franklixuefei:
> I ran a full pass over this: code quality, tests, types, comments, and error handling. The implementation is correct. The widen-at-schema, defer, render-in-executor, cast-in-resolver chain holds on every path I traced, and the new tests pass. No blocking issues.
> 
> A few non-blocking things, in rough priority:
> 
> Harden the two resolver casts (see the inline notes on reasoning.py and context_tier.py). The cast is sound today only because the executor always renders first, and the one direct provider.execute caller (the output validator) builds a synthetic agent without these fields. Nothing enforces that, so a guard is cheap insurance against a future caller.
> 
> The ("{{" in v or "{%" in v) check is now duplicated across five sites: both schema validators, the two executor blocks, and the capability cross-check. A shared is_jinja_template() helper keeps them from drifting. One catch worth flagging: it has to return TypeGuard[str], not bool, or the executor loses the narrowing it depends on and ty fails. A leaf module such as conductor/templating.py avoids an import cycle. Leave _validate_wait_duration as-is; its {{-only check is deliberate.
> 
> Two test gaps. Nothing renders a {% %} statement template through the executor, which is the only layer that actually renders it; if that guard regressed, the schema and validator tests would still pass. Separately, an empty resolution (a {% if %} with no else) raises instead of falling back to the runtime default, which is a real behavior choice worth pinning with a test.
> 
> Pre-existing, not introduced here: a context_tier set on a non-Copilot provider is silently dropped, with no capability check and no warning, unlike reasoning effort which is rejected at validate time. Templating extends that silent drop to the templated path. Probably a follow-up.

Thanks @jrob5756! I addressed all comments.

re: **shared `is_jinja_template()` helper (`TypeGuard[str]`, leaf module)**

Done (`a4c698c`) — added `conductor/templating.py` with `is_jinja_template(value) -> TypeGuard[str]`, a leaf module with no intra-`conductor` imports so there's no cycle. Routed the duplicated sites through it: both schema validators, the two executor blocks, and the capability cross-check — plus the `model` render block (same predicate, same function, so leaving it inline would invite exactly the drift you're guarding against) and the two new resolver guards. `_validate_wait_duration` stays as-is per your note (its `{{`-only check is deliberate), as does router.py's `{{`+`}}` `when` check (a different predicate). Confirmed it returns `TypeGuard[str]`, not `bool` — ty passes and the executor keeps its narrowing into `_render_enum_field`. Added a small unit test for the helper itself (detects `{{` and `{%`, rejects plain strings and non-strings — `None` in particular, so the `model` guard could drop its old `agent.model and …` short-circuit).

re: **test gap (`{% %}` statement through the executor)**

Added (`a4c698c`) — `test_statement_template_effort_is_rendered` drives a `{% if … %}…{% else %}…{% endif %}` statement template through the executor (the only layer that renders), plus a matching one for `context_tier`. As you said, the schema/validator tests would have stayed green if that guard regressed, so these pin the actual render path.

re: **review summary: empty-resolution behavior choice**

Pinned (`a4c698c`) — `test_effort_resolving_to_empty_raises` covers a `{% if … %}…{% endif %}` with no matching branch (resolves to `""`).

On the behavior itself, I made a deliberate call to keep it fail-closed (raise) rather than fall back to the runtime default, and gave the empty case its own actionable message ("…resolved to an empty value" + a hint to add an else-branch or omit the field). The reasoning: it's consistent with the surrounding contracts — `_render_enum_field` already rejects any value not in `allowed`, and the new B1/B2 resolver guards are fail-closed too — so treating empty as a special "unset → default" would add a third, silent meaning for empty that could mask a genuinely broken conditional. It also keeps this a pure pin rather than a behavior change with a type wrinkle: `ReasoningConfig.effort` is `ReasoningEffort | str` (not None-able), so "fall back to default" would mean making the override itself droppable, not just resolving it.

If you'd rather it fall back to the runtime default (treat an empty conditional as "no override"), I'm happy to flip it — I mainly wanted to pin the current behavior and surface the choice explicitly, as you suggested.

re: **`context_tier` silently dropped on non-Copilot providers (pre-existing)**

Agreed this is worth a follow-up, and that it's pre-existing — templating just extends the existing silent drop to the templated path. It's a real asymmetry vs reasoning effort (which is rejected at validate time via the capability cross-check). I'd keep it out of this PR to keep the scope tight; happy to open an issue to add a capability-style check (or at least a warning) for `context_tier` on providers that don't consume it. Let me know if you'd like me to file that.

jrob5756:
Thanks for turning these around fast. One snag: commit `a4c698c` that all five replies point to isn't on the branch.

The head is still `f5d5ee52` from June 19 (2 commits, both pre-review), and when I pull the live files none of the changes are there:

- `reasoning.py` / `context_tier.py` resolvers still do a bare `cast` with no guard
- the shared `is_jinja_template()` helper and `templating.py` module 404
- the `agent.py` comment still says `Enum | str | None`
- the schema docstrings still claim they mirror `_validate_wait_duration`

Looks like the push landed on a different branch, or got dropped. Could you push `a4c698c` to `xuefl/templatable-reasoning-enums`? Once the head moves I'll re-verify. None of it's blocking; I just want the branch history to match the replies.

franklixuefei:
> Thanks for turning these around fast. One snag: commit `a4c698c` that all five replies point to isn't on the branch.
> 
> The head is still `f5d5ee52` from June 19 (2 commits, both pre-review), and when I pull the live files none of the changes are there:
> 
> * `reasoning.py` / `context_tier.py` resolvers still do a bare `cast` with no guard
> * the shared `is_jinja_template()` helper and `templating.py` module 404
> * the `agent.py` comment still says `Enum | str | None`
> * the schema docstrings still claim they mirror `_validate_wait_duration`
> 
> Looks like the push landed on a different branch, or got dropped. Could you push `a4c698c` to `xuefl/templatable-reasoning-enums`? Once the head moves I'll re-verify. None of it's blocking; I just want the branch history to match the replies.

Ah sorry about that... now my local commit is pushed, please check again. Thanks Jason!

jrob5756 (approved):
LGTM. Re-verified at `7d42fb06`: all eight points are in, the shared `is_jinja_template` helper is wired through every site, both resolver guards are there, and the docstrings/comments read correctly now. Lint and format clean, `ty` is at baseline (the lone `dialog_evaluator` diagnostic is pre-existing), and the affected suites are green (357 passed). Thanks for the thorough turnaround.

PR #268, opened 2026-06-24
fix(copilot): capture resolved model for auto-routed runs

ldavidgomez:
## Summary

This PR captures the concrete model resolved by the Copilot SDK for auto-routed runs.

Previously, when an agent was configured with `model: auto`, Conductor recorded `AgentOutput.model` as `"auto"` even though the Copilot SDK had resolved the request to a concrete model. Token usage was available, but downstream accounting could not attribute that usage to the actual model.

This change captures `event.data.model` from the SDK `assistant.usage` event, stores it as `SDKResponse.resolved_model`, preserves it through response reconstruction paths, and uses it for `AgentOutput.model` when available.

## What changed

- Add `resolved_model` to `SDKResponse`.
- Capture the resolved model from `assistant.usage`.
- Preserve `resolved_model` through `_execute_sdk_call`.
- Use the resolved model when constructing `AgentOutput`.
- Keep existing fallback behavior when no resolved model is available.
- Add tests for:
  - resolved model propagation;
  - fallback behavior;
  - cost calculation when the resolved model is priceable;
  - SDK `assistant.usage` event extraction.

## Validation

Unit tests:

- `uv run pytest tests/test_providers/test_copilot.py::TestCopilotProviderResolvedModel -v` — passed
- `uv run pytest tests/test_providers/test_copilot.py -v` — passed

Local smoke tests with Copilot:

- `model: auto` structured-output run resolved to a concrete model instead of `"auto"`.
- explicit resolved-model run succeeded.
- `model: auto` no-schema/raw-output run resolved to a concrete model instead of `"auto"`.

Observed resolved models included `gpt-5.3-codex` / `gpt-5.4-mini` depending on the run.

## Notes

This PR does not attempt to calculate GitHub Copilot credit or USD cost. Copilot pricing/accounting is plan and policy dependent and may change independently from the SDK. The change records the concrete resolved model so downstream pricing or accounting logic can make an informed decision instead of receiving `"auto"`.

### DIFF
--- a/src/conductor/providers/copilot.py
+++ b/src/conductor/providers/copilot.py
@@ -121,8 +121,9 @@ class SDKResponse:
         cache_read_tokens: Tokens read from cache (if available).
         cache_write_tokens: Tokens written to cache (if available).
         partial: Whether this response is partial (from a mid-agent interrupt).
-        resolved_model: Actual model name reported by the SDK in the assistant.usage
-            event. None if the event was not emitted (e.g., error paths).
+        resolved_model: Model name the SDK reported in the assistant.usage event's
+            model field. None when that event is absent (error or interrupt paths)
+            or carries no usable model name (the model field is missing or empty).
     """
 
     content: str
@@ -576,8 +577,11 @@ class CopilotProvider(AgentProvider):
                     input_tokens=input_tokens,
                     output_tokens=output_tokens,
                     cache_read_tokens=cache_read,
-                    cache_write_tokens=cache_write,
-                    model=(sdk_response.resolved_model if sdk_response else None)
+                    model=(
+                        sdk_response.resolved_model
+                        if sdk_response and agent.model in (None, "auto")
+                        else None
+                    )
                     or agent.model
                     or self._default_model,
                     partial=is_partial,
@@ -1240,6 +1244,7 @@ class CopilotProvider(AgentProvider):
         session: Any,
         guidance: str,
         agent_name: str | None = None,
+        agent_model: str | None = None,
     ) -> AgentOutput:
         """Send follow-up guidance to an interrupted session.
 
@@ -1256,6 +1261,8 @@ class CopilotProvider(AgentProvider):
                 interrupts only fire on sequential agents (for-each iterations
                 do not forward ``interrupt_signal`` to the executor), so the
                 tag, when supplied, is the unqualified agent name.
+            agent_model: Optional configured model for the interrupted agent,
+                used when the follow-up response does not report a resolved model.
 
         Returns:
             AgentOutput with the follow-up response content.
@@ -1292,7 +1299,13 @@ class CopilotProvider(AgentProvider):
                 output_tokens=sdk_response.output_tokens,
                 cache_read_tokens=sdk_response.cache_read_tokens,
                 cache_write_tokens=sdk_response.cache_write_tokens,
-                model=sdk_response.resolved_model or self._default_model,
+                model=(
+                    sdk_response.resolved_model
+                    if agent_model in (None, "auto")
+                    else None
+                )
+                or agent_model
+                or self._default_model,
             )
         finally:
             await session.disconnect()
--- a/src/conductor/providers/copilot.py
+++ b/src/conductor/providers/copilot.py
@@ -577,6 +577,7 @@ class CopilotProvider(AgentProvider):
                     input_tokens=input_tokens,
                     output_tokens=output_tokens,
                     cache_read_tokens=cache_read,
+                    cache_write_tokens=cache_write,
                     model=(
                         sdk_response.resolved_model
                         if sdk_response and agent.model in (None, "auto")
@@ -1299,11 +1300,7 @@ class CopilotProvider(AgentProvider):
                 output_tokens=sdk_response.output_tokens,
                 cache_read_tokens=sdk_response.cache_read_tokens,
                 cache_write_tokens=sdk_response.cache_write_tokens,
-                model=(
-                    sdk_response.resolved_model
-                    if agent_model in (None, "auto")
-                    else None
-                )
+                model=(sdk_response.resolved_model if agent_model in (None, "auto") else None)
                 or agent_model
                 or self._default_model,
             )
--- a/src/conductor/providers/copilot.py
+++ b/src/conductor/providers/copilot.py
@@ -121,6 +121,8 @@ class SDKResponse:
         cache_read_tokens: Tokens read from cache (if available).
         cache_write_tokens: Tokens written to cache (if available).
         partial: Whether this response is partial (from a mid-agent interrupt).
+        resolved_model: Actual model name reported by the SDK in the assistant.usage
+            event. None if the event was not emitted (e.g., error paths).
     """
 
     content: str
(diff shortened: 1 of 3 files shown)


src/conductor/providers/copilot.py
```diff
@@ -574,7 +577,9 @@ async def _execute_with_retry(
                     output_tokens=output_tokens,
                     cache_read_tokens=cache_read,
                     cache_write_tokens=cache_write,
-                    model=agent.model or self._default_model,
+                    model=(sdk_response.resolved_model if sdk_response else None)
+                    or agent.model
+                    or self._default_model,
```
jrob5756: This prefers the SDK's resolved model on every run, not just when `model: auto`. Since `assistant.usage.model` is always set, an explicitly configured priceable model gets replaced by the SDK's name too. If that name isn't in the pricing table (some dotted Copilot IDs aren't), `cost_usd` becomes None and is dropped from the budget total in `usage.py`, so spend is undercounted with no error.

If the intent is auto-routing, gating to that case avoids the regression:

```suggestion
                    model=(
                        sdk_response.resolved_model
                        if sdk_response and agent.model in (None, "auto")
                        else None
                    )
                    or agent.model
                    or self._default_model,
```

If you'd rather keep the universal override for parity with `claude_agent_sdk.py`, the alternative is to take `resolved_model` only when `get_pricing()` recognizes it and keep the user's name otherwise.

src/conductor/providers/copilot.py
```diff
@@ -121,6 +121,8 @@ class SDKResponse:
         cache_read_tokens: Tokens read from cache (if available).
         cache_write_tokens: Tokens written to cache (if available).
         partial: Whether this response is partial (from a mid-agent interrupt).
+        resolved_model: Actual model name reported by the SDK in the assistant.usage
+            event. None if the event was not emitted (e.g., error paths).
```
jrob5756: `resolved_model` is also None when the `assistant.usage` event arrives without a `model` (the `if sdk_model:` guard drops an empty string) and on the partial/interrupt return, not only on error paths. As written, a reader could assume a successful turn always sets it.

```suggestion
        resolved_model: Model name the SDK reported in the assistant.usage event's
            model field. None when that event is absent (error or interrupt paths)
            or carries no usable model name (the model field is missing or empty).
```

src/conductor/providers/copilot.py
```diff
@@ -1271,7 +1292,7 @@ async def send_followup(
                 output_tokens=sdk_response.output_tokens,
                 cache_read_tokens=sdk_response.cache_read_tokens,
                 cache_write_tokens=sdk_response.cache_write_tokens,
-                model=self._default_model,
+                model=sdk_response.resolved_model or self._default_model,
```
jrob5756: Pre-existing and minor: this skips the `agent.model` tier the main path has, so a follow-up that doesn't re-resolve a model reports `self._default_model` instead of the agent's configured one. `send_followup` only receives `agent_name`, but the caller in `workflow.py` has the full `agent` if you ever want to thread it through. Still better than the old hard-coded default.

tests/test_providers/test_copilot.py
```diff
@@ -1747,3 +1747,127 @@ async def test_tier_and_reasoning_compose(self, monkeypatch: pytest.MonkeyPatch)
...
+        async def noop() -> None:
+            return None
+
+        monkeypatch.setattr(provider, "_ensure_client_started", noop)
+        monkeypatch.setattr(provider, "_send_and_wait", fake_send)
+
+        result = await provider.execute(agent=agent, context={}, rendered_prompt="p")
+        assert result.model == "auto"
+
+    def test_resolved_model_enables_cost_calculation(self) -> None:
```
jrob5756: This passes on `main` without the PR. It builds an `AgentOutput` directly with a priceable model and checks that `UsageTracker` prices it, which is existing behavior, so it doesn't exercise the resolved-model capture or propagation at all. To pin what the PR adds, run it through `execute()` with a fake `_send_and_wait` that returns a resolved model, price the result, and assert the `auto` case still comes out as None for contrast.

tests/test_providers/test_copilot.py
```diff
@@ -1747,3 +1747,127 @@ async def test_tier_and_reasoning_compose(self, monkeypatch: pytest.MonkeyPatch)
...
+
+
+class TestCopilotProviderResolvedModel:
+    """Tests for SDKResponse.resolved_model propagation into AgentOutput.model."""
+
+    # These are mocked tests; model availability in the live Copilot environment is not required.
+    # claude-sonnet-4 is used in pricing/usage tests as the canonical model and exists in the
+    # Conductor pricing table, so it satisfies both propagation and cost-calculation tests.
+    _RESOLVED_MODEL_FROM_SDK = "claude-sonnet-4"
+    _PRICEABLE_MODEL = "claude-sonnet-4"
```
jrob5756: Both values are dashed and priceable, so the tests never reach the case that actually carries risk: an explicit priceable model where the SDK reports a different name that isn't priceable. Something like `agent.model='gpt-4o'` with a resolved `'claude-sonnet-4.5'` would pin whichever precedence you settle on above.

jrob5756 (commented):
Reviewed at the PR head in an isolated checkout. Lint, format, typecheck, and the full copilot suite (88 tests) pass, and the approach matches how claude_agent_sdk.py already reads msg.model from the SDK, so nothing here blocks merge.

My main note is one behavioral point. The capture isn't limited to model: auto: assistant.usage.model is always set, so the resolved name replaces the configured model on every run. For an explicitly pinned priceable model that can swap in a name the pricing table doesn't recognize (some dotted Copilot IDs like claude-sonnet-4.5 and gemini-2.5-pro aren't in it), and then cost_usd goes None and quietly drops out of the budget total in usage.py, so spend can be undercounted with no error. Suggestion inline.

Two smaller things. claude.py still reports the configured model rather than the SDK-served one, so the three providers now diverge on this. Anthropic has no auto-routing so it's a reasonable choice, but AGENTS.md treats model as a parity-tracked field, so a line in the PR description would make the divergence intentional. And the parsed-JSON branch that agents with an output schema take isn't covered by the new tests, so its resolved_model propagation could regress unnoticed.

ldavidgomez:
Updated.

The main execution path now only uses the SDK `resolved_model` when the configured agent model is `None` or `"auto"`, so explicitly configured models remain authoritative for reporting and pricing.

I also threaded the configured agent model into the Copilot follow-up path and applied the same precedence there:
- `auto` / default-routed agents use the SDK resolved model when available;
- explicitly configured agent models are preserved;
- provider default remains the final fallback.

I also updated the `resolved_model` docstring to clarify that it can be `None` when the usage event is absent or when the event does not carry a usable model name.

Validation:
- `uv run pytest tests/test_providers/test_copilot.py -k resolved_model -q`
- `uv run pytest tests/test_providers/test_copilot.py -q`

jrob5756 (approved):
All five review comments are addressed (auto-gating, docstring, follow-up path threads agent.model, and tests now exercise execute() including the explicit-priceable + unpriceable-resolved precedence case). Pushed db188c8 to restore the dropped cache_write_tokens on the main-path AgentOutput (fixed the F841 lint + a silent cache-write accounting regression) and a format fix. CI is green.

Issue #241, opened 2026-06-02
Introduce experimental provider tier with capability declaration + validate-time cross-checks


jrob5756:
## Summary

Introduce a formal **experimental provider tier** with a **provider capability declaration** so that workflows can be statically validated against what each provider actually supports.

Motivated by two in-flight provider PRs (#104 `claude-agent-sdk`, #235 `hermes`) that exposed the gap: both delegate parts of the agentic loop to an upstream SDK/framework and therefore can't honor every parity rule in `AGENTS.md` (streaming events, MCP passthrough, `reasoning.effort`, structured-output schema, resume, etc.). Rather than rejecting these case-by-case or silently letting parity rules erode, formalize a tier with explicit allowed carve-outs and a static validator that catches workflow ↔ provider mismatches at `conductor validate` time.

## Goals

- Land a small, opinionated framework that #104 can become the first member of, and #235 can be evaluated against.
- Keep the supply-chain blast radius small (per-provider extras, no bulk `[experimental]` extra).
- Surface mismatches early via `conductor validate`; surface runtime mismatches via console + JSONL.
- Avoid experimental becoming permanent purgatory: written promotion criteria.

## Non-goals (for v1)

- A plugin / entry-point architecture. Tier is metadata, not package structure.
- A runtime `--allow-experimental` opt-in. Optional extras already gate installation.
- Automatic feature shimming. If a feature isn't supported, surface it — don't silently swallow.
- A separate `conductor.providers.experimental` namespace. Promotion shouldn't require an import-path change.

### 1. Provider capability descriptor (applies to ALL providers, not just experimental)

Each provider declares its capabilities so `conductor validate` and the runtime can cross-check. Capabilities are orthogonal to tier — Copilot/Claude declare them too. Without that, the validator only catches issues in experimental, which is backwards.

Proposed mechanism (TBD during implementation): a `ProviderCapabilities` Pydantic model attached to each `AgentProvider` subclass as a class-level `CAPABILITIES` attribute. The factory exposes capabilities for a configured provider without needing to instantiate it (so `validate` works without API keys).

Proposed initial capability vocabulary (lock this list before letting providers declare against it):

| Capability | Type | Notes |
|---|---|---|
| `tier` | `stable \| experimental` | The tier label itself lives on the descriptor. |
| `mcp_tools` | `bool` | Honors `runtime.mcp_servers` and workflow `tools:`. |
[...]

PR #269, opened 2026-06-24
fix(claude-agent-sdk): grant default tool preset when `tools:` is omitted

lukeellison:
## Summary

An agent that omits `tools:` on the `claude-agent-sdk` provider was receiving **zero** tools instead of the full `claude_code` preset. A simple "read a file and answer" agent came up with no filesystem tools and failed.

## Root cause

The distinction between *omitted* `tools:` (means "all tools") and explicit `tools: []` (means "no tools") was lost at the executor→provider boundary. `resolve_agent_tools(agent.tools, workflow_tools)` returns an empty list for an omitted `tools:` whenever the workflow declares no MCP tools — the common case for this provider, which rejects `runtime.mcp_servers` at the factory. Both cases therefore arrived at the provider as `[]`, so the provider disabled all tools for an agent that had simply forgotten to declare any.

## What changed

- **Provider** (`claude_agent_sdk.py`): disambiguate from the raw `agent.tools` field — the only place the omitted-vs-explicit signal survives.
  - omitted (`agent.tools is None`) → `claude_code` preset + `bypassPermissions` (matches the bare `claude` CLI)
  - explicit `tools: []` → no tools, no permission bypass
  - explicit non-empty list → refused loudly (workflow tool names don't map to Claude CLI tool IDs)
- **Validator** (`config/validator.py`): a workflow-level `tools:` block combined with an agent that omits `tools:` would inherit a non-empty list at runtime and hit that same refusal mid-run, with a misleading *"declares tools=[…]"* message. `conductor validate` now rejects this early — for both top-level agents and `for_each` inline agents — keyed off each agent's *resolved* provider, so mixed-provider workflows aren't over-rejected.
- **Docs**: the README and provider comparison claimed workflow `tools`/`mcp_servers` are "ignored" — they aren't (MCP rejected at the factory, tools rejected at validate). Corrected, plus a new `examples/claude-agent-sdk-repo-qa.yaml` demonstrating the fix.

## Scope

Deliberately tight — this fixes the omitted-tools bug and the directly-coupled validate-time footgun, nothing more.

**Out of scope:** inline `for_each` agents currently receive only the *tools* capability cross-check; the other per-agent capability checks (reasoning effort, MCP, output schema, session timeout) remain top-level-only. This was deemed out of scope to keep the PR focused — a follow-up issue should be filed to extend the remaining capability checks to inline `for_each` agents.

## Testing

- **Provider**: omitted → preset (incl. end-to-end through `AgentExecutor`), explicit `[]` → no tools, explicit non-empty → raises.
- **Validator**: inherited-tools rejection for top-level + `for_each` inline agents; explicit `[]` and passthrough providers still pass.
- `make check` clean (lint + typecheck); full test suite green under CI markers (`-m "not real_api and not performance"`); all examples validate.

## Related

Builds on the experimental-provider capability framework (#241).


lukeellison:

tests/test_providers/test_claude_agent_sdk.py
```diff
@@ -616,14 +620,145 @@ async def fake_query(**kwargs):
...
+            with pytest.raises(ProviderError, match="does not support workflow tool allowlists"):
+                await provider.execute(
+                    agent=agent,
+                    context={},
+                    rendered_prompt="hi",
+                    tools=["search", "read_file"],
+                )
+
+
+class TestOmittedToolsDefaultPreset:
```
jrob5756, line 632: Heads up: this module is gated by `pytest.importorskip("claude_agent_sdk")` (line 11), and CI runs `uv sync --group dev` without the `claude-agent-sdk` extra (it lives in `[project.optional-dependencies]`, not the `dev` group). So every test in this class, plus the end-to-end one below, is skipped in CI, which leaves the actual behavior this PR fixes with no automated guard. A later change that reverts `_resolve_tool_config` back to `if tools is None` would stay green.

Installing the extra in the test job would close that:

```yaml
- run: uv sync --group dev --extra claude-agent-sdk
```

The skip gate predates this PR, but this is the first change to put load-bearing tests under it, so it seems worth handling here.

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -444,24 +447,35 @@ def _resolve_tool_config(
         identifiers. We therefore refuse to forward a non-empty allowlist
         to the SDK rather than silently grant the wrong native tools.
 
+        The ``tools`` argument is the executor's *resolved* list from
+        :func:`conductor.executor.agent.resolve_agent_tools`. That function
+        erases the distinction between an omitted ``tools:`` and an explicit
+        ``tools: []``: both arrive here as an empty list whenever the
+        workflow declares no ``runtime`` MCP tools (the common case for this
+        provider, which rejects ``runtime.mcp_servers`` at the factory). We
```
jrob5756: This pins the empty list on MCP tools, but that isn't what drives it. The resolved list is `resolve_agent_tools(agent.tools, config.tools)`, so for an omitted `tools:` it comes back empty exactly when the workflow-level `tools:` (`config.tools`) is empty. `runtime.mcp_servers` never feeds into it, and it's rejected at the factory regardless. The `_check_agent_tools` branch you added at validator.py:1621 only exists because `config.tools` can be non-empty here, which is the opposite of the "common case ... rejects mcp_servers" framing. The inline comment at line 488 already says it correctly.

```suggestion
        workflow declares no workflow-level ``tools:`` (``config.tools`` is
        empty; a non-empty list makes an omitted agent resolve non-empty). We
```

examples/claude-agent-sdk-repo-qa.yaml
```diff
@@ -0,0 +1,77 @@
+# Repo Q&A with the claude-agent-sdk provider (default tools preset)
+#
+# This single-agent workflow asks the claude-agent-sdk provider to READ a file
+# in its working directory and answer a question about it. The agent does NOT
+# declare a `tools:` list — it relies on the provider's default behavior.
+#
+# WHY THIS EXAMPLE EXISTS (the bug it guards against):
+#   Before the fix, an agent that omitted `tools:` received NO tools and could
+#   not read the file: the executor's resolve_agent_tools() turned the omitted
+#   `tools:` into an empty list (the workflow had no `runtime` MCP tools to
```
jrob5756: Same point as the provider docstring: the list was empty because this example declares no workflow-level `tools:`, not because there were no MCP tools to inherit.

```suggestion
#   `tools:` into an empty list (this workflow declares no workflow-level `tools:` to
```

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -83,9 +83,12 @@ def _build_output_format(output: dict[str, OutputField]) -> dict[str, Any]:
...
 
-# Default tool preset granted when an agent does not declare a `tools:` list.
-# This mirrors the SDK's `claude_code` preset (filesystem, bash, web, etc.) —
-# i.e. the same behavior the user gets when running the `claude` CLI directly.
+# Default tool preset granted when an agent omits the `tools:` list. This
+# mirrors the SDK's `claude_code` preset (filesystem, bash, web, etc.) — i.e.
+# the same behavior the user gets when running the `claude` CLI directly. It is
+# selected from the RAW ``agent.tools is None`` signal, NOT from the executor's
+# resolved list: the executor erases the "omitted" distinction by returning an
+# empty list (the workflow tools copy) for an agent that declares no `tools:`.
```
jrob5756: Minor precision point: an omitted `tools:` only resolves to an empty list when the workflow declares no `tools:`. If `config.tools` is non-empty the agent inherits a non-empty copy and the provider raises, so the distinction isn't really erased. The parenthetical hints at this; the main clause overstates it.

```suggestion
# resolved list: for an agent that declares no `tools:`, the executor returns the
# workflow-tools copy, which is empty only when the workflow declares no `tools:`.
```

src/conductor/config/validator.py
```diff
@@ -1597,6 +1597,38 @@ def _caps_for(name: str) -> ProviderCapabilities | None:
...
+
+        Two failure modes against a non-passthrough provider:
+
+        * Explicit non-empty ``tools:`` — a declared allowlist the provider
+          cannot honor; silently granting different tools is a security
+          regression.
+        * Omitted ``tools:`` + non-empty workflow-level ``tools:`` — the agent
+          inherits that list at runtime (``resolve_agent_tools`` returns a copy)
+          and hits the same refusal mid-run with a confusing "declares
+          tools=[...]" error even though it declared none. An explicit
```
jrob5756: This quotes `declares tools=[...]`, but the PR renames the runtime error to `resolves to tools=[...]` (claude_agent_sdk.py:497). After the rename `declares tools=` only matches the validate-time message, so someone grepping for the mid-run path will land on the wrong line.

```suggestion
          and hits the same refusal mid-run (now a ``resolves to tools=[...]``
          ``ProviderError``) rather than failing fast at validate. An explicit
```

src/conductor/config/validator.py
```diff
@@ -1743,6 +1768,24 @@ def _caps_for(name: str) -> ProviderCapabilities | None:
...
 
+    # ----- For-each inline agents: tools capability cross-check -----
+    # A for_each group carries an INLINE ``AgentDef`` (not in ``config.agents``)
+    # that runs with ``workflow_tools=config.tools``, exactly like a top-level
+    # agent. The per-agent loop above skips it, so re-run the tools check here —
+    # otherwise an inline agent that omits ``tools:`` against a non-passthrough
+    # provider slips past ``validate`` and fails mid-iteration with the same
+    # confusing runtime error. (The other per-agent capability checks for inline
+    # agents are a separate, broader follow-up; this closes the tools footgun.)
+    for fe in config.for_each:
```
jrob5756, line 1779: Could this get a tracked issue rather than living as a code comment? This loop gives for_each inline agents the tools check but none of the other per-agent checks above (reasoning effort, output schema, MCP, session timeout). I confirmed an inline agent with `reasoning.effort: high` against this provider produces zero validator errors, while the same agent at top level errors out. That's the same silent-degradation class this PR is closing, just on a different axis. Pulling the per-agent capability block into a shared helper, the way you did with `_check_agent_tools`, would let both loops reuse it.

tests/test_providers/test_claude_agent_sdk.py
```diff
@@ -616,14 +620,145 @@ async def fake_query(**kwargs):
...
+            assert agent.tools == []
+            await provider.execute(agent=agent, context={}, rendered_prompt="hi", tools=[])
+
+        call_kwargs = options_mock.call_args[1]
+        assert call_kwargs["tools"] == []
+        assert call_kwargs["permission_mode"] is None
+
+    @patch("conductor.providers.claude_agent_sdk.CLAUDE_AGENT_SDK_AVAILABLE", True)
+    @patch("conductor.providers.claude_agent_sdk.ClaudeAgentOptions", Mock)
+    async def test_explicit_non_empty_tools_still_raises(self) -> None:
```
jrob5756, line 709: Two gaps around the reworded error:

1. This asserts the message but not the `suggestion`, which the PR also rewrote. Capturing the exception locks it in:

```python
with pytest.raises(ProviderError) as exc:
    ...
assert "tools: []" in exc.value.suggestion
```

2. Nothing covers the inheritance path (`agent.tools is None` plus a resolved non-empty list), which is the entire reason for the new "inherited from the workflow-level" wording. A case asserting that string would guard it.

jrob5756 (commented):
Thanks so much for contributing to Conductor! I took a pass over this one. Nice fix: it turns a silent zero-tools failure into a loud one at both validate and runtime, and the root-cause write-up in the provider is genuinely useful for the next person who trips over this.

The code itself checks out. I ran the suite (40 pass), validated the new example, and confirmed the footgun gets rejected at validate time and that a mixed-provider workflow only flags the offending agent.

My one real concern is that the regression tests for this fix don't actually run in CI (details inline). The rest are smaller test-coverage and comment notes.

tests/test_providers/test_claude_agent_sdk.py
lukeellison: Ah thanks for letting me know! I've changed that now, does the CI automatically rerun on push or do we need to re-trigger to test it?

src/conductor/config/validator.py
lukeellison: Yes definitely! Was considering that anyway but wasn't sure if you might suggest we tackle it as part of this :)
Here's the issue and I've updated the comment too.
https://github.com/microsoft/conductor/issues/270

tests/test_providers/test_claude_agent_sdk.py
jrob5756: Pushed in 84632f8:

- \`test_explicit_non_empty_tools_still_raises\` now captures the exception and asserts \`exc.value.suggestion\` (contains \`tools: []\`) alongside the message.
- New \`test_inherited_workflow_tools_raise_with_inheritance_wording\` covers the inheritance path (\`agent.tools is None\` + a non-empty resolved list) and asserts the new "inherited from the workflow-level" wording.

Both run (no longer skipped) now that the `claude-agent-sdk` extra is installed in CI.

jrob5756 (approved):
Thanks @lukeellison! All review feedback is addressed: the reworded refusal message **and** its suggestion are now pinned by tests, the inherited-tools path is covered, and the regression tests actually run in CI now that the `claude-agent-sdk` extra is installed. All checks green. 🚀

lukeellison:
Thanks for your help - happy to contribute! 😊

PR #280, opened 2026-07-06
[codex] Make env-var default hint copyable

AmberGuozx:
## Summary
- Fix the missing-environment-variable suggestion so it prints a copyable `${VAR:-default}` example.
- Add a regression assertion for the rendered suggestion.
- Leave environment variable resolution behavior unchanged.

## Test plan
- `uv run pytest tests/test_config/test_loader.py -q`
- `uv run ruff check src/conductor/config/loader.py tests/test_config/test_loader.py`


AmberGuozx:

PR #281, opened 2026-07-06
[codex] Reject non-finite duration values

AmberGuozx:
## Summary
- Reject `NaN` and infinity values in `parse_duration()`.
- Also reject overflowing numeric strings that parse to infinity.
- Add regression tests while preserving existing duration syntax.

## Test plan
- `uv run pytest tests/test_engine/test_duration.py -q`
- `uv run ruff check src/conductor/duration.py tests/test_engine/test_duration.py`


AmberGuozx:

PR #285, opened 2026-07-12
docs(cli): document --quiet/--silent as global options that precede the subcommand

decherneyge:
## Problem

`--quiet`/`-q` and `--silent`/`-s` are defined on the Typer **app callback** (`src/conductor/cli/app.py`), not on the `run` command, which makes them global options that must appear **before** the subcommand:

```bash
conductor --quiet run workflow.yaml     # works
conductor run workflow.yaml --quiet     # exit 2: "No such option: --quiet"
```

The documentation contradicted this in several places:

- `docs/cli-reference.md` listed both flags in the `conductor run` options table (under the `conductor run <workflow.yaml> [OPTIONS]` syntax line) and had two examples that fail as written — the `--quiet` debugging example and the CI/CD `--silent --log-file auto` pattern.
- `README.md` had the same table problem, and its flag descriptions were also inaccurate (`--silent` prints the JSON result, it doesn't "suppress all output except errors").
- The conductor skill's quick reference (`plugins/conductor/skills/conductor/SKILL.md`) showed `conductor run workflow.yaml -q ...`.

(`plugins/conductor/skills/conductor/references/execution.md` already documented the ordering correctly and was used as the template.)

## Changes

- Add a **Global Options** section to `docs/cli-reference.md` with correct/incorrect ordering examples; remove `--quiet`/`--silent` from the `run` options table and leave a pointer note; fix both broken examples.
- Fix the `README.md` run-options table the same way and show correctly ordered examples with accurate descriptions.
- Fix the mis-ordered examples in the conductor skill's quick reference.
- Add an `Unreleased > Documentation` CHANGELOG entry.

Docs-only change — no code touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   stays clickable for running items too.
   ([#273](https://github.com/microsoft/conductor/pull/273))
 
+### Documentation
+
+- **`--quiet` / `--silent` documented as `run` options** — both flags are
+  global options defined on the app callback, so they must appear *before* the
+  subcommand (`conductor --quiet run workflow.yaml`); placing them after it is
+  rejected with "No such option". `docs/cli-reference.md` and the README listed
+  them in the `conductor run` options table and showed post-subcommand examples
+  (`conductor run workflow.yaml --quiet`) that fail as written. Moved the flags
+  to a new Global Options section, corrected the examples, and fixed the same
+  mis-ordering in the conductor skill's quick reference.
+
 ## [0.1.20](https://github.com/microsoft/conductor/compare/v0.1.19...v0.1.20) - 2026-06-26
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,13 +73,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Documentation
 
 - **`--quiet` / `--silent` documented as `run` options** — both flags are
-  global options defined on the app callback, so they must appear *before* the
-  subcommand (`conductor --quiet run workflow.yaml`); placing them after it is
-  rejected with "No such option". `docs/cli-reference.md` and the README listed
-  them in the `conductor run` options table and showed post-subcommand examples
-  (`conductor run workflow.yaml --quiet`) that fail as written. Moved the flags
-  to a new Global Options section, corrected the examples, and fixed the same
-  mis-ordering in the conductor skill's quick reference.
+  root-level options defined on the app callback, so they must appear *before*
+  the subcommand (`conductor --quiet run workflow.yaml`); placing them after it
+  is rejected with "No such option". `docs/cli-reference.md` and the README
+  listed them in the `conductor run` options table, while `docs/cli-reference.md`
+  also showed post-subcommand examples (`conductor run workflow.yaml --quiet`)
+  that fail as written — as did the `conductor run --help` epilog. Moved the
+  flags to a new Root-Level Options section, corrected the examples (docs and
+  `--help` text), and fixed the same mis-ordering in the conductor skill's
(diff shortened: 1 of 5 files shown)


decherneyge:

docs/cli-reference.md
```diff
@@ -4,13 +4,39 @@ Complete command-line reference for Conductor.
...
+| `--version` | `-v` | Show version and exit |
+
+`--quiet` and `--silent` are mutually exclusive.
+
+```bash
+# Correct: global option before the subcommand
+conductor --quiet run workflow.yaml
+
+# Incorrect: rejected with "No such option: --quiet"
+conductor run workflow.yaml --quiet
```
jrob5756, line 38: Could this PR also update the example in `src/conductor/cli/app.py:409`? `conductor run --help` still tells users to run `conductor run workflow.yaml --silent --log-file auto`, which the new section correctly explains will fail.

docs/cli-reference.md
```diff
@@ -4,13 +4,39 @@ Complete command-line reference for Conductor.
...
 - [`conductor stop`](#conductor-stop)
 - [`conductor gate-respond`](#conductor-gate-respond)
 - [`conductor validate`](#conductor-validate)
 - [`conductor doctor`](#conductor-doctor)
 - [`conductor registry`](#conductor-registry)
 
+## Global Options
+
+The following options apply to every subcommand and **must appear before the
+subcommand name**:
```
jrob5756: These options are parsed at the root, but their output behavior does not apply to every command. `conductor --silent validate ...`, for example, still prints the normal Rich report. I would describe them as root-level options unless the intent is to make every command honor the verbosity setting.

```suggestion
The following root-level options **must appear before the
subcommand name**:
```

CHANGELOG.md
```diff
@@ -70,6 +70,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   ([#273](https://github.com/microsoft/conductor/pull/273))
 
+### Documentation
+
+- **`--quiet` / `--silent` documented as `run` options** — both flags are
+  global options defined on the app callback, so they must appear *before* the
+  subcommand (`conductor --quiet run workflow.yaml`); placing them after it is
+  rejected with "No such option". `docs/cli-reference.md` and the README listed
+  them in the `conductor run` options table and showed post-subcommand examples
+  (`conductor run workflow.yaml --quiet`) that fail as written. Moved the flags
```
jrob5756: This reads as though both documents contained broken examples. The README only listed the flags in the `run` options table; the failing examples were in the CLI reference. Could we separate those two claims?

```suggestion
  rejected with "No such option". `docs/cli-reference.md` and the README listed
  them in the `conductor run` options table, while `docs/cli-reference.md` also
  showed a post-subcommand example (`conductor run workflow.yaml --quiet`) that
  fails as written. Moved the flags
```

jrob5756 (commented):
Thank you for contributing to Conductor!  The placement fixes are right, but a few user-facing details still conflict with the CLI. Could you address those as well? Then we can merge! Thanks again!

decherneyge:
Thanks for the review, @jrob5756 — all three points addressed in 7084be0:

- `conductor run --help` epilog now shows `conductor --silent run workflow.yaml --log-file auto`; swept the rest of the CLI help text and that was the only misordered example.
- Reframed the section as **Root-Level Options** using your suggested wording, plus a note that the verbosity flags govern workflow progress output (`run`/`resume`) and may not affect other commands. Synced the terminology in the run-table note, the examples, the README, and the skill quick reference.
- CHANGELOG reworded per your suggestion so the failing examples are attributed to `docs/cli-reference.md` (and the `--help` epilog) — the README only had the options-table placement issue.

jrob5756 (approved):
All review feedback is addressed. The corrected help text matches the CLI behavior, the documentation now scopes the root-level options accurately, and CI is green.

jrob5756 (approved):
LGTM. Approved!

PR #291, opened 2026-07-13
feat(template): support Jinja include/import/extends for prompts loaded via !file

### LINKED ISSUE #287: Jinja prompt templates cannot use include/import/extends because TemplateRenderer uses BaseLoader
## Summary

Prompt templates loaded via `prompt: !file ...` are rendered as Jinja2 templates, but loader-dependent Jinja constructs such as `{% include %}`, `{% import %}`, and `{% extends %}` cannot work in normal CLI usage.

The renderer creates the Jinja environment with `loader=BaseLoader()` and renders file-backed prompts via `env.from_string(...)`. Because the YAML `!file` constructor inlines the file contents into the prompt string, the original template file path/directory is not available at render time, and there is no user-facing way to provide a Jinja search path.

This makes it impossible to factor prompt templates into reusable partials.

## Minimal Reproduction

Using Conductor `0.1.20`, Jinja2 `3.1.6`, Python `3.12`.

Example files:

```text
workflow.yaml
prompts/main.md.jinja
prompts/_shared.md.jinja
```

`workflow.yaml`:

```yaml
version: "1"
tasks:
  example:
    prompt: !file prompts/main.md.jinja
```

`prompts/main.md.jinja`:

```jinja
Before include.

{% include "_shared.md.jinja" %}

After include.
```

`prompts/_shared.md.jinja`:

```jinja
Shared prompt text.
```

Run the workflow with the normal Conductor CLI.

## Expected Behavior

The prompt should render successfully and include the contents of `_shared.md.jinja`, either relative to the prompt file directory or via a documented template search path.

## Actual Behavior

Rendering fails with a wrapped template error similar to:

```text
Template rendering failed: _shared.md.jinja
```

The underlying Jinja error is `jinja2.exceptions.TemplateNotFound`.

## Root Cause

In `src/conductor/executor/template.py`, `TemplateRenderer.__init__` creates the environment with a hard-coded `BaseLoader`:

```python
self.env = _DictSafeEnvironment(
    loader=BaseLoader(),
    undefined=StrictUndefined,
    autoescape=False,
    keep_trailing_newline=True,
)
```

`TemplateRenderer.render()` then compiles prompt text using `self.env.from_string(template)`.

Jinja's `BaseLoader` cannot resolve template names for `{% include %}`, `{% import %}`, or `{% extends %}`. Those constructs require a real loader such as `FileSystemLoader`, `PackageLoader`, or a custom loader.

Separately, `src/conductor/config/loader.py` implements the `!file` YAML constructor by reading the referenced file and inlining its raw contents into the schema field. That means the prompt's source path/directory is discarded before rendering, so the renderer has no template origin from which to resolve relative includes.

## Suggested Direction

Possible fixes:

1. Preserve source metadata for `!file` prompt values so the renderer knows the prompt file directory.
2. Configure a Jinja loader for file-backed prompt templates, such as a `FileSystemLoader` rooted at the prompt directory, workflow directory, or an explicitly configured template search path.
3. Document the resolution rules for `{% include %}`, `{% import %}`, and `{% extends %}` in prompt templates.
4. If loader-dependent Jinja features are intentionally unsupported, fail earlier with a clearer error message and document that limitation.

## Workaround

The only practical workaround I found is to manually inline shared prompt text into each prompt file instead of using Jinja includes.

## Environment

- Conductor: `0.1.20`
- Jinja2: `3.1.6`
- Python: `3.12`

hertznsk:
## Summary

Closes #287.

Prompt templates loaded via `prompt: !file ...` (and `system_prompt: !file ...`) now support loader-dependent Jinja constructs — `{% include %}`, `{% import %}`, and `{% extends %}` — resolved relative to the prompt file's own directory. This enables reusable prompt partials across workflows.

The fix follows the suggested direction from the issue (options 1 + 2 + 3 + 4 combined): preserve the source path of `!file`-loaded prompts, give the Jinja renderer a `FileSystemLoader` rooted at that path, document the resolution rules, and produce clear error messages for both missing includes and inline prompts that attempt loader-dependent constructs.

## Changes

- **`src/conductor/file_string.py`** (new leaf module): `FileString(str)` subclass carrying `source_path: Path`; `__getnewargs__` keeps it safe across `copy.deepcopy` / `pickle` / `model_copy(deep=True)`.
- **`src/conductor/config/loader.py`**: `!file` raw-string content is returned as `FileString` (dict/list parsing unchanged); `_resolve_env_vars_recursive()` preserves `FileString` metadata through `${VAR}` resolution.
- **`src/conductor/config/schema.py`**: Pydantic wrap-validators on `AgentDef.prompt` / `system_prompt` preserve the `FileString` subclass through validation; public annotations unchanged; `model_dump(mode="json")` still yields plain strings. Other `!file` fields (`command`, `stdin`, `reason`, `value`, ...) intentionally keep coercing to plain `str`.
- **`src/conductor/executor/template.py`**: `TemplateRenderer.render()` branches on `isinstance(FileString)` — file-backed prompts render in a `_DictSafeEnvironment` with `FileSystemLoader(source_path.parent)` via `env.compile(..., filename=...)` + `from_code(...)`; inline prompts use the unchanged `BaseLoader` + `from_string` path. `TemplateNotFound` errors now name the searched directory (`Template not found: '<name>'. Searched in: <dir>`), and inline prompts with `{% include %}` get an explicit error explaining that loader-dependent constructs require `prompt: !file ...`.
- **`docs/workflow-syntax.md`**: new "Jinja Includes in Prompt Files" subsection with example, resolution rules, and the documented limitation that `conductor validate` does not scan inside included files.
- **Tests**: 36 new tests — unit (FileString, loader, env preservation, Pydantic, renderer) and end-to-end integration covering `prompt`, `system_prompt`, `for_each` inline agents, and `human_gate` (shared-renderer side effect), plus missing-include failure path.

## Design decisions (scope guardrails)

- Relative includes resolve **only** against the prompt file's own directory — no workflow-directory fallback, no user-configurable template search path (can be added later if needed).
- Inline prompts behave exactly as before; all changes are gated behind `isinstance(FileString)`.
- `human_gate` prompts loaded via `!file` gain include support as a natural side effect of the shared renderer.

## Test plan

- `uv run pytest -m "not install_scripts" tests/` — 3857 passed (1 pre-existing env-only `real_api` failure reproduced identically on `main`, unrelated).
- `make check` (ruff lint + format + ty) — exit 0.
- Manual QA: extracted the docs example into fixtures and validated with `uv run conductor validate` (exit 0); verified rendered prompt/system_prompt reach the provider with partials inlined; verified both new error messages.

### DIFF
--- a/src/conductor/executor/template.py
+++ b/src/conductor/executor/template.py
@@ -9,10 +9,11 @@ from __future__ import annotations
 import json
 from typing import Any
 
-from jinja2 import BaseLoader, Environment, StrictUndefined, TemplateSyntaxError
+from jinja2 import BaseLoader, Environment, FileSystemLoader, StrictUndefined, TemplateSyntaxError
 from jinja2 import UndefinedError as Jinja2UndefinedError
 
 from conductor.exceptions import TemplateError
+from conductor.file_string import FileString
 
 
 class _DictSafeEnvironment(Environment):
@@ -63,8 +64,12 @@ class TemplateRenderer:
         )
 
         # Register custom filters
-        self.env.filters["json"] = self._json_filter
-        self.env.filters["default"] = self._default_filter
+        self._register_filters(self.env)
+
+    def _register_filters(self, env: Environment) -> None:
+        """Register custom filters on the Jinja2 environment."""
+        env.filters["json"] = self._json_filter
+        env.filters["default"] = self._default_filter
 
     @staticmethod
     def _json_filter(value: Any, indent: int = 2) -> str:
@@ -108,7 +113,18 @@ class TemplateRenderer:
             TemplateError: If rendering fails due to missing variables or syntax errors.
         """
         try:
-            tmpl = self.env.from_string(template)
+            if isinstance(template, FileString) and template.source_path.exists():
+                env = _DictSafeEnvironment(
+                    loader=FileSystemLoader(str(template.source_path.parent)),
+                    undefined=StrictUndefined,
+                    autoescape=False,
+                    keep_trailing_newline=True,
+                )
+                self._register_filters(env)
+                code = env.compile(str(template), filename=str(template.source_path))
+                tmpl = env.template_class.from_code(env, code, env.globals)
+            else:
+                tmpl = self.env.from_string(template)
             return tmpl.render(**context)
         except Jinja2UndefinedError as e:
             # Extract the variable name from the error message
--- a/src/conductor/executor/template.py
+++ b/src/conductor/executor/template.py
@@ -19,7 +19,8 @@ from jinja2 import (
 )
 from jinja2 import UndefinedError as Jinja2UndefinedError
 
-from conductor.exceptions import TemplateError
+from conductor.config.loader import resolve_env_vars
+from conductor.exceptions import ConfigurationError, TemplateError
 from conductor.file_string import FileString
 
 
@@ -45,6 +46,29 @@ class _DictSafeEnvironment(Environment):
                 pass
         # Fall back to the default resolution (getattr → getitem → undefined)
         return super().getattr(obj, attribute)
+
+
+class _EnvResolvingFileSystemLoader(FileSystemLoader):
+    """FileSystemLoader that resolves ``${VAR}`` env placeholders in partials.
+
+    The config loader resolves environment variables in the *root* ``!file``
+    prompt at YAML load time, but Jinja loads ``{% include %}`` /
+    ``{% import %}`` / ``{% extends %}`` targets itself at render time — long
+    after config loading finished.  Without this hook, ``${VAR}`` text inside
+    a partial would reach the model literally, and an unset required variable
+    would silently pass through instead of raising the usual configuration
+    error.
+
+    Each partial source is therefore run through the same
+    :func:`~conductor.config.loader.resolve_env_vars` resolver used by the
+    config loader before Jinja compiles it, so partials behave identically
+    to the root prompt file.
+    """
+
+    def get_source(self, environment: Environment, template: str) -> tuple[str, str, Any]:
+        """Load *template* and resolve env vars in its source before compile."""
+        source, filename, uptodate = super().get_source(environment, template)
+        return resolve_env_vars(source), filename, uptodate
 
 
 class TemplateRenderer:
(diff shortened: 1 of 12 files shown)


hertznsk:

src/conductor/executor/template.py
```diff
@@ -107,9 +119,38 @@ def render(self, template: str, context: dict[str, Any]) -> str:
         Raises:
             TemplateError: If rendering fails due to missing variables or syntax errors.
         """
+        loader = None
+        is_file_backed = False
         try:
-            tmpl = self.env.from_string(template)
+            if isinstance(template, FileString) and template.source_path.exists():
+                is_file_backed = True
+                loader = FileSystemLoader(str(template.source_path.parent))
```
jrob5756: Jinja loads partials after the config loader has finished `${VAR}` expansion. Variables in an included, imported, or base template therefore stay literal, while the same text in the main `!file` prompt is expanded. An unset required variable also stops producing the normal configuration error. Please run each template source through the existing environment resolver before Jinja compiles it, and add an end-to-end partial test.

src/conductor/executor/template.py
```diff
@@ -107,9 +119,38 @@ def render(self, template: str, context: dict[str, Any]) -> str:
         Raises:
             TemplateError: If rendering fails due to missing variables or syntax errors.
         """
+        loader = None
+        is_file_backed = False
         try:
-            tmpl = self.env.from_string(template)
+            if isinstance(template, FileString) and template.source_path.exists():
```
jrob5756: If the original file is removed or inaccessible after loading, this condition quietly treats the `FileString` as an inline prompt. Includes then fail with a message telling the user to switch to `!file`, even though they already did. Please report the unavailable `source_path` directly instead of changing rendering modes.

jrob5756 (commented):
Another great PR. Thanks @hertznsk ! I found two behavior issues in the file-backed rendering path. Both are worth fixing before merge because they can leave placeholders unresolved or turn a missing file into a misleading inline-template error.

src/conductor/executor/template.py
hertznsk: Fixed in b80e95e.

Each partial source now goes through the config loader's own env resolver before Jinja compiles it: `_EnvResolvingFileSystemLoader(FileSystemLoader).get_source()` runs `resolve_env_vars(source)` on the text of every `{% include %}` / `{% import %}` / `{% extends %}` target (nested and dynamic includes included), so `${VAR}` / `${VAR:-default}` in partials behaves identically to the root `!file` prompt. An unset required variable propagates as the original `ConfigurationError` (dedicated `except ConfigurationError: raise` ahead of the template-error handlers), preserving its message and suggestion.

Tests: unit coverage in `tests/test_executor/test_template_filestring.py` (render-time resolution, default fallback, unset-required `ConfigurationError`) plus an end-to-end test in `tests/test_integration/test_file_prompt_includes.py` that sets the env var *after* config load to prove resolution happens at render time.

src/conductor/executor/template.py
hertznsk: Fixed in b80e95e.

The file-backed branch no longer degrades silently: `TemplateRenderer.render()` now checks `template.source_path.is_file()` for every `FileString` *before* entering the try block, and raises immediately:

`TemplateError: File-backed prompt source is no longer available: '<path>' (loaded via !file).`

with a suggestion to restore the file — no inline fallback, no misleading "convert to `prompt: !file`" guidance, and no double-wrapping by the generic handler (the tests assert the exact message and `__cause__ is None`). Note the check uses `is_file()`, so a directory or broken symlink at that path is reported the same way.

Tests: `test_missing_source_path_raises_explicit_error` (unit) and `test_deleted_prompt_source_fails_with_explicit_error` (end-to-end: load config, delete the prompt file, run the engine, assert the error names the missing path).

jrob5756 (approved):
lgtm. Thanks!

PR #292, opened 2026-07-13
fix(template): support boolean argument in Jinja default filter

### LINKED ISSUE #288: Custom default filter breaks Jinja's default(value, default_value, boolean) calling convention
## Summary

Conductor overrides Jinja2's built-in `default` filter with a custom implementation that only accepts two parameters:

```python
def _default_filter(value: Any, default: Any = "") -> Any:
```

This breaks valid Jinja syntax that uses the standard third `boolean` argument:

```jinja
{{ value | default("unknown", true) }}
```

At render time, this raises a Python `TypeError` because Jinja passes three positional arguments to the custom filter.

## Minimal Reproduction

Using Conductor `0.1.20`, Jinja2 `3.1.6`, Python `3.12`.

Example prompt template:

```jinja
Value: {{ value | default("unknown", true) }}
```

Render it through a normal Conductor workflow where `value` is missing, `None`, or otherwise intended to fall back to `"unknown"`.

## Expected Behavior

The template should render successfully.

Since `default` is a standard Jinja filter, templates should be able to use its documented calling convention:

```jinja
{{ value | default("unknown", true) }}
```

## Actual Behavior

Rendering fails with an error similar to:

```text
Template rendering failed: _default_filter() takes from 1 to 2 positional arguments but 3 were given
```

## Root Cause

In `src/conductor/executor/template.py`, `TemplateRenderer.__init__` replaces Jinja's built-in `default` filter:

```python
self.env.filters["default"] = self._default_filter
```

The custom implementation only accepts `value` and `default`:

```python
def _default_filter(value: Any, default: Any = "") -> Any:
    if value is None:
        return default
    return value
```

Jinja2's built-in `default` filter has the calling convention:

```python
default(value, default_value="", boolean=False)
```

So valid templates using the third `boolean` argument fail when rendered by Conductor.

The intent of Conductor's override appears reasonable: treat `None` as missing while preserving empty strings. For example, `tests/test_executor/test_template.py` includes coverage for preserving an empty string. However, the override does not preserve the built-in filter's signature.

It may also change standard behavior for undefined values under `StrictUndefined`, because Jinja's built-in `default` handles `Undefined` values specially.

## Suggested Direction

Make the custom filter signature-compatible with Jinja's built-in `default` filter:

```python
def _default_filter(value: Any, default_value: Any = "", boolean: bool = False) -> Any:
    ...
```

Suggested semantics:

1. Return `default_value` for Jinja `Undefined` values.
2. Return `default_value` when `value is None`.
3. Preserve Conductor's current behavior for empty strings when `boolean` is omitted or false.
4. When `boolean=True`, follow Jinja's standard behavior for falsey values, or document any intentional deviation.

This would preserve the existing Conductor behavior for `None` and `""` while avoiding runtime failures for valid Jinja templates that pass the third argument.

## Workaround

Avoid the third argument to `default`, or rewrite templates to use explicit conditionals instead of Jinja's standard `default(..., true)` form.

## Environment

- Conductor: `0.1.20`
- Jinja2: `3.1.6`
- Python: `3.12`

hertznsk:
## Summary

Fixes #288.

Conductor overrides Jinja2's built-in `default` filter with a custom implementation that previously only accepted two parameters. Valid Jinja templates using the standard third `boolean` argument raised `TypeError: _default_filter() takes from 1 to 2 positional arguments but 3 were given`.

This PR makes `TemplateRenderer._default_filter` signature-compatible with Jinja's built-in `default(value, default_value="", boolean=False)`.

## What changed

**`src/conductor/executor/template.py`** — `_default_filter` now accepts `boolean: bool = False`:

- `boolean=False` (default): preserves existing Conductor behavior — only `None` triggers the default; empty strings, `0`, `False`, `[]`, `{}` are returned as-is.
- `boolean=True`: standard Jinja falsey semantics — any falsey value (`None`, `""`, `0`, `False`, `[]`, `{}`) returns the default; truthy values pass through unchanged.

The parameter name `default` is kept (not renamed to `default_value`) to avoid breaking any existing keyword-argument callers.

## Tests

**`tests/test_executor/test_template.py`** — 14 new tests under `TestTemplateRendererDefaultFilter`:

- `boolean=True` with each falsey value (`None`, `""`, `0`, `False`, `[]`, `{}`) → default
- `boolean=True` with a truthy value → value returned unchanged
- Two-arg form regression: `None` → default; `""`, `0`, `False`, `[]`, `{}` → returned as-is
- Explicit `boolean=False` regression for the same edge cases

## Verification

- `uv run pytest tests/test_executor/test_template.py -v` → **60/60 passed**
- `make check` (ruff check + ruff format + ty) → **exit 0**
- Manual: `uv run python -c 'from conductor.executor.template import TemplateRenderer; r = TemplateRenderer(); print(r.render("{{ value | default(\\"x\\", true) }}", {"value": None}))'` → prints `x` without `TypeError`

## Out of scope

Per the issue's suggested direction, item 1 ("Return default_value for Jinja `Undefined` values") is intentionally not changed here: Conductor runs with `StrictUndefined`, so undefined variables raise `TemplateError` before reaching the filter — the existing fail-fast behavior is preserved.

### DIFF
--- a/src/conductor/executor/template.py
+++ b/src/conductor/executor/template.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 import json
 from typing import Any
 
-from jinja2 import BaseLoader, Environment, StrictUndefined, TemplateSyntaxError
+from jinja2 import BaseLoader, Environment, StrictUndefined, TemplateSyntaxError, Undefined
 from jinja2 import UndefinedError as Jinja2UndefinedError
 
 from conductor.exceptions import TemplateError
@@ -85,18 +85,20 @@ class TemplateRenderer:
 
         Args:
             value: The value to check.
-            default: Default value to return if value is None or falsey.
+            default: Value returned for None, Undefined, or for any falsey value
+                when boolean=True.
             boolean: If True, return default for any falsey value.
 
         Returns:
-            The value if not None (or not falsey if boolean=True), otherwise the default.
+            The value if not None/Undefined (or not falsey if boolean=True),
+            otherwise the default.
         """
-        if boolean:
-            if not value:
-                return default
-        else:
-            if value is None:
-                return default
+        if isinstance(value, Undefined):
+            return default
+        if boolean and not value:
+            return default
+        if value is None:
+            return default
         return value
 
     def render(self, template: str, context: dict[str, Any]) -> str:
--- a/src/conductor/executor/template.py
+++ b/src/conductor/executor/template.py
@@ -80,18 +80,23 @@ class TemplateRenderer:
         return json.dumps(value, indent=indent, default=str)
 
     @staticmethod
-    def _default_filter(value: Any, default: Any = "") -> Any:
+    def _default_filter(value: Any, default: Any = "", boolean: bool = False) -> Any:
         """Return default if value is None or undefined.
 
         Args:
             value: The value to check.
-            default: Default value to return if value is None.
+            default: Default value to return if value is None or falsey.
+            boolean: If True, return default for any falsey value.
 
         Returns:
-            The value if not None, otherwise the default.
+            The value if not None (or not falsey if boolean=True), otherwise the default.
         """
-        if value is None:
-            return default
+        if boolean:
+            if not value:
+                return default
+        else:
+            if value is None:
+                return default
         return value
 
     def render(self, template: str, context: dict[str, Any]) -> str:
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -3998,8 +3998,8 @@ class WorkflowEngine:
             raise
 
     # Type-appropriate zero values for optional inputs with no declared default.
-    # Using None causes templates to render "None" instead of empty string,
-    # and | default() won't catch None without the boolean=true flag.
+    # Using None causes templates to render the literal "None" instead of a clean
+    # value, so we substitute a type-appropriate zero value up front.
     # Note: mutable types (array, object) return fresh copies via the method below.
     _TYPE_ZERO_VALUES: dict[str, Any] = {
         "string": "",
--- a/tests/test_executor/test_template.py
+++ b/tests/test_executor/test_template.py
@@ -263,6 +263,55 @@ class TestTemplateRendererDefaultFilter:
         )
         assert result == "Value: {}"
 
+    def test_default_filter_missing_variable_returns_default(self) -> None:
+        """Test default filter returns fallback for an undefined variable.
+
+        Requirements:
+        - A missing variable guarded by default() must render the fallback instead
+          of raising TemplateError under StrictUndefined.
+        """
+        renderer = TemplateRenderer()
+        result = renderer.render(
+            "Value: {{ missing | default('fallback') }}",
+            {},
+        )
+        assert result == "Value: fallback"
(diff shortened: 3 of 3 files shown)


src/conductor/executor/template.py
```diff
@@ -80,18 +80,23 @@ def _json_filter(value: Any, indent: int = 2) -> str:
...
         """
-        if value is None:
-            return default
+        if boolean:
+            if not value:
+                return default
+        else:
+            if value is None:
+                return default
         return value
```
jrob5756: This still breaks the missing-value case from #288. With `StrictUndefined`, `not value` calls `StrictUndefined.__bool__` and raises, while Jinja's built-in `default` returns the fallback. Please check `Undefined` before testing truthiness and add coverage for a missing variable in both forms. This replacement also needs `Undefined` imported from `jinja2`.

```suggestion
if isinstance(value, Undefined):
    return default
if boolean and not value:
    return default
if value is None:
    return default
return value
```

src/conductor/executor/template.py
```diff
@@ -80,18 +80,23 @@ def _json_filter(value: Any, indent: int = 2) -> str:
...
 
     @staticmethod
-    def _default_filter(value: Any, default: Any = "") -> Any:
+    def _default_filter(value: Any, default: Any = "", boolean: bool = False) -> Any:
         """Return default if value is None or undefined.
 
         Args:
             value: The value to check.
-            default: Default value to return if value is None.
+            default: Default value to return if value is None or falsey.
```
jrob5756: This reads as though every falsey value uses the fallback. Empty strings, zero, `False`, and empty collections are preserved unless `boolean=True`.

```suggestion
            default: Value returned for None, or for any falsey value when boolean=True.
```

jrob5756 (commented):
Good change. Some small things, the new argument works for defined falsey values, but the missing-variable path still differs from Jinja and needs a regression test. I also left a small docstring clarification.

If you can wrap these up, we can merge! Thanks again.

src/conductor/executor/template.py
hertznsk: Fixed in 0ab6c6c. `Undefined` is now imported from `jinja2` and checked first, before any truthiness test, so `StrictUndefined.__bool__` is never reached:

```python
if isinstance(value, Undefined):
    return default
if boolean and not value:
    return default
if value is None:
    return default
return value
```

Both `{{ v | default('x') }}` and `{{ v | default('x', true) }}` now return the fallback for a missing variable, while unguarded `{{ missing }}` still raises under `StrictUndefined` (fail-fast preserved). Added regression tests for a missing top-level variable in both the two-arg and `boolean=True` forms (including the positional `true` shorthand), plus a missing nested dict key.

src/conductor/executor/template.py
hertznsk: Reworded in 0ab6c6c to make clear that empty strings, zero, `False`, and empty collections are preserved unless `boolean=True`:

```python
default: Value returned for None, Undefined, or for any falsey value
    when boolean=True.
```

I used your suggested phrasing and added `Undefined` since the filter now also returns the fallback for undefined values.

jrob5756 (approved):
lgtm. approved

PR #293, opened 2026-07-13
fix(providers): pass agent system_prompt as native system parameter in Claude provider

### LINKED ISSUE #289: Claude provider silently drops agent.system_prompt during agent execution
## Проблема

Native `claude` provider (`src/conductor/providers/claude.py`, Anthropic Messages API) **никогда не передаёт `agent.system_prompt`** при выполнении агентов. Поле `system_prompt` есть в YAML-схеме, рендерится `AgentExecutor` с полным workflow-контекстом, но в `ClaudeProvider.execute()` не попадает ни в `messages`, ни в top-level параметр `system` Anthropic API.

Это silently меняет поведение workflows: инструкции, persona, methodology и статические role constraints из `system_prompt:` полностью игнорируются при `runtime.provider: claude`, хотя тот же workflow на других провайдерах работает корректно.

## Evidence

- `AgentDef.system_prompt` объявлен в schema: `src/conductor/config/schema.py:722`.
- `AgentExecutor.execute()` рендерит `agent.system_prompt` с workflow context и обновляет копию agent: `src/conductor/executor/agent.py:269-278`.
- `ClaudeProvider.execute()` строит initial messages только из user prompt: `src/conductor/providers/claude.py:956-957`.
- `_build_messages()` возвращает ровно один user message (`[{"role": "user", "content": rendered_prompt}]`): `src/conductor/providers/claude.py:2083-2097`.
- `_execute_api_call()` формирует Anthropic kwargs без `system` (только `model`, `messages`, `max_tokens`, опционально `temperature`, `tools`, `thinking`): `src/conductor/providers/claude.py:1350-1364`.
- `grep agent.system_prompt` по `claude.py` — 0 совпадений. Единственный путь, передающий `system`, — `execute_dialog_turn()` (`"system": system_prompt`), но это отдельный lightweight dialog path для human-gate диалогов, не основной agent execution path: `src/conductor/providers/claude.py:601-647`.
- Ни `_execute_agentic_loop`, ни `_execute_with_parse_recovery`, ни `_request_partial_output`, ни retry-обёртка не инжектят system-контент ни в messages, ни в kwargs.
- Validator already documents this as a known portability hazard (warns that providers "e.g., Claude" ignore `system_prompt` entirely): `src/conductor/config/validator.py:184-204`.

## Provider parity

Native `claude` — единственный провайдер, который silently дропает `system_prompt`:

- `claude_agent_sdk`: native `ClaudeAgentOptions(system_prompt=agent.system_prompt)` — `src/conductor/providers/claude_agent_sdk.py:224-227`.
- `hermes`: native `system_message=agent.system_prompt or None` — `src/conductor/providers/hermes.py:324-329`.
- `copilot`: конкатенирует `f"System: {system_prompt}\n\nUser: {rendered_prompt}"` в один prompt — `src/conductor/providers/copilot.py:769-772` (семантически передаёт, хотя и не как system role).

## Impact

- Workflows, использующие `system_prompt:`, ведут себя по-разному на `claude` и на других провайдерах — нарушение provider parity (AGENTS.md).
- Статические инструкции silently игнорируются → потенциально более низкое качество или небезопасные ответы.
- Prompt-caching оптимизация стабильного system-префикса принципиально невозможна: нет top-level `system` блока, который можно было бы закэшировать (Anthropic cache prefix order: `tools` → `system` → `messages`). Это будет предметом отдельного follow-up issue.
- Validator warning, утверждающий, что Claude игнорирует `system_prompt`, должен быть обновлён после фикса.

## Proposed fix

Передавать rendered `agent.system_prompt` как native Anthropic Messages API параметр `system` во всех API-вызовах agent execution path:

- В `ClaudeProvider.execute()` resolve `system_prompt = agent.system_prompt or None`.
- Протянуть значение через `_execute_agentic_loop`, `_execute_with_parse_recovery`, `_request_partial_output` → `_execute_api_call`.
- В `_execute_api_call()` добавлять `kwargs["system"] = system_prompt` только когда значение непустое.
- НЕ инжектить `system_prompt` в `messages` — использовать native top-level `system` параметр.
- Prompt caching (`cache_control` breakpoints) — вне scope этого fix.

## Acceptance criteria

1. Unit-тест: агент с `system_prompt: "..."` → `AsyncAnthropic.messages.create()` получает `system="..."`.
2. Тест: templated `system_prompt` (с `{{ ... }}`) рендерится `AgentExecutor` до передачи в provider.
3. Tool-use iterations, parse recovery calls и interrupt partial-output calls сохраняют тот же `system` параметр.
4. Агенты без `system_prompt` сохраняют текущую форму запроса: пустой `system` kwarg не отправляется.
5. `config/validator.py` warning больше не утверждает, что Claude игнорирует `system_prompt`; связанные validator-тесты обновлены.
6. Provider parity docs/tests обновлены, если упоминают, что Claude игнорирует `system_prompt`.

## Out of scope (follow-ups)

- Anthropic prompt caching config (`cache_control`, TTL, automatic vs explicit breakpoints, пользовательский API типа `runtime.prompt_caching`).
- Перенос workspace-instructions preamble из user prompt в system block (cross-provider behavioral change).
- Переключение Copilot agent execution на native SDK system-message support (сейчас — текстовая конкатенация).
- Агрегация `cache_read_tokens` / `cache_write_tokens` по всем turn'ам agentic loop (сейчас — только final response).

hertznsk:
## Summary
Fixes #289. The native `claude` provider (Anthropic Messages API) silently dropped `agent.system_prompt`. It is now passed as the native top-level `system` parameter of `messages.create()` on every API call in the agent execution path: main loop, tool-use iterations, parse recovery, interrupt partial output, and retries. Empty/whitespace system prompts are normalized away and not sent.

Secondary fix: the `OutputValidator` rubric (synthetic agent with `prompt=""` and a formatted `system_prompt`) now reaches the model on the claude provider through the same native `system` parameter.

## Changes
- **Provider** (`src/conductor/providers/claude.py`): thread a keyword-with-default `system_prompt: str | None = None` (appended at the end of each private signature, mirroring the existing `thinking` precedent) through `_execute_with_retry` → `_execute_agentic_loop` → `_request_partial_output` / `_execute_with_parse_recovery` → `_execute_api_call`, where `kwargs["system"]` is set only when non-empty after strip-normalization. Public `AgentProvider.execute` signature unchanged.
- **Validator warning** (`src/conductor/config/validator.py`): the system_prompt-without-prompt warning no longer claims Claude ignores `system_prompt`; rationale is now provider-neutral (empty user message on every provider).
- **Tests**: new `tests/test_providers/test_claude_system_prompt.py` (8 tests — system kwarg on plain execute / every tool-loop call / parse-recovery calls / interrupt partial-output call; absent for None/empty/whitespace; validator rubric arrives formatted, not raw) + negative assertions in `tests/test_config/test_validator.py`.
- **Docs**: new "System Prompt" section in `docs/providers/claude.md`; wording fix in the conductor skill's execution reference.

## Testing
- New test file: 8 passed.
- Full suite: 3803 passed (one deselected integration test hits the live GitHub Copilot API and fails in this sandbox due to org policy — it fails identically on unmodified `main`).
- `ruff check` clean; `conductor validate` shows the rewritten warning on a real workflow.

## Notes
Behavior change (intended): claude agents that define `system_prompt` now actually receive those instructions; previously they were silently dropped.

### DIFF
--- a/plugins/conductor/skills/conductor/references/execution.md
+++ b/plugins/conductor/skills/conductor/references/execution.md
@@ -221,7 +221,7 @@ Performs both schema and **semantic** checks:
 - Parallel group agent references
 - For-each `source` format and reserved names
 - Stale agent references and undeclared explicit-mode dependencies in `prompt`, `system_prompt`, `command`, `args`, `working_dir`, `input_mapping`, parallel-group inputs, and workflow `output:` templates
-- Warning when an agent defines `system_prompt` but no `prompt:` (unusual — the user message would be empty)
+- Warning when an agent defines `system_prompt` but no `prompt:` (unusual because the user-authored task prompt would be empty)
 
 The success summary table includes Parallel Groups and For-each Groups counts.
 
--- a/plugins/conductor/skills/conductor/references/execution.md
+++ b/plugins/conductor/skills/conductor/references/execution.md
@@ -221,7 +221,7 @@ Performs both schema and **semantic** checks:
 - Parallel group agent references
 - For-each `source` format and reserved names
 - Stale agent references and undeclared explicit-mode dependencies in `prompt`, `system_prompt`, `command`, `args`, `working_dir`, `input_mapping`, parallel-group inputs, and workflow `output:` templates
-- Warning when an agent defines `system_prompt` but no `prompt:` (unusual — system prompts are only meaningful alongside a user prompt)
+- Warning when an agent defines `system_prompt` but no `prompt:` (unusual — the user message would be empty)
 
 The success summary table includes Parallel Groups and For-each Groups counts.
 
--- a/src/conductor/config/validator.py
+++ b/src/conductor/config/validator.py
@@ -182,10 +182,9 @@ def validate_workflow_config(
             errors.extend(tool_errors)
 
         # Warn when an LLM agent has system_prompt but no (non-empty) prompt.
-        # Even when system_prompt is rendered correctly, omitting `prompt:` sends
-        # an empty user message to the model. That is a provider-neutral authoring
-        # hazard and almost always means dynamic, must-execute content belongs in
-        # `prompt:` alongside the persona/methodology in `system_prompt:`.
+        # Omitting `prompt:` leaves the user-authored task prompt empty, which
+        # almost always means dynamic, must-execute content belongs in `prompt:`
+        # alongside the persona/methodology in `system_prompt:`.
         if (
             agent.type in (None, "agent")
             and agent.system_prompt
@@ -194,8 +193,8 @@ def validate_workflow_config(
             warnings.append(
                 f"Agent '{agent.name}' defines `system_prompt` but no `prompt` "
                 "(or only whitespace). "
-                "This sends an empty user message to the model on every provider, "
-                "which almost always indicates a latent authoring mistake. Move "
+                "The user-authored task prompt is empty, which almost always "
+                "indicates a latent authoring mistake. Move "
                 "the dynamic, must-execute content (input references, instructions) "
                 "into a `prompt:` block; keep the persona and static methodology "
                 "in `system_prompt:`."
--- a/src/conductor/config/validator.py
+++ b/src/conductor/config/validator.py
@@ -182,12 +182,10 @@ def validate_workflow_config(
             errors.extend(tool_errors)
 
         # Warn when an LLM agent has system_prompt but no (non-empty) prompt.
-        # The Copilot provider concatenates `agent.system_prompt` into the prompt,
-        # but providers that ignore system_prompt entirely (e.g., Claude) would
-        # leave such agents with an empty user message. Even with the executor
-        # rendering system_prompt, omitting `prompt:` is a portability hazard
-        # and almost always indicates the author meant to include a `prompt:`
-        # block alongside the persona/methodology in `system_prompt:`.
+        # Even when system_prompt is rendered correctly, omitting `prompt:` sends
+        # an empty user message to the model. That is a provider-neutral authoring
+        # hazard and almost always means dynamic, must-execute content belongs in
+        # `prompt:` alongside the persona/methodology in `system_prompt:`.
         if (
             agent.type in (None, "agent")
             and agent.system_prompt
@@ -196,12 +194,11 @@ def validate_workflow_config(
             warnings.append(
                 f"Agent '{agent.name}' defines `system_prompt` but no `prompt` "
                 "(or only whitespace). "
-                "Some providers (e.g., Claude) ignore `system_prompt` entirely, "
-                "which would send an empty user message to the model. "
-                "Even with the Copilot provider, the model often responds poorly "
-                "to a missing user prompt. Move the dynamic, must-execute content "
-                "(input references, instructions) into a `prompt:` block; keep the "
-                "persona and static methodology in `system_prompt:`."
+                "This sends an empty user message to the model on every provider, "
+                "which almost always indicates a latent authoring mistake. Move "
+                "the dynamic, must-execute content (input references, instructions) "
+                "into a `prompt:` block; keep the persona and static methodology "
+                "in `system_prompt:`."
             )
 
     # Validate parallel groups
--- a/src/conductor/providers/claude.py
+++ b/src/conductor/providers/claude.py
@@ -977,7 +977,13 @@ class ClaudeProvider(AgentProvider):
         # Done before the per-model max_tokens warning so the warning logic
         # accounts for thinking-aware caps.
         thinking = self._resolve_thinking_for_agent(agent, model)
-        system_prompt = (agent.system_prompt or "").strip() or None
+        # Use strip() only as a blank-prompt predicate: the rendered value is
+        # forwarded verbatim for cross-provider parity (Copilot, Claude Agent
+        # SDK, and Hermes all pass agent.system_prompt through unchanged).
+        raw_system_prompt = agent.system_prompt
+        system_prompt = (
+            raw_system_prompt if raw_system_prompt and raw_system_prompt.strip() else None
+        )
 
         # Validate max_tokens against model-specific limits.
         # Skip the warning when extended thinking is enabled — the per-call
--- a/src/conductor/providers/claude.py
+++ b/src/conductor/providers/claude.py
@@ -977,6 +977,7 @@ class ClaudeProvider(AgentProvider):
         # Done before the per-model max_tokens warning so the warning logic
         # accounts for thinking-aware caps.
         thinking = self._resolve_thinking_for_agent(agent, model)
+        system_prompt = (agent.system_prompt or "").strip() or None
 
         # Validate max_tokens against model-specific limits.
         # Skip the warning when extended thinking is enabled — the per-call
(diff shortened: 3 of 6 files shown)


src/conductor/providers/claude.py
```diff
@@ -977,6 +977,7 @@ async def _execute_with_retry(
         # Done before the per-model max_tokens warning so the warning logic
         # accounts for thinking-aware caps.
         thinking = self._resolve_thinking_for_agent(agent, model)
+        system_prompt = (agent.system_prompt or "").strip() or None
```
jrob5756: Using `.strip()` here changes every non-empty system prompt with leading or trailing whitespace. The other providers pass the rendered value through unchanged, so Claude should only use trimming to detect a blank prompt. An exact-equality test with a multiline templated prompt would catch this.

```suggestion
        raw_system_prompt = agent.system_prompt
        system_prompt = (
            raw_system_prompt if raw_system_prompt and raw_system_prompt.strip() else None
        )
```

src/conductor/config/validator.py
```diff
@@ -196,12 +194,11 @@ def validate_workflow_config(
...
                 f"Agent '{agent.name}' defines `system_prompt` but no `prompt` "
                 "(or only whitespace). "
-                "Some providers (e.g., Claude) ignore `system_prompt` entirely, "
-                "which would send an empty user message to the model. "
-                "Even with the Copilot provider, the model often responds poorly "
-                "to a missing user prompt. Move the dynamic, must-execute content "
-                "(input references, instructions) into a `prompt:` block; keep the "
-                "persona and static methodology in `system_prompt:`."
+                "This sends an empty user message to the model on every provider, "
+                "which almost always indicates a latent authoring mistake. Move "
```
jrob5756: `on every provider` isn't true for Copilot, which sends the system and user sections as one non-empty SDK prompt. Describing the missing workflow task prompt keeps the warning accurate across providers.

```suggestion
                "The user-authored task prompt is empty, which almost always "
                "indicates a latent authoring mistake. Move "
```

plugins/conductor/skills/conductor/references/execution.md
```diff
@@ -221,7 +221,7 @@ Performs both schema and **semantic** checks:
 - Parallel group agent references
 - For-each `source` format and reserved names
 - Stale agent references and undeclared explicit-mode dependencies in `prompt`, `system_prompt`, `command`, `args`, `working_dir`, `input_mapping`, parallel-group inputs, and workflow `output:` templates
-- Warning when an agent defines `system_prompt` but no `prompt:` (unusual — system prompts are only meaningful alongside a user prompt)
+- Warning when an agent defines `system_prompt` but no `prompt:` (unusual — the user message would be empty)
```
jrob5756: This has the same provider-specific wording problem as the validator warning. Copilot doesn't send an empty SDK prompt; the workflow's task prompt is what is missing.

```suggestion
- Warning when an agent defines `system_prompt` but no `prompt:` (unusual because the user-authored task prompt would be empty)
```

tests/test_providers/test_claude_system_prompt.py
```diff
@@ -0,0 +1,248 @@
...
+                "prompt": "Test prompt",
+                "system_prompt": "Rendered tool-loop instructions",
+            }
+        )
+
+        await provider.execute(agent=agent, context={}, rendered_prompt="Test prompt")
+
+        assert mock_client.messages.create.call_count == 2
+        for call in mock_client.messages.create.call_args_list:
+            assert call.kwargs["system"] == "Rendered tool-loop instructions"
```
jrob5756, line 121: These two calls are turns within one execution, not the outer retry loop. Since the PR promises the system prompt survives retries, please add a retryable failure followed by success and assert both requests receive the same `system` value.

tests/test_providers/test_claude_system_prompt.py
```diff
@@ -0,0 +1,248 @@
...
+    async def test_interrupt_partial_output_passes_system_prompt(
+        self, mock_anthropic_module: Mock, mock_anthropic_class: Mock
+    ) -> None:
+        """Requirement AC-A(d): interrupt partial-output SDK call sends system_prompt."""
+        provider, mock_client = _mock_claude_provider(mock_anthropic_module, mock_anthropic_class)
+        mock_client.messages.create = AsyncMock(
+            return_value=_tool_response("emit_output", {"result": "partial"})
+        )
+        interrupt_signal = asyncio.Event()
+        interrupt_signal.set()
```
jrob5756, line 205: Setting the event before `execute()` only covers the between-turn branch. The in-flight race takes a different path, so please start a blocked API call, set the event after it begins, and assert the partial-output request keeps the system prompt.

jrob5756 (commented):
Thank you for contributing to Conductor! A couple things that I caught... this fixes the missing Claude system prompt path, but the current normalization changes non-empty prompt text. I also left two wording corrections and two coverage requests for the retry and interrupt branches.

If you can wrap these up, we can merge! Thanks again.

src/conductor/providers/claude.py
hertznsk: Fixed in bd0aa63. `.strip()` is now used only as a blank-prompt predicate — the rendered value is forwarded verbatim:

```python
raw_system_prompt = agent.system_prompt
system_prompt = (
    raw_system_prompt if raw_system_prompt and raw_system_prompt.strip() else None
)
```

This matches the pass-through behavior of the Copilot, Claude Agent SDK, and Hermes providers. The new retry and in-flight interrupt tests also pin verbatim forwarding with a leading-whitespace / trailing-newline prompt (`"  Preserve me exactly\n"`).

src/conductor/config/validator.py
hertznsk: Fixed in bd0aa63 — the warning (and the comment above it) now uses your suggested wording:

> The user-authored task prompt is empty, which almost always indicates a latent authoring mistake. Move ...

The test in `tests/test_config/test_validator.py` now also pins the new phrasing and asserts the old "empty user message" / "on every provider" claims are absent.

plugins/conductor/skills/conductor/references/execution.md
hertznsk: Fixed in bd0aa63 — applied your suggestion verbatim:

> - Warning when an agent defines `system_prompt` but no `prompt:` (unusual because the user-authored task prompt would be empty)

tests/test_providers/test_claude_system_prompt.py
hertznsk: Added in bd0aa63: `test_retry_loop_passes_system_prompt_on_every_attempt`. The first `messages.create` raises a retryable `ProviderError(status_code=503)`, the second succeeds, and the test asserts both requests carry the same verbatim `system` value (`"  Preserve me exactly\n"`), plus `len(provider.get_retry_history()) == 1` to prove the outer `_execute_with_retry` loop actually re-ran. Backoff is neutralized via `RetryConfig(base_delay=0, max_delay=0, jitter=0)` and a patched `asyncio.sleep`.

tests/test_providers/test_claude_system_prompt.py
hertznsk: Added in bd0aa63: `test_in_flight_interrupt_partial_output_passes_system_prompt`. The first `messages.create` signals `call_started` and then blocks on an event that is never released; the test sets the interrupt only after the call has begun, so the provider takes the `asyncio.wait(FIRST_COMPLETED)` race branch, cancels the in-flight task, and issues the partial-output request. The test asserts that request keeps the verbatim system prompt (trailing newline included) and that the appended interrupt message is present. The pre-existing between-turn test is kept as-is.

PR #294, opened 2026-07-13
fix(cli): restore terminal from cbreak mode after run exits

### LINKED ISSUE #290: Terminal left in cbreak mode (no echo) after conductor run exits
### Symptom

After a `conductor run <workflow.yaml>` completes (exit code 0), the invoking terminal is left with `-echo -icanon`: typed characters are not displayed (input still works). Recovery requires typing `reset` or `stty sane` blindly.

Observed on Linux (kernel pty, tmux 3.7b and bare terminal), conductor-cli `v0.1.20` (also reproduced on `main` @ `7aaa589`), Python 3.12. The failure is **intermittent** — most runs restore the terminal correctly; some runs leave it broken.

### Root cause

`KeyboardListener._original_settings` (the termios attrs to restore on exit) is captured once in `start()`:

```python
# src/conductor/interrupt/listener.py:106-111
try:
    self._original_settings = termios.tcgetattr(sys.stdin.fileno())
except termios.error:
    ...
```

but the listener lifecycle allows cbreak mode to be re-entered **without re-capturing or invalidating** the saved attrs:

- `suspend()` (called around gates/dialogs, e.g. `engine/workflow.py` `_suspend_listener`) restores the terminal via `tcsetattr(TCSADRAIN, self._original_settings)` **but keeps `_original_settings` set** (listener.py:157-185);
- `resume()` re-enters cbreak via `tty.setcbreak()` **without re-reading the current attrs** (listener.py:187-221);
- if `start()` or `setcbreak` is ever executed while the tty is *already* in cbreak (double listener instance, or a suspend/resume interleaving), `tcgetattr` captures the **cbreak state itself** as the "original".

`stop()` then "restores" by applying that captured cbreak state, and the `atexit` fallback is a no-op (`_original_settings` was already cleared), so the terminal stays at `-echo -icanon`.

### Evidence (instrumented real run)

Monkey-patched `termios.tcsetattr`/`KeyboardListener` in a real `conductor run` of an 11-agent workflow. Broken run:

```
02:07:11  tcsetattr(0) TCSAFLUSH  lflag=0x8a31   ← start(): entering cbreak (echo/icanon OFF)
02:08:08  tcsetattr(0) TCSADRAIN  lflag=0x8a31   ← stop(): "restore" applies the SAME cbreak lflag
02:08:08  tcgetattr(0) → echo=False icanon=False ← terminal remains broken
02:08:11  ATEXIT → echo=False icanon=False       ← fallback handler is a no-op
```

Clean run (for comparison) — the restore applies the correct pre-cbreak attrs (`lflag=0x8a3b`) and the terminal survives:

```
02:17:27  tcsetattr(0) TCSADRAIN  lflag=0x8a3b   ← stop(): real restore
02:17:27  tcgetattr(0) → echo=True icanon=True
```

Reproduced 8 times across runs; the deciding factor is whether `_original_settings` still holds the true pre-cbreak attrs at `stop()` time.

Minimal unit demonstration of the capture bug:

```python
# second start() while already in cbreak saves the cbreak state as "original"
l1 = KeyboardListener(interrupt_event=ev); await l1.start()   # saves sane attrs
l2 = KeyboardListener(interrupt_event=ev); await l2.start()   # saves cbreak attrs (lflag & ~ECHO & ~ICANON)
await l2.stop()   # "restores" cbreak — terminal stays broken
# → l2._original_settings shows echo=False icanon=False
```

### Why current tests don't catch it

`tests/test_interrupt/test_listener.py` mocks `termios` (`test_stop_restores_terminal` asserts `tcsetattr` was *called* with the mock's settings) — it never exercises a real pty, so capturing a degraded state goes unnoticed.

### Suggested fixes

1. **Re-capture on every cbreak entry.** Before each `tty.setcbreak()` (both `start()` and `resume()`), read `tcgetattr` and store it as the restore target **only if the tty is not already in cbreak** — or simpler: capture the initial attrs once per *process* (module-level), never overwrite them.
2. **Invalidate on suspend.** `suspend()` should clear `_original_settings` the same way `stop()` does, forcing `resume()` to re-capture.
3. Use `TCSANOW` instead of `TCSADRAIN` for restore so it cannot be delayed by output drain.
4. Add an integration test on a real pty (`pty.fork()`): start → suspend → resume → stop, then `tcgetattr` and assert `ECHO|ICANON` are set.

### Workaround for users

`reset` or `stty sane` restores echo. Alternatively run `conductor run --no-interactive` when Esc-interrupt is not needed.

hertznsk:
## Summary

Fixes #290 — Conductor could leave the invoking terminal in cbreak mode (no echo / no line editing) after a workflow exited, requiring a blind `reset` or `stty sane` to recover.

## Why this was reworked

An earlier revision of this PR only fixed the SIGTERM cleanup handler (delegating to the previous disposition instead of swallowing the signal, and capturing the disposition into the handler closure to avoid self-recursion). Verified against a real pseudo-terminal, that alone **did not fix the issue**: the terminal was still left in cbreak after ordinary exits and Ctrl+C, because the two actual root causes were elsewhere. The implementation was therefore reworked around them; the SIGTERM correctness fix is retained as one part of the whole.

## Root cause

1. **Baseline re-snapshot.** `KeyboardListener.start()` captured the tty settings on every call. Under specific interleavings — a second listener instance in the same process, or a second `start()` on an already-active listener — the captured "original" settings were the already-broken cbreak state, which `stop()` then faithfully restored.
2. **Post-restore corruption.** Provider cleanup could alter terminal settings *after* `listener.stop()` had completed, so even a correct restore at the listener level did not guarantee a sane terminal at process exit.

Additionally, the SIGTERM handler swallowed the signal, leaving the process running.

## Changes

- **Process-global baseline**: Capture the tty baseline **once per process** into a module-level cache before the first `tty.setcbreak()`; every later listener instance reuses it instead of re-snapshotting the live tty.
- **Idempotent start**: Make `start()` idempotent while the listener is truly active (baseline held AND reader thread running), so a duplicate call can't overwrite the baseline or spawn duplicate threads. Start-after-`suspend()` still restarts correctly.
- **Robust restore**: Restore with `TCSANOW` instead of `TCSADRAIN` so a blocked output drain can't delay the restore.
- **Retry capability**: Clear the saved baseline only after a successful `tcsetattr`, so a transient restore failure can be retried by `atexit`/SIGTERM/`stop()`.
- **SIGTERM correctness**: The SIGTERM cleanup handler now delegates to the previously-installed disposition after restoring: `SIG_DFL` resets and re-raises via `os.kill`, `SIG_IGN` stays ignored, a callable previous handler is invoked. The previous disposition is captured into the handler closure at registration time to prevent infinite self-recursion.
- **Outermost teardown protection**: The run and resume commands now reapply and retire the process baseline at their **outermost cleanup boundary** (after provider shutdown and all other teardown steps, via `restore_terminal_baseline(clear=True)`). This ensures that even if a provider's cleanup puts the terminal back into cbreak, normal exit, Ctrl+C, and late cleanup failures all return a sane terminal to the shell — and that a later invocation in the same process never applies a stale baseline to a new terminal.

## Tests

- **New real-pty integration tests** (`tests/test_interrupt/test_listener_pty.py`) reproduce both corruption scenarios on a genuine pseudo-terminal: two listener instances in one process, and double-`start()` on one instance.
- **CLI boundary real-pty tests** (`tests/test_interrupt/test_cli_terminal_pty.py`) verify that provider teardown cannot bypass final TTY restoration, that a later non-interactive run doesn't reuse an old TTY baseline, and that the real CLI restores exact TTY attrs on exit and Ctrl+C.
- **New mocked unit tests** for the baseline cache, idempotent start, `TCSANOW` restore paths, restore retry semantics, all SIGTERM delegation cases, and process baseline restoration after listener stop.
- Full suite passes; `make lint` and `make typecheck` clean.

## Scope

- `src/conductor/interrupt/listener.py`
- `src/conductor/cli/run.py`
- Tests in `tests/test_interrupt/`
- `CHANGELOG.md`

No behavior change for `suspend()`/`resume()` semantics. Windows is unaffected (the listener is already Unix-only).

### DIFF
--- a/src/conductor/interrupt/listener.py
+++ b/src/conductor/interrupt/listener.py
@@ -31,6 +31,15 @@ _CTRL_G_BYTE = 0x07
 # Timeout for disambiguating bare Esc from escape sequences (seconds)
 _ESC_DISAMBIGUATE_TIMEOUT = 0.05
 
+_CAPTURED_BASELINE_SETTINGS: Any = None
+"""Process-wide tty baseline captured on the FIRST successful start().
+
+Subsequent KeyboardListener instances reuse this baseline so a second
+listener never re-captures cbreak state as "original". Process-lifetime
+and intentionally never reset by production code: after stop() the tty
+has been restored to exactly this baseline, so reuse is always
+legitimate. Tests reset it via a fixture. See issue #290."""
+
 
 @dataclass
 class KeyboardListener:
@@ -103,9 +112,27 @@ class KeyboardListener:
         self._loop = asyncio.get_running_loop()
         self._stop_flag = False
 
-        # Save original terminal settings
+        # Save original terminal settings. A second listener must never
+        # re-snapshot an already-cbreak terminal as its "original" state, so
+        # the baseline is captured once per process into a module-level cache
+        # (issue #290).
+        global _CAPTURED_BASELINE_SETTINGS
+
+        # Idempotent guard: a start() on a truly-active listener (baseline
+        # held AND reader thread running) is a no-op so it cannot overwrite
+        # the baseline or spawn duplicate threads. After suspend() the thread
+        # is None, so start-after-suspend correctly falls through below.
+        if self._original_settings is not None and self._reader_thread is not None:
+            logger.debug("Keyboard listener already active, start() is a no-op")
+            return
+
         try:
-            self._original_settings = termios.tcgetattr(sys.stdin.fileno())
+            if _CAPTURED_BASELINE_SETTINGS is not None:
+                # Reuse the process-wide pre-listener baseline.
+                self._original_settings = _CAPTURED_BASELINE_SETTINGS
+            else:
+                self._original_settings = termios.tcgetattr(sys.stdin.fileno())
+                _CAPTURED_BASELINE_SETTINGS = self._original_settings  # pyright: ignore[reportConstantRedefinition]
         except termios.error:
             logger.debug("Failed to get terminal settings, listener not started")
             return
@@ -178,7 +205,7 @@ class KeyboardListener:
             try:
                 import termios
 
-                termios.tcsetattr(sys.stdin.fileno(), termios.TCSADRAIN, self._original_settings)
+                termios.tcsetattr(sys.stdin.fileno(), termios.TCSANOW, self._original_settings)
             except (ImportError, termios.error, ValueError, OSError):
                 pass
 
@@ -226,10 +253,12 @@ class KeyboardListener:
             try:
                 import termios
 
-                termios.tcsetattr(sys.stdin.fileno(), termios.TCSADRAIN, self._original_settings)
+                termios.tcsetattr(sys.stdin.fileno(), termios.TCSANOW, self._original_settings)
+                # Clear the baseline only after a successful restore so a
+                # transient failure keeps it for a later retry (issue #290).
+                self._original_settings = None
             except (ImportError, termios.error, ValueError, OSError):
                 pass
-            self._original_settings = None
 
     def _register_cleanup_handlers(self) -> None:
         """Register atexit and SIGTERM handlers for crash safety."""
--- a/src/conductor/interrupt/listener.py
+++ b/src/conductor/interrupt/listener.py
@@ -7,6 +7,11 @@ Esc vs ANSI escape sequence disambiguation using a 50ms read-ahead timeout.
 Uses a dedicated daemon thread for blocking stdin reads, delivering bytes
 into an ``asyncio.Queue`` via ``loop.call_soon_threadsafe``. This avoids
 thread leaks from abandoned ``run_in_executor`` futures.
+
+Terminal safety (issue #290): the original tty settings are captured once per
+process into the module-level ``_CAPTURED_BASELINE_SETTINGS`` cache so a second
+listener never re-captures cbreak state as "original", and the SIGTERM cleanup
+handler restores the terminal before delegating to the previous disposition.
 """
 
 from __future__ import annotations
@@ -15,6 +20,7 @@ import asyncio
 import atexit
 import contextlib
 import logging
+import os
 import select
 import signal
 import sys
@@ -85,6 +91,9 @@ class KeyboardListener:
 
     _previous_sigterm: Any = field(default=None, repr=False)
     """Previous SIGTERM handler for restoration."""
+
+    _sigterm_handler: Any = field(default=None, repr=False)
+    """This instance's own installed SIGTERM handler closure (issue #290)."""
 
     _byte_queue: asyncio.Queue[int | None] = field(default_factory=asyncio.Queue, repr=False)
     """Async queue for delivering bytes from the reader thread."""
@@ -266,17 +275,52 @@ class KeyboardListener:
             atexit.register(self._restore_terminal)
             self._atexit_registered = True
 
-        # Install SIGTERM handler that restores terminal then re-raises
+        # Install a SIGTERM handler that restores the terminal, then delegates
+        # to the previously-installed disposition so the process terminates
+        # with its expected action (issue #290).
         try:
-            self._previous_sigterm = signal.getsignal(signal.SIGTERM)
+            # Guard: if THIS instance's own handler is still installed, skip
+            # re-registration. Re-registering would capture our own closure as
+            # the "previous" disposition, recursing forever on invocation. The
+            # identity check targets this instance's stored handler only — a
+            # stale closure from a stopped listener must not block a new one.
+            if (
+                self._sigterm_handler is not None
+                and signal.getsignal(signal.SIGTERM) is self._sigterm_handler
+            ):
+                return
+
+            # Capture the previous disposition into a LOCAL variable so the
+            # closure is immutable: reading the mutable field at invocation
+            # time would let a later re-registration rewrite what an
+            # already-installed closure delegates to.
+            previous = signal.getsignal(signal.SIGTERM)
+            self._previous_sigterm = previous  # backward-compat introspection
 
             def _sigterm_handler(signum: int, frame: Any) -> None:
-                self._restore_terminal()
-                # Call previous handler if it was callable
-                if callable(self._previous_sigterm):
-                    self._previous_sigterm(signum, frame)
+                # A failed restore must never swallow the signal: swallow any
+                # error here so the handler always reaches the delegation path.
+                # The baseline is dropped unconditionally afterward — the
+                # process is terminating, so a stale value must not be reused.
+                try:
+                    self._restore_terminal()
+                except Exception:
+                    pass
+                finally:
+                    self._original_settings = None
+                if previous is signal.SIG_DFL:
+                    # Reset to default and re-raise so the default action
+                    # (terminate) actually runs.
+                    signal.signal(signal.SIGTERM, signal.SIG_DFL)
+                    os.kill(os.getpid(), signum)
+                elif previous is signal.SIG_IGN:
+                    # Caller asked to ignore SIGTERM: restore and stay alive.
+                    return
+                elif callable(previous):
+                    previous(signum, frame)
 
             signal.signal(signal.SIGTERM, _sigterm_handler)
+            self._sigterm_handler = _sigterm_handler
         except (OSError, ValueError):
             # Can't set signal handler (not main thread, etc.)
             pass
--- a/tests/test_interrupt/test_listener_pty.py
+++ b/tests/test_interrupt/test_listener_pty.py
@@ -1,0 +1,205 @@
+"""Real-pty integration tests for KeyboardListener terminal restore (issue #290).
+
+These tests allocate a genuine pseudo-terminal via ``pty.openpty()`` and point
+fd 0 (and ``sys.stdin``) at its slave so the production ``KeyboardListener``
+lifecycle (``termios.tcgetattr`` / ``tty.setcbreak`` / ``termios.tcsetattr``)
+runs against a real tty instead of mocks. This is the only reliable way to
+reproduce the bug where a listener started on an already-cbreak terminal
+snapshots cbreak as its "original" settings and later restores them.
+
+Expected state on unfixed code (TDD red):
+
+- ``test_full_lifecycle_restores_terminal``            — PASSES (regression guard)
+- ``test_second_listener_does_not_capture_cbreak``     — FAILS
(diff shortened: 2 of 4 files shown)


hertznsk:
@jrob5756 please take a look at this PR

src/conductor/interrupt/listener.py
```diff
@@ -32,6 +37,32 @@
...
+    if _captured_baseline_settings is None:
+        return
+
+    try:
+        import termios
+
+        termios.tcsetattr(sys.stdin.fileno(), termios.TCSANOW, _captured_baseline_settings)
+        if clear:
+            _captured_baseline_settings = None
+    except (ImportError, ValueError, OSError):
```
jrob5756: **BLOCKING**: `termios.error` isn't a subclass of `OSError`, so it slips past this handler. `suspend()` and `_restore_terminal()` have the same gap. If stdin is closed or no longer a tty during teardown, `listener.stop()` and the new outermost run/resume restoration can raise a terminal-cleanup error that overwrites the workflow's actual result or exception. On SIGTERM, an exception from the instance-level restore also skips the process-baseline restore that's supposed to run after it. Worth noting: the new retry test hides this by raising `OSError` instead of the real `termios.error`, so it never exercises the actual failure mode.

Separate the termios import from the restore call, then catch `(termios.error, ValueError, OSError)` around every `tcsetattr()` in all three restoration paths. Keep the saved settings around on failure — only clear them once restoration actually succeeds — and update the tests to raise real `termios.error` and assert the original workflow outcome survives a failed restore.

tests/test_interrupt/test_cli_terminal_pty.py
```diff
@@ -0,0 +1,148 @@
...
+        "run",
+        str(Path("examples/wait-smoke.yaml").resolve()),
+    ]
+    if interrupt:
+        command.extend(["--input", "middle_duration_ms=30000"])
+
+    pid, master_fd = pty.fork()
+    if pid == 0:
+        os.execv(command[0], command)
+
```
jrob5756, line 196: **BLOCKING**: `os.execv()` replaces the process, which drops the autouse fixtures' monkeypatches on `tempfile.gettempdir()`, `pid_dir()`, and `runs_dir()`. `CONDUCTOR_HOME` survives the exec, but `HOME` and the tmp-dir env vars don't, so the child's `run_workflow_async()` startup sweep runs against the developer's real environment — it can prune real legacy PID files and delete real event logs. That's exactly the kind of side effect the fixtures elsewhere in this suite exist to prevent.

Have this test take `tmp_path`, build a dedicated home/tmp directory tree under it, and launch with `os.execve()` passing an explicit environment that overrides `HOME`, `TMPDIR`, `TMP`, `TEMP`, and `CONDUCTOR_HOME` to those directories — don't rely on the parent process's monkeypatches to carry through an exec.

tests/test_interrupt/test_listener_pty.py
```diff
@@ -0,0 +1,228 @@
...
+- ``test_double_start_same_instance_does_not_corrupt_baseline`` — FAILS
+"""
+
+from __future__ import annotations
+
+import asyncio
+import contextlib
+import io
+import os
+import pty
```
jrob5756: **BLOCKING**: `pty` and `termios` are imported before the Windows skip marker gets a chance to run, and neither module exists on Windows — so instead of skipping, this fails collection entirely on the Windows CI job. `test_cli_terminal_pty.py` has the same problem and no platform guard at all.

```suggestion
import sys

import pytest

if sys.platform == "win32":
    pytest.skip("PTY tests are Unix-only", allow_module_level=True)

import pty
import termios
```

Apply the same guard (or `pytest.importorskip`) to `test_cli_terminal_pty.py` before it imports the PTY helper.

src/conductor/interrupt/listener.py
```diff
@@ -32,6 +37,32 @@
...
+    """Restore the process-wide TTY baseline and optionally retire it."""
+    global _captured_baseline_settings
+
+    if _captured_baseline_settings is None:
+        return
+
+    try:
+        import termios
+
+        termios.tcsetattr(sys.stdin.fileno(), termios.TCSANOW, _captured_baseline_settings)
```
jrob5756: **RECOMMENDED**: The cached settings aren't tied to the terminal they were captured from — restoration just applies whatever's cached to whatever `sys.stdin` happens to be at that moment. If the final restore (with `clear=True`) hits a caught error, like `ValueError` on a closed stdin, the old baseline stays cached. A later invocation on a different terminal — even a non-interactive one, since the outer restore runs unconditionally — can then reuse or apply that stale baseline and change the second terminal's echo/canonical settings instead of leaving it alone.

Store the terminal identity and an owned file descriptor alongside the captured settings, restore through that descriptor rather than the current `sys.stdin`, and only reuse a baseline for the same terminal (closing the retained fd when it's retired). A two-PTY test covering a failed final restore followed by a second terminal's invocation would catch regressions here.

jrob5756 (commented):
CI update: the Python 3.12 telemetry failure was a flaky test already on main. I reproduced it there and fixed it in #532, then rebased this branch onto the corrected main without changing your patch. Fresh CI is at https://github.com/microsoft/conductor/actions/runs/35010351700. That shared failure is no longer something this PR needs to fix.

The three blocking findings in the inline comments remain open:
- `src/conductor/interrupt/listener.py:63`: `termios.error` isn't caught during restoration, so a real `tcsetattr` failure can overwrite the workflow's result or exception.
- `tests/test_interrupt/test_cli_terminal_pty.py:123`: the exec'd child can reach the developer's real HOME/tmp and prune real event logs and PID records.
- `tests/test_interrupt/test_listener_pty.py:23`: `pty`/`termios` are imported before the Windows skip can fire. `test_cli_terminal_pty.py` has the same collection problem. This caused both Windows CI job failures; they are two reports of the same import issue, not separate defects.

Please address those in the branch and rerun CI. I have left the contributor-specific fixes to you.

The non-blocking recommendation also remains: tie the retained terminal settings to the terminal they came from so a later invocation cannot apply a stale baseline to a different terminal.

jrob5756:
> Jason Robert (@jrob5756) please take a look at this PR

Sorry, fell off my radar!

src/conductor/interrupt/listener.py
hertznsk: Verified — `termios.error.__mro__` is `(termios.error, Exception, BaseException, object)`, so it slipped past all three `(ImportError, ValueError, OSError)` handlers exactly as described. Fixed in 0a863eb:

- The termios import is separated from the restore call (`_try_import_termios()`), and every `tcsetattr()` in `restore_terminal_baseline()`, `suspend()`, and `_restore_terminal()` now catches `(termios.error, ValueError, OSError)`. Saved settings are still cleared only after a successful restore, so a transient failure stays retryable.
- The retry tests now raise a real `termios.error` (via `pytest.importorskip("termios")`) instead of `OSError`, so they exercise the actual failure mode.
- New CLI-level test `test_run_outcome_survives_failed_final_restore` arms `tcsetattr` to fail with a real `termios.error` for the whole teardown and asserts the workflow outcome survives in both modes: the result dict is returned on success, and the provider-cleanup `RuntimeError` propagates unchanged on failure.

tests/test_interrupt/test_cli_terminal_pty.py
hertznsk: Correct — the monkeypatches on `tempfile.gettempdir()`, `pid_dir()`, and `runs_dir()` are process-local and are lost across the exec, so the child's startup retention sweep would have run against the real directories. Fixed in fd86f25: the test now takes `tmp_path`, builds `home/` and `tmp/` trees under it, and launches with `os.execve()`, passing an explicit environment that overrides `HOME`, `TMPDIR`, `TMP`, `TEMP`, and `CONDUCTOR_HOME` (everything else inherited from `os.environ`). Also added `os._exit(127)` after the exec, mirroring the other fork-based test.

tests/test_interrupt/test_listener_pty.py
hertznsk: Fixed in 94b83f1 — both modules now run the win32 module-level `pytest.skip(..., allow_module_level=True)` before importing `pty`/`termios`, and before `test_cli_terminal_pty.py` imports the shared PTY helper. The deferred imports carry `# noqa: E402` since ruff selects pycodestyle errors, and the now-redundant `pytestmark` skipif in `test_listener_pty.py` was removed.

src/conductor/interrupt/listener.py
hertznsk: Implemented in 0a863eb. The cache is now a `_TerminalBaseline`: the captured settings, an owned `os.dup()` of the capture-time descriptor (holding it open also pins the pts index so it cannot be recycled while the baseline lives), and the `(st_dev, st_ino)` identity of the capture-time terminal. `start()` reuses the baseline only when the current stdin's identity matches, and retires a stale one (closing its fd) rather than letting a different terminal inherit or be overwritten with it; `restore_terminal_baseline()` restores through the owned descriptor rather than the current `sys.stdin`, and a successful `clear=True` retires it, closing the fd.

Two extensions beyond the sketch, from a second review pass over the same mechanism: each listener now also keeps its own owned fd, so `suspend()`/`stop()`/`resume()` target the terminal the settings were captured from even if `sys.stdin` was replaced while the listener was still active; and a failed `start()` (capture or `setcbreak`) now disposes of the descriptors and of a baseline it created instead of leaving them for the rest of the process — while a listener that merely reused the shared baseline leaves it alone.

Coverage: `test_failed_restore_does_not_leak_baseline_into_other_terminal` (failed final restore on pty A, then a full listener lifecycle on pty B) and `test_listener_restore_targets_original_terminal_after_stdin_swap` (stdin swapped to a different pty while the listener is active; suspend and stop must leave the replacement terminal untouched).

hertznsk:
All three blocking findings and the recommendation are addressed and pushed:

- 0a863eb `fix(interrupt): catch termios.error and bind the tty baseline to its terminal` — findings 1 and 4
- fd86f25 `test(interrupt): isolate the exec'd CLI PTY test environment` — finding 2
- 94b83f1 `test(interrupt): guard Unix-only PTY imports so Windows CI skips cleanly` — finding 3

Details in the respective threads. Local verification: `tests/test_interrupt` + `tests/test_cli` — 1283 passed, 3 skipped; ruff check/format clean; `ty` clean. Thanks for the rebase onto the corrected main — looking forward to the fresh CI run.

jrob5756 (approved):

PR #297, opened 2026-07-15
feat(working-dir): working_dir for LLM agents and their MCP servers

hertznsk:
## Summary

Adds an optional `working_dir` to LLM agents (and a workflow-wide `runtime.working_dir` default) so an agent and its MCP servers run in a chosen directory. Engine-resolved with precedence **agent > runtime > `os.getcwd()`**, supporting static values and Jinja templates.

## What's included

- **Schema** — `RuntimeConfig.working_dir` + per-agent `working_dir`; rejected on `wait`/`set`/`terminate`/`human_gate`/`workflow` step types; new `ProviderCapabilities.working_dir` (copilot/claude = true, hermes/claude-agent-sdk = false) with a `conductor validate` gate.
- **Engine** — single resolver applied in all three execution paths (linear, parallel group, for_each iteration).
- **Copilot provider** — session `working_directory` + per-server MCP cwd stamping; `copilot_session_cwds` persisted in checkpoints; resume skips session resume when the cwd changed.
- **Claude provider** — MCP connections pooled per working directory (`MCPManager.connect_server(cwd)`), with per-cwd double-checked locking so parallel agents with different cwds stay isolated and agents sharing a cwd reuse one manager.
- **Validator** — the synthetic validator agent inherits the primary agent's resolved `working_dir`.
- **Events** — additive LLM-only `parallel_agent_started` / `for_each_agent_started` carry the resolved `working_dir` into the JSONL log; existing envelopes unchanged (the dashboard safely drops unknown types).
- **Docs + example** — new *Working Directory* section in `docs/mcp-tools.md`, `docs/workflow-syntax.md` updates, and `examples/working-dir.yaml`.

## Scope notes (deliberate)

- `working_dir` is **not** a sandbox — it only sets the process cwd; docs say so explicitly.
- Sub-workflows do **not** inherit the parent's `working_dir`; a child resolves its own relative path against its own workflow file (pinned by test).
- Dialog turns are unaffected.

## Verification

- Full suite: `1 failed, 3937 passed` — the single failure is `test_copilot_large_write.py::test_large_create_tool_call_does_not_truncate`, a pre-existing environmental Copilot 403 (enterprise-policy) failure reproduced identically on clean `main`.
- `make check` (ruff + ty) green (one pre-existing ty warning in `dialog_evaluator.py`).
- `make validate-examples` and `conductor validate examples/working-dir.yaml` pass.
- Mock-provider E2E smoke exercised runtime/agent cwd across linear, parallel, and for_each paths plus Copilot session/MCP/checkpoint and Claude pool-by-cwd.
- Four independent final-wave reviewers (plan compliance, code quality, real manual QA, scope fidelity) all approved.


src/conductor/providers/claude.py
```diff
@@ -520,49 +556,66 @@ async def _ensure_mcp_connected(self) -> None:
...
-                    logger.error(f"Failed to connect to MCP server '{name}': {e}")
-                    # Continue with other servers
-            else:
-                logger.warning(
-                    f"MCP server '{name}' has unsupported type '{server_type}' "
-                    "(Claude provider only supports 'stdio')"
-                )
+
+            self._mcp_managers[resolved_cwd] = manager
+            return manager
```
jrob5756: This caches `manager` even when every configured server failed to connect for `resolved_cwd`. The fast path above (`if resolved_cwd in self._mcp_managers: return ...`) means a later call for the same cwd returns this same broken manager without ever calling `connect_server` again -- a one-time transient failure (spawn race, directory not fully populated yet) becomes permanent for the rest of the run, and after the initial `logger.error` there's no further signal that this cwd is running with zero MCP tools.

`has_servers()` already exists and is used for exactly this check a few lines down (`if mcp_manager and mcp_manager.has_servers():`), so gating the cache write on it fixes this without touching the locking structure:

```suggestion
            if manager.has_servers():
                self._mcp_managers[resolved_cwd] = manager
            return manager
```

tests/test_providers/test_claude.py
```diff
@@ -3188,3 +3192,279 @@ async def test_thinking_kwarg_forwarded_through_parse_recovery_path(
...
+        servers = {"fs": {"command": "npx", "args": []}}
+        provider = self._build_provider(mock_anthropic_module, mock_anthropic_class, servers)
+
+        instances: list[Mock] = []
+        fake_cls = self._manager_factory(instances)
+        with (
+            patch("conductor.mcp.manager.MCP_SDK_AVAILABLE", True),
+            patch("conductor.mcp.manager.MCPManager", fake_cls),
+        ):
+            results = await asyncio.gather(
```
jrob5756, line 3582: The fake `connect_server` in `_manager_factory` (around line 3232) never awaits anything, so `asyncio.gather` here runs all four coroutines to completion sequentially rather than actually interleaving them. I removed the entire locking mechanism (guard lock, per-cwd lock, and the re-check) from `_get_mcp_manager_for_cwd` and re-ran this test -- it still passes. The double-checked locking this test is meant to protect only gets exercised when something inside the critical section actually suspends.

Worth giving the fake a real yield point (`await asyncio.sleep(0)`, or gate on an `asyncio.Event` released after the first N callers enter) so concurrent callers genuinely race, then assert the lock still produces exactly one manager per cwd.

tests/test_providers/test_claude.py
```diff
@@ -3188,3 +3192,279 @@ async def test_thinking_kwarg_forwarded_through_parse_recovery_path(
...
+    async def test_mcp_pool_connect_failure_fail_open_per_cwd(
+        self, mock_anthropic_module: Mock, mock_anthropic_class: Mock
+    ) -> None:
+        """Requirement: a connect failure for one cwd must not break another cwd.
+
+        Fail-open per-server connect is existing behavior (the error is
+        logged and remaining servers still connect). The pool must preserve
+        it across pool keys: cwd A failing to connect leaves cwd B fully
+        functional, and the failed key is not cached so a later retry can
+        succeed.
```
jrob5756: This claims the failed key isn't cached so a later retry can succeed, but the test never calls `_get_mcp_manager_for_cwd("/repo/bad")` a second time to check that. I did -- it returns the same never-connected manager with no second `connect_server` attempt (see the cache-write comment on `claude.py:593`). Either this docstring should describe the actual behavior (manager is cached even on total connect failure, no retry), or the missing retry test/behavior should be added.

src/conductor/engine/workflow.py
```diff
@@ -3073,20 +3123,44 @@ async def _execute_loop(self, current_agent_name: str) -> dict[str, Any]:
...
+                        started_payload: dict[str, Any] = {
+                            "agent_name": agent.name,
+                            "iteration": agent_execution_count,
+                            "agent_type": agent.type or "agent",
+                            "context_window_max": await self._get_context_window_for_agent(
+                                resolved_agent
+                            ),
+                        }
+                        if agent.type in (None, "agent"):
+                            started_payload["working_dir"] = resolved_agent.working_dir
```
jrob5756: Minor: `agent.type in (None, "agent")` is checked twice here -- once to decide whether to resolve, once to decide whether to include it in the payload. Naming it once removes the risk of the two checks drifting if that tuple ever changes in only one spot:

```suggestion
                        is_llm_agent = agent.type in (None, "agent")
                        resolved_agent = (
                            self._resolve_agent_working_dir(agent, agent_context)
                            if is_llm_agent
                            else agent
                        )

                        started_payload: dict[str, Any] = {
                            "agent_name": agent.name,
                            "iteration": agent_execution_count,
                            "agent_type": agent.type or "agent",
                            "context_window_max": await self._get_context_window_for_agent(
                                resolved_agent
                            ),
                        }
                        if is_llm_agent:
                            started_payload["working_dir"] = resolved_agent.working_dir
```

src/conductor/providers/copilot.py
```diff
@@ -851,15 +866,36 @@ async def _execute_sdk_call(
...
+                        "Skipping resume of Copilot session %s for agent '%s': "
+                        "working directory changed from %s to %s. Creating a new session.",
                         resume_sid,
-                        on_permission_request=self._default_permission_handler,
+                        agent.name,
+                        recorded_cwd,
+                        resolved_cwd,
                     )
+                    resume_sid = None
+            if resume_sid is not None:
```
jrob5756: Two things here:

1. `resume_sid` gets set back to `None` on the mismatch branch purely so the second `if resume_sid is not None:` on line 890 reads as "else, actually try to resume." Folding this into one `if/else` on the same condition would be easier to follow than two sequential checks of the same variable.
2. The `recorded_cwd is None` case (checkpoints saved before this PR) falls straight through to `resume_session` with zero logging. If someone resumes an old checkpoint after upgrading and adding `working_dir:`, the old session resumes under a cwd it may never have run in before, with nothing in the logs explaining that assumption. Worth an `logger.info` in that branch, e.g. right before the `try:` below:

```python
if recorded_cwd is None:
    logger.info(
        "Resuming Copilot session %s for agent '%s' with no recorded working "
        "directory (checkpoint predates working_dir tracking); assuming it "
        "matches resolved cwd %s.", resume_sid, agent.name, resolved_cwd,
    )
```

docs/workflow-syntax.md
```diff
@@ -272,6 +276,42 @@ agents:
...
+
+#### Symlink Semantics
+
+Because paths are normalized lexically instead of resolving to their real paths:
+- Different symlink aliases pointing to the same folder are treated as distinct paths.
+- For the Claude provider, distinct paths trigger separate MCP manager connections. This spawns separate MCP server subprocesses for each unique path alias.
+
+#### Key Restrictions and Exclusions
+
+- **Rejected Step Types:** The `working_dir` field is strictly rejected on `wait`, `set`, `terminate`, `human_gate`, and `workflow` (sub-workflow) step types. Defining `working_dir` on these steps raises a `ValidationError` at load time.
```
jrob5756, line 308: This lists rejected step types and implies every other step type resolves `working_dir` the same way the LLM-agent path does. `script` steps don't -- `ScriptExecutor` renders `agent.working_dir` on its own, with no `runtime.working_dir` fallback, no absolutize-against-the-workflow-file, and no pre-flight existence check (a missing directory surfaces as a raw subprocess error, not the `ExecutionError` the LLM path raises). A one-line callout here would keep someone who sets `runtime.working_dir` globally from assuming it also reaches `script` steps.

src/conductor/providers/claude.py
```diff
@@ -520,49 +556,66 @@ async def _ensure_mcp_connected(self) -> None:
                 "MCP servers configured but MCP SDK not installed. "
                 "Install with: uv add 'mcp>=1.0.0'"
             )
-            return
+            return None
 
-        self._mcp_manager = MCPManager()
+        async with self._mcp_manager_locks_guard:
+            lock = self._mcp_manager_locks.setdefault(resolved_cwd, asyncio.Lock())
```
jrob5756: `asyncio.Lock()` construction and `dict.setdefault()` here are both plain sync calls with no `await` inside, so nothing can interleave between them on a single-threaded event loop -- `_mcp_manager_locks_guard` isn't actually protecting a contested critical section. I checked this with a standalone repro (50 concurrent tasks across 3 cwds, a real `await asyncio.sleep()` inside "connect"): removing the guard and running `setdefault` unguarded still produces exactly one manager per cwd. If you drop it, `self._mcp_manager_locks_guard` and its declaring comment a few lines up can go too.

```suggestion
        lock = self._mcp_manager_locks.setdefault(resolved_cwd, asyncio.Lock())
```

jrob5756 (commented):
This is a great addition. Reviewed end to end: schema, validator, the engine resolver across linear/parallel/for_each, Copilot session/MCP stamping plus checkpoint resume, and Claude's per-cwd MCP pool. The precedence and resolution logic hold up consistently across all three execution paths. Found one bug and a couple of test gaps worth closing before merge -- inline comments below.

src/conductor/providers/claude.py
hertznsk: Good catch — fixed in 4849fbf. The manager is now pooled only when `has_servers()` returns true; a zero-connection manager is returned but not pooled, so the next agent for the same cwd retries the connect, and a warning is logged so the cwd isn't silently running with zero MCP tools. The fail-open test (`test_mcp_pool_connect_failure_fail_open_per_cwd`) was extended to prove the retry: it now uses a stateful fail-then-succeed fake and asserts the second call builds a new manager and reconnects.

tests/test_providers/test_claude.py
hertznsk: Fixed in 4849fbf. The fake `connect_server` now starts with `await asyncio.sleep(0)` so the four coroutines in `asyncio.gather` genuinely interleave inside the critical section, and the test additionally asserts exactly one connect sequence ran per cwd (`per_cwd_connects == ["/repo/a", "/repo/b"]`). The double-checked locking is now actually exercised.

tests/test_providers/test_claude.py
hertznsk: Fixed in 4849fbf, coupled with the cache-write fix above. The docstring's claim is now true: with the manager no longer pooled when zero servers connect, the test calls `_get_mcp_manager_for_cwd("/repo/bad")` a second time using a stateful fake (first attempt raises, second succeeds) and asserts a new manager instance was built and reconnected (`attempts["/repo/bad"] == 2`).

src/conductor/engine/workflow.py
hertznsk: Fixed in 79a57b4. Extracted `is_llm_agent = agent.type in (None, "agent")` once and reused it for both the resolve decision and the payload inclusion. Checked the rest of the file for the same double-check pattern — the call sites at ~4882 and ~5365 are single checks after an early `set`-step return, so no change needed there.

src/conductor/providers/copilot.py
hertznsk: Fixed in 178eecb. Restructured with a `should_resume` flag instead of mutating `resume_sid`, and added the `logger.info` for the `recorded_cwd is None` branch. I checked the installed SDK: `working_directory` is serialized into `payload["workingDirectory"]` for `resume_session` too, so the assumption is real — but since we can't prove how the runtime interprets it without a live session, the log says "checkpoint predates working_dir tracking, preserving legacy resume-by-id behavior" rather than claiming it matches. Added `test_legacy_checkpoint_resume_logs_info` covering resume-by-id, no `create_session` fallback, and the new log.

docs/workflow-syntax.md
hertznsk: Fixed in 3bd94ca — added a **Script Steps** bullet right after Rejected Step Types, documenting that script steps honor only their own `working_dir` (no `runtime.working_dir` fallback, relative paths resolve against the conductor process cwd, missing dirs surface as a subprocess startup error).

On the behavioral divergence itself: do you think it's worth a separate issue to decide whether `ScriptExecutor` should be brought to parity with the LLM-agent resolution (runtime fallback + workflow-file-relative absolutize + pre-flight existence check)? That's a behavior change beyond docs, and I'm happy to do the corresponding work if you and the maintainers want it.

src/conductor/providers/claude.py
hertznsk: Fixed in 4849fbf. Removed `_mcp_manager_locks_guard` (field + comment + usage); lock creation is now a bare `get`/`assign` with a comment noting the no-await invariant between the fast-path check and the per-cwd lock acquisition. The hardened race test (real `await asyncio.sleep(0)` yield in the fake) still produces exactly one manager per cwd, confirming the guard was dead weight. Also dropped the now-dead `provider._mcp_manager_locks_guard` fixture line in `test_claude_mcp_tool_filter.py`.

docs/workflow-syntax.md
jrob5756: This sounds like a good follow-up on its own -- please go ahead and open an issue for it, and thanks for offering to pick it up. I'd rather keep this PR scoped to LLM agents and their MCP servers as it stands.

jrob5756 (approved):
LGTM. Approved!

PR #306, opened 2026-07-16
feat(providers): force emit_output tool choice in Claude provider when no MCP tools

hertznsk:
## Summary

When an agent has an `output:` schema and the request carries no MCP tools, the Claude provider now sends `tool_choice={"type": "tool", "name": "emit_output", "disable_parallel_tool_use": true}` on every `messages.create` call. This aligns the implementation with the declared `CAPABILITIES.structured_output = "native"` contract (*"schema is enforced via a forced tool call"*) and eliminates an entire failure class where the model answers in prose and the provider falls into the expensive parse-recovery loop.

When MCP tools ARE present, no `tool_choice` is sent — the model keeps the freedom to call MCP tools first (the existing `tool_choice=auto` default).

When the Anthropic API rejects a forced `tool_choice` request (e.g. forced tool use is incompatible with extended thinking on the selected model), the provider now raises a targeted `ValidationError` naming the feature and suggesting the fix (remove `reasoning.effort` or switch model), instead of a generic `ProviderError`.

## Changes

- **`src/conductor/providers/claude.py`**
  - Compute `force_emit_output = has_output_schema and not mcp_tools` in `execute()` (by-fact-of-added-MCP-tools criterion; `mcp_tools` is initialized to `[]` before the conditional block so the no-MCP-manager path cannot raise `UnboundLocalError`).
  - Thread `force_emit_output` (default `False`, backward compatible) through `_execute_agentic_loop`, `_execute_with_parse_recovery`, and `_request_partial_output` into `_execute_api_call` (10 call sites).
  - Set `kwargs["tool_choice"]` in `_execute_api_call` before `messages.create` when the flag is set.
  - Extend the `BadRequestError` branch: when `force_emit_output` actually sent `tool_choice` and the normalized (lowercased) API message matches `"tool_choice"` or `"tool choice"`, raise `ValidationError` with a suggestion mentioning extended thinking. The pre-existing temperature-validation branch is unchanged.

- **`tests/test_providers/test_claude.py`** — 11 new tests:
  - `TestClaudeToolChoice` (8): happy path, MCP present (negative guard), MCP filter yielding no tools, raw output mode (negative guard), interrupt partial-output path, extended-thinking combo, retry persistence (both calls), no-MCP-manager regression.
  - `TestClaudeBadRequestToolChoice` (3): targeted ValidationError on rejection, `"tool choice"` (space) wording variant, negative guard locking the MCP path on generic `ProviderError`.
  - Each test carries a `# Requirement:` comment.

- **`docs/providers/claude.md`** — new *Structured Output* section covering the forced tool choice mechanism, MCP coexistence rule, and the extended-thinking interaction caveat.
- **`docs/providers/comparison.md`** — comparison table and Claude strengths list note the forced tool call.

## Test plan

- `uv run pytest tests/test_providers/test_claude.py -k tool_choice -v` → 11 passed
- `uv run pytest tests/test_providers/test_claude.py -k BadRequest -v` → 3 passed
- `uv run pytest tests/test_providers/` → 658 passed, 7 skipped
- `make test` → 3956 passed (2 failures reproduced on clean `main` — pre-existing, unrelated: checkpoint contamination test + Copilot 403)
- `make validate-examples` → green
- `make check` → clean (1 pre-existing `ty` warning in `engine/dialog_evaluator.py:131`, present on `main` too)
- Manual smoke (real `ClaudeProvider.execute` with mocked transport): happy path sends forced `tool_choice` and extracts content; rejection path raises `ValidationError` with `tool_choice` + `thinking` in the message; MCP path sends no `tool_choice`.

## Scope

Out of scope (unchanged): `copilot.py`, `hermes.py`, `claude_agent_sdk.py` (`tool_choice` is an Anthropic-API concept — no Provider Parity violation), parse-recovery logic (kept as a safety net for the MCP path), `OutputField`, the `emit_output` schema, MCP-tools path behavior.

🤖 Generated with [OhMyOpenCode](https://github.com/oh-my-opencode) orchestration

### DIFF
--- a/docs/providers/claude.md
+++ b/docs/providers/claude.md
@@ -11,6 +11,7 @@ The Claude provider enables Conductor workflows to use Anthropic's Claude models
 - [System Prompt](#system-prompt)
 - [Streaming Limitations](#streaming-limitations)
 - [Extended Thinking](#extended-thinking)
+- [Structured Output](#structured-output)
 - [Troubleshooting](#troubleshooting)
 - [Cost Optimization](#cost-optimization)
 
@@ -374,6 +375,34 @@ emits the same event shape so workflows that mix providers render consistently.
 
 See [`examples/reasoning-effort.yaml`](../../examples/reasoning-effort.yaml) for
 a runnable end-to-end example.
+
+## Structured Output
+
+The Claude provider enforces structured outputs natively by converting the `output` schema into an Anthropic tool definition.
+
+### Forced Tool Choice
+
+When you configure an agent with a defined `output` schema and don't declare any MCP tools, Conductor uses Anthropic's forced tool choice parameter. It sends the following configuration in the API request:
+
+```json
+{
+  "tool_choice": {
+    "type": "tool",
+    "name": "emit_output",
+    "disable_parallel_tool_use": true
+  }
+}
+```
+
+This forces the model to call the synthetic `emit_output` tool, which prevents the model from returning unstructured prose. Such enforcement aligns with the provider's capability contract declared as `CAPABILITIES.structured_output = "native"`.
+
+### Coexistence with MCP Tools
+
+If the agent has registered MCP tools, forced tool choice is disabled. The model needs to remain free to call any of the available MCP tools multiple times before concluding. In this case, Conductor instructs the model to return its final answer through the `emit_output` tool, but doesn't enforce it at the API parameter level.
+
+### Interaction with Extended Thinking
+
+Combining forced tool choice with Anthropic's extended thinking is not fully verified across all models. API support depends on Anthropic's backend validation rules. If the API rejects the request with a `400 BadRequestError` due to this combination, Conductor catches the error and raises a clear `ValidationError`. The error message will suggest adjusting the `reasoning.effort` setting or using a different model.
 
 ## Troubleshooting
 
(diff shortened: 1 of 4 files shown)


hertznsk:
Sorry to bother you. Before considering the possibility of merging this PR, I would like to perform one more check. It may turn out that this PR will only make it worse if users use the anthropic api through something like omniroute, choosing a model from an openai provider. I'll figure it out and let you know during the day.

hertznsk:
I tested this locally against an Anthropic-compatible omniroute endpoint, and unfortunately the PR breaks execution for OpenAI-provider models routed through that proxy.

Models checked:

- `kmc/kimi-k2.6` — without this PR it returns valid structured output via the existing parse-recovery path. With this PR it fails with a 400: `tool_choice 'specified' is incompatible with thinking enabled`.
- `kmc/kimi-k2.6-thinking` — same behavior: works without the PR, fails with the same BadRequestError when forced `tool_choice` is sent.

So the forced `tool_choice` is incompatible with the way omniroute/upstream enables extended thinking for these models. Without the PR everything works because the parse-recovery path still handles prose responses. This confirms the concern raised earlier — this PR makes things worse for users who run the Claude provider against non-Claude models through Anthropic-compatible proxies.

I’m going to close this PR and we won’t merge it as-is. If we want forced `tool_choice` later, it needs to be opt-in or have a fallback that retries without forced `tool_choice` when this specific 400 is returned.

hertznsk:
Closing as discussed — the forced tool_choice change causes regressions with Anthropic-compatible proxies routing non-Claude models.

PR #308, opened 2026-07-16
fix(web): clone subworkflowContexts tree in processEvent (#307)

### LINKED ISSUE #307: Web dashboard: agent node stuck showing 'running' after completion when adjacent to a sub-workflow step
## Summary

In the web dashboard, an agent node can remain visually stuck showing **"running" (blue border)** in the graph even though the agent has actually completed — confirmed via the detail panel, which correctly shows `COMPLETED` with full output/cost/token data. This has been observed specifically around transitions to/from a `type: workflow` (sub-workflow) step: the agent immediately *before* a sub-workflow step, or the return to a *parent* workflow after a sub-workflow completes.

Manually navigating into the sub-workflow node and back out immediately corrects the display (no data is actually lost — it's a pure rendering desync), which points to a React/Zustand reactivity bug rather than an engine-side data bug.

### Repro (as reported)

Workflow shape: `$start → architect (loops, "x2" shown) → document_review (type: workflow) → $end`.

1. Run a workflow where an agent (`architect`) is followed by a `type: workflow` step (`document_review`), with `architect` looping back at least once (iteration 2 shown in the report).
2. Observe the graph after `architect`'s second iteration completes and the workflow has moved on to `document_review`.
3. **Observed**: `architect`'s node border renders blue (`running`, `NODE_STATUS_HEX.running = #3b82f6`) — see attached screenshot. Clicking the node opens the detail panel, which correctly shows `COMPLETED`, iteration 2, full output, tokens, and cost.
4. **Workaround**: Double-click into the `document_review` sub-workflow panel, then navigate back out. `architect`'s border immediately renders green (`completed`) — see second screenshot. No new data appeared; the same underlying store data was already correct.

This has reportedly happened "a few times" with this particular workflow, always around a sub-workflow transition.

## Investigation notes

I couldn't spin up a live repro in this session, but traced the code paths involved and found a concrete correctness issue that's a strong suspect and should be fixed regardless of whether it's the exact trigger for this report:

### `processEvent` doesn't clone `subworkflowContexts` (or nested context state)

`src/conductor/web/frontend/src/stores/workflow-store.ts`:

```ts
processEvent: (event: WorkflowEvent) => {
  const handler = eventHandlers[event.type];
  set((state) => {
    const newState = { ...state, nodes: { ...state.nodes }, groupProgress: { ...state.groupProgress }, eventLog: [...state.eventLog], activityLog: [...state.activityLog], lastEventTime: event.timestamp };
    if (handler) {
      handler(newState, event.data, event.timestamp);
    }
    ...
    return newState;
  });
},
```

Every event handler call gets a **fresh shallow clone** of `nodes`/`groupProgress`/`eventLog`/`activityLog` so React/Zustand reference-equality checks correctly detect changes. `subworkflowContexts` is conspicuously **not** in that list — it's carried over by reference from the previous state, for the entire lifetime of a run.

Handlers routinely mutate it and its nested contexts **in place**:

- `subworkflow_started` (workflow-store.ts:1695-1765): `state.subworkflowContexts.push(ctx)` and `parent.children.push(ctx)` mutate the array in place; it is never replaced with a new array.
- `agent_started` / `agent_completed` / `agent_failed` / etc. route to a target via `activeTarget()` (workflow-store.ts:965-1013), which for a sub-workflow event returns `ctx.nodes`/`ctx.groupProgress` (nested inside the un-cloned `subworkflowContexts` tree) and mutates those objects directly (`ensureNode`/`replaceNode` reassign individual keys, but the containing `ctx.nodes` object and the `ctx` object itself never get a new reference).
- `subworkflow_completed` / `subworkflow_failed` mutate `targetNodes[data.agent_name]` in the same un-cloned tree.

Every other piece of node/progress state in this store follows a strict "always produce a new container reference so React can see the change" convention (see `replaceNode`'s own doc comment: *"Create a new reference for a node to ensure React/ReactFlow detects the change"*) — `subworkflowContexts` is the one exception, which breaks that invariant for any component/selector/memo that depends on it (directly or transitively, e.g. `useViewedGraphData()`/`useViewedNodes()` in `src/conductor/web/frontend/src/hooks/use-viewed-context.ts`, which include `subworkflowContexts`/`subContexts` in their `useMemo` dependency arrays).

In most cases this is masked because `state.nodes` (which *is* cloned every event) also changes reference on the same event, which is enough to force most subscribers to re-render and read the correct live value off the mutated objects. But it's exactly the kind of latent bug that can produce a transient desync depending on render/batching timing — matching the reported "stuck until you navigate away and back" symptom, since navigating (`viewContextPath` change) forces `WorkflowGraph`'s full rebuild effect (`src/conductor/web/frontend/src/components/graph/WorkflowGraph.tsx`, the `graphBuilt.current` reset on `viewPathKey` change), which reads fresh state and always shows the correct result.

## Suggested fix

Bring `subworkflowContexts` in line with the rest of the store's immutability convention in `processEvent` (and check `replayState`, which has similar mutate-through-loop semantics but commits once at the end so is lower risk): clone the `subworkflowContexts` array before handlers run, and make sure any handler that mutates a nested `ctx` (nodes, groupProgress, routes, highlightedEdges, children, etc.) does so on a fresh copy rather than the shared object graph — mirroring the existing `replaceNode` pattern used for individual nodes.

## Additional finding: zero frontend test coverage

There are currently **no frontend automated tests** (`src/conductor/web/frontend` has no `*.test.*`/`*.spec.*` files outside `node_modules`). This class of store-reactivity bug is exactly what a focused Vitest suite around `workflow-store.ts`'s `processEvent` (feeding a synthetic event sequence: `agent_started` → `agent_completed` (iteration 2) → `subworkflow_started` → `subworkflow_completed`, then asserting on `state.nodes`/`state.subworkflowContexts` reference identity and status) would catch. Recommend adding a minimal Vitest setup as part of the fix for this issue.

## Environment

- Reported by @jrob5756 with dashboard screenshots showing the `architect` agent node bordered blue ("running") while its own detail panel shows `COMPLETED` (iteration 2, `claude-opus-4.8`, 130.5K tokens, $0.7757), immediately followed by a second screenshot (after navigating into/out of the adjacent `document_review` sub-workflow node) showing the same node correctly bordered green ("completed") with no other data changes.

jrob5756:
## Summary

Fixes #307: an agent node adjacent to a `type: workflow` sub-workflow step
could render stuck showing "running" even though the underlying store data
was already `completed` — a pure rendering/reactivity desync, not an engine
data bug (confirmed correct via the detail panel per the issue report).

## Root cause

`processEvent` in `workflow-store.ts` clones `nodes`/`groupProgress`/
`eventLog`/`activityLog` on every event so React/Zustand reference-equality
checks pick up changes (per `replaceNode`'s own doc comment: *"Create a new
reference for a node to ensure React/ReactFlow detects the change"*) — but
it never cloned `subworkflowContexts`. Handlers reached through
`activeTarget()`/`resolveContext()`/`resolveSlotPath()` (`subworkflow_started`,
`subworkflow_completed`, agent events carrying a `subworkflow_path`, etc.)
mutate nested `ctx.nodes`/`groupProgress`/`routes`/`highlightedEdges` in
place, and `subworkflow_started` pushes onto `subworkflowContexts`/
`children` arrays in place. Nothing forced a fresh reference anywhere in
that tree, so any selector/`useMemo` keyed on `subworkflowContexts` (e.g.
`useViewedNodes`/`useViewedGraphData` in `use-viewed-context.ts`) could miss
an update depending on render/batching timing.

## Fix

- Added `cloneSubworkflowContexts()`: a structural clone of the
  `SubworkflowContext` tree (each context plus its own `nodes`/
  `groupProgress`/`routes`/`highlightedEdges`/`children`/etc.), wired into
  `processEvent` alongside the existing clones.
- Audited `replayState`: it already rebuilds `subworkflowContexts` fresh
  from `{}` per replay and commits once at the end, so it needed no change.

## Testing

There were previously **zero** frontend automated tests. Added a minimal
Vitest setup (`vitest` devDependency, `vitest.config.ts`, `npm run test`,
and a `make test-frontend` target) plus a regression test in
`workflow-store.test.ts` that:

- Simulates the reported repro (an agent looping back once, then an
  adjacent `subworkflow_started`/`subworkflow_completed` pair) and asserts
  the agent's node stays `completed` and `subworkflowContexts` gets a fresh
  reference on every mutating event.
- Covers a second case: a node update nested *inside* a running child
  sub-workflow context (via `subworkflow_path`), asserting the same
  reference-identity behavior.
- **Verified fails without the fix**: temporarily reverted the
  `cloneSubworkflowContexts` call and confirmed both tests fail with the
  exact `Object.is` reference-equality assertion; restored the fix and
  confirmed both pass again.

```
make test-frontend   # 2 passed
npx tsc -b            # clean
uv run ruff check src tests   # All checks passed!
uv run ty check src   # 1 pre-existing warning (dialog_evaluator.py), unrelated
```

Closes #307.


jrob5756:
Ran a multi-agent code review (code-reviewer, pr-test-analyzer, comment-analyzer) against the original commit and addressed the findings in a follow-up commit:

- **Performance/re-render fix**: replaced the whole-tree clone of `subworkflowContexts` with a path-scoped clone-on-write (`resolveMutableContext`/`resolveMutableSlotPath`). Only the mutated branch's ancestor chain gets a fresh reference now; untouched siblings (and root-level events with no active sub-workflow) keep their existing reference, so this no longer forces every `subworkflowContexts`-dependent component to re-render on unrelated events.
- **New regression tests**: nested (grandchild) sub-workflow propagation, concurrent for_each siblings not cross-cloning each other, root-only events never touching `subworkflowContexts`, and field preservation across a clone-and-mutate cycle. Verified each fails when the corresponding part of the fix is reverted.
- **CI**: added a `frontend` job (typecheck + Vitest + build) so this regression suite runs on every PR instead of relying on someone remembering `make test-frontend`.
- **Comment hygiene**: trimmed doc-comment cross-references prone to drift (pointing at another comment, naming specific hook consumers, citing the issue number as load-bearing).

All 6 frontend tests pass, `tsc -b` is clean, and `ruff check` / `ty check` are unaffected.

PR #313, opened 2026-07-17
feat(providers): add runtime.tool_output limits for oversized MCP tool results

hertznsk:
## Summary

Workflows that call chatty MCP tools (web scrapers, code search, Git hosts) previously died with a fatal `request exceeded model token limit` API error when a single tool result grew too large. This PR adds a configurable per-result cap: oversized tool results are truncated to a configurable size, the full text is spilled to a temp file the agent can page through, and a notice surfaces in the console, event log, and dashboard whenever a result is capped.

**Provider behavior:**
- **Claude** — Conductor truncates the result in `MCPManager.call_tool` (the single choke point through which every MCP result passes) before it reaches the API. The full text is written to a process-private spill file (mode `0o600`, dir `0o700`, `O_EXCL` create). A single-line marker is appended with a generic hint by default; when the agent has filesystem-like tools, the hint is rewritten to point at the spill file.
- **Copilot** — The Copilot SDK already has a native spill-to-file feature, so Conductor just forwards the new config to `create_session` / `resume_session` / dialog sessions as `large_output` instead of re-inventing it.
- **claude-agent-sdk** — Managed by the native `MAX_MCP_OUTPUT_TOKENS` env var on the `claude` CLI (outside this config).
- **Hermes** — No MCP tools, N/A.

**Event:** `agent_tool_output_truncated` (tool_name, original_chars, kept_chars, spill_path) is emitted **Claude-only** (the Copilot SDK owns its tool loop and exposes no hook). It renders in the console (verbose), the JSONL event log, and the dashboard activity feed.

## Configuration

New `runtime.tool_output` block (opt-out, safe defaults — existing workflows behave identically unless a result exceeds the cap):

```yaml
workflow:
  runtime:
    tool_output:
      enabled: true          # master switch
      max_chars: 50000       # per-result cap (min 1000)
      spill_to_file: true    # write full output to a temp file
      spill_dir: null        # default: <tempdir>/conductor/tool-output
```

See [`examples/tool-output-limits.yaml`](examples/tool-output-limits.yaml).

## Important semantics

- **Per-result cap, not a cumulative context budget.** Each tool result is evaluated independently. Many truncated results + prompt/history can still exceed the window — tune via `max_chars` / `max_agent_iterations`. Cumulative budgeting is out of scope.
- **Copilot:** `max_chars` is forwarded to the SDK as **bytes**, so multibyte UTF-8 (CJK/emoji) may truncate earlier than `max_chars` characters.
- **Spill files** contain raw tool output (may include secrets) and are not deleted by Conductor (OS temp directory cleanup).

## What this does NOT do

- No LLM summarization of tool output
- No tokenizer dependency (chars, not tokens)
- No changes to `engine/context.py` trim logic
- No behavior change to `claude-agent-sdk` or `hermes`
- No per-agent `tool_output` override
- `MCPManager.call_tool` signature unchanged (still returns `str`)

## Testing

- `make test` — 4212 passed (1 pre-existing live-API failure on `main`: `test_large_create_tool_call_does_not_truncate` requires Copilot enterprise auth, unrelated to this PR)
- `make check` (ruff + ty) — clean
- `make validate-examples` — clean
- `make build-frontend` — clean

New/updated tests cover: schema defaults/validation, truncation + spill (filename sanitization, `0o600`/`0o700`, UTF-8 round-trip, `O_EXCL`, symlink rejection for explicit/default/ancestor dirs, chmod hardening, fd-leak on `fdopen` failure, Unicode cleanup), hint rewrite gating (fs-tools + real spill path required), marker parsing (both hint variants, long paths, dot-space paths, payload-literal spoofing, rightmost-marker), event emission (exactly once per truncated call), console subscriber branch, Copilot `large_output` forwarding for all 3 session sites (including `enabled=False` forwarding since the SDK defaults to enabled), and dashboard store/type wiring.

## Docs

- `docs/mcp-tools.md` — new "Tool output limits" section + Provider Support table row
- `docs/workflow-syntax.md` — `runtime.tool_output` reference
- `docs/providers/comparison.md` — capability table row
- `AGENTS.md` — Key Patterns entry
- `examples/tool-output-limits.yaml` — new example


src/conductor/mcp/manager.py
```diff
@@ -221,13 +275,168 @@ async def call_tool(
...
+            full_text: The full tool result text to persist.
+            server_name: Name of the MCP server that produced the result.
+            original_name: Original tool name without the server prefix.
+
+        Returns:
+            The absolute path of the spill file, or None if writing failed.
+        """
+        try:
+            spill_dir_str = self._tool_output.spill_dir
+            if spill_dir_str:
```
jrob5756, line 383: This walks the raw, unresolved `spill_dir` for symlinks, but `/tmp` and `/var/tmp` are themselves symlinks on macOS (to `/private/tmp` and `/private/var/tmp`). So a totally ordinary `spill_dir: /tmp/my-dir` trips this check and the feature quietly falls back to a pathless marker on every call, on the exact platform most of us are developing on.

The default-dir branch below doesn't have this problem because it resolves before walking ancestors. Same fix should work here.

```suggestion
            if spill_dir_str:
                spill_dir = Path(spill_dir_str).resolve()
                if _contains_symlink(spill_dir, stop_at=spill_dir.parent.parent if spill_dir.parent != spill_dir else None):
```

src/conductor/mcp/manager.py
```diff
@@ -221,13 +275,168 @@ async def call_tool(
             if not response_text and result.structuredContent:
                 response_text = str(result.structuredContent)
 
+            response_text = self._maybe_truncate_response(
```
jrob5756: `_maybe_truncate_response`/`_spill_full_output` are documented as best-effort and never-raise, but this call sits inside `call_tool`'s outer `except Exception`. If a bug ever slips into the truncation path, it gets caught here, logged as `"MCP tool call failed"`, and re-raised as `RuntimeError` — even though the tool call itself already succeeded. Whoever debugs that later is going to go looking at the wrong server.

```suggestion
            try:
                response_text = self._maybe_truncate_response(
                    response_text,
                    server_name=server_name,
                    original_name=original_name,
                )
            except Exception as truncation_err:
                logger.warning(
                    "Failed to apply output truncation for %s; returning untruncated result: %s",
                    prefixed_name,
                    truncation_err,
                )
```

src/conductor/config/schema.py
```diff
@@ -1959,6 +1959,51 @@ def is_enabled(self) -> bool:
...
+
+    spill_to_file: bool = True
+    """Whether oversized results should be written to a temporary file.
+
+    When ``True``, the full result is persisted to disk and a marker containing
+    the file path is appended to the truncated prefix. When ``False``, no file is
+    created and truncation is applied in-place.
+    """
+
+    spill_dir: str | None = None
```
jrob5756, line 2070: No normalization on this field, and the two consumers don't agree on what an empty string means: `manager.py` treats `""` as falsy and falls back to the default temp dir, `copilot.py` checks `is not None` and forwards `output_directory=""` straight to the SDK. Same config value, two different behaviors depending on which provider you're running. A validator that turns `""` into `None` here fixes both call sites at once.

```suggestion
    spill_dir: str | None = None

    @field_validator("spill_dir")
    @classmethod
    def _normalize_spill_dir(cls, v: str | None) -> str | None:
        return v.strip() or None if v is not None else None
```

src/conductor/providers/claude.py
```diff
@@ -24,13 +24,22 @@
...
 import time
-from typing import TYPE_CHECKING, Any, Protocol, get_args
+from typing import Any, Protocol, get_args
 
 from pydantic import BaseModel
 
+from conductor.config.schema import AgentDef, OutputField, ToolOutputConfig
 from conductor.exceptions import ProviderError, ValidationError
 from conductor.executor.output import validate_output
+from conductor.mcp.manager import (
```
jrob5756, line 36: `manager.py`'s own docstring says the marker is "generated entirely inside this manager," and then this file imports four underscore-prefixed constants from it directly. Nothing in ruff or ty will catch someone renaming these during a `manager.py` refactor, since private-name imports aren't a flagged contract violation, so this can break silently. Promoting `_FS_HINT`, `_GENERIC_HINT`, `_TAIL_WINDOW`, and `_TRUNCATION_MARKER_PREFIX` to public names (or their own small module) would make the dependency explicit instead of accidental.

src/conductor/providers/copilot.py
```diff
@@ -332,6 +338,41 @@ def _default_permission_handler(
         logger.debug("auto-approved permission request: %s", request)
         return PermissionHandler.approve_all(request, invocation)
 
+    def _large_output_config(self) -> dict[str, Any] | None:
```
jrob5756: Typed `-> dict[str, Any] | None`, and all three call sites guard with `if large_output_cfg is not None:`, but every branch here returns a dict — there's no `None` path anywhere in the method. Looks like a leftover from an earlier version where `None` meant "skip this SDK kwarg entirely." Either drop the `| None` and the three now-dead checks, or bring back whatever case was supposed to return `None`.

src/conductor/providers/copilot.py
```diff
@@ -332,6 +338,41 @@ def _default_permission_handler(
...
+        The SDK has no "truncate in-place without spilling" mode. Therefore
+        ``spill_to_file=False`` is mapped to ``{"enabled": False}``, which
+        disables large-output handling entirely for the session.
+
+        ``output_directory`` is only included when ``spill_dir`` is explicitly
+        configured; otherwise the SDK uses its own OS-level temporary
+        directory.
+        """
+        if not self._tool_output_config.enabled:
+            return {"enabled": False}
```
jrob5756, line 377: Worth calling out how different this is from Claude's behavior: on Claude, `spill_to_file=False` still truncates to `max_chars`, it just skips the file write. Here it disables output-size limiting entirely, because the SDK has no truncate-without-spill mode. Someone sets this to keep tool output off disk and ends up with unlimited tool output instead, with nothing at runtime telling them that happened — it's only in the docstring. A log line when this branch fires would close that gap cheaply.

src/conductor/providers/claude.py
```diff
@@ -2474,6 +2504,133 @@ def _extract_text_content(self, response: Any) -> dict[str, Any]:
...
+
+        The check is heuristic: after stripping the ``server__`` prefix, the
+        lowercase tool name is matched for common filesystem/shell substrings.
+        A None tool list means no tools were allowed, so filesystem-like tools
+        are not available.
+        """
+        if not tools:
+            return False
+
+        fs_names = (
```
jrob5756: Substring containment, not word boundaries, so any tool name that happens to contain `"ls"` or `"file"` trips this and gets the filesystem-hint rewrite whether or not it can actually read a file off disk. Telling the model to go use "your filesystem tools" on a false positive just sends it down a dead end. `re.search(r"\b(read|grep|glob|bash|shell|ls|find|view|edit)\b", lower)` would tighten this without much extra code.

docs/providers/comparison.md
```diff
@@ -22,6 +22,7 @@ This guide helps you choose between GitHub Copilot, Anthropic Claude, Claude Age
 | **Agentic Loop** | SDK-managed | Manual (provider code) | SDK-managed (delegated to CLI) | SDK-managed (delegated to hermes) |
 | **Structured Output** | Prompt injection | Native | Prompt injection | Prompt injection |
 | **Session Resume** | Yes | No | No | Yes |
+| **Tool Output Limits** | native SDK spill+compaction | conductor-side truncation+spill | native CLI env var | N/A |
```
jrob5756: "native SDK spill+compaction" reads like tool-output limiting also triggers context compaction, but it doesn't — `large_output` (what this PR forwards) only handles spill-to-file, and compaction is a separate SDK mechanism for the session's context window that `_large_output_config` never touches. Worth splitting those apart so nobody assumes a connection that isn't there.

```suggestion
| **Tool Output Limits** | native SDK spill (large_output) | conductor-side truncation+spill | native CLI env var | N/A |
```

jrob5756 (commented):
Thanks for the contribution. I reviewed and have some inline comments. Thanks!

src/conductor/mcp/manager.py
hertznsk: Fixed, thanks. The explicit `spill_dir` branch now resolves the path and applies a symlink policy: a symlink is only rejected when the resolved path lands **outside** the system temp root. That keeps the macOS layout working (`/tmp` and `/var/tmp` are themselves symlinks into `/private/...`) while still refusing a symlink that would redirect spilled output to an attacker-chosen location elsewhere. I kept the check on the raw path for the outside-temp case so the security guarantees the existing tests pin (symlinked leaf, symlink mid-path) still hold. Covered by `test_explicit_spill_dir_under_symlinked_parent_is_allowed` plus the updated symlink-rejection tests.

src/conductor/mcp/manager.py
hertznsk: Good catch. `_maybe_truncate_response` now runs in its own `try/except` inside `call_tool`, separate from the tool call's outer handler. A failure in the best-effort truncation path is logged as a warning (`Failed to apply output truncation for ...`) and the untruncated result is returned, instead of being re-raised as `RuntimeError("MCP tool call failed")` and pointing whoever debugs it at the wrong server. Covered by `test_truncation_failure_does_not_fail_successful_tool_call`.

src/conductor/config/schema.py
hertznsk: Done. `ToolOutputConfig.spill_dir` now has a `field_validator` that normalizes empty/whitespace strings to `None`, so both consumers see the same value: the MCP manager falls back to the default temp dir and the Copilot provider no longer forwards `output_directory=""` to the SDK. Covered by `test_spill_dir_empty_string_normalizes_to_none`.

src/conductor/providers/claude.py
hertznsk: Agreed — promoted the four constants to public names (`TRUNCATION_MARKER_PREFIX`, `GENERIC_HINT`, `FS_HINT`, `TAIL_WINDOW`) since they are the shared contract with ClaudeProvider's marker parser, and updated the import here (plus the two test files that referenced them) to match. The dependency is now explicit rather than a private-name import that a manager-side rename could silently break.

src/conductor/providers/copilot.py
hertznsk: Right, it was a leftover. Dropped the `| None` from the return type and removed the three now-dead `is not None` guards at the call sites — every branch returns a dict, so `large_output` is now forwarded unconditionally.

src/conductor/providers/copilot.py
hertznsk: Good point — added a `logger.warning` when this branch fires, spelling out that `spill_to_file=False` disables tool output size limiting entirely on Copilot (the SDK has no truncate-without-spill mode) and that output will not be capped at `max_chars`, unlike Claude which still truncates in place. Also clarified the `spill_to_file` field description in `docs/mcp-tools.md` to spell out the provider difference. Covered by `test_spill_to_file_false_logs_a_warning`.

src/conductor/providers/claude.py
hertznsk: Tightened. `_has_fs_like_tool` now splits the tool name on non-alphanumeric characters and requires a keyword to equal a whole segment, so `translate`/`fileupload` no longer match. I also narrowed the keyword set to read-by-path capability (`read, grep, view, cat, open, load, file, bash, shell`) and dropped the search/write keywords (`find`, `ls`, `glob`, `edit`) — the agent already has the exact spill path from the marker, so it only needs to read/grep by path, and hinting at search or edit tools would send it down the same kind of dead end you flagged. Calibrated the set against real MCP server tool names (`@modelcontextprotocol/server-filesystem`, git, github, bash servers) so common readers like `read_file`, `read_multiple_files`, `view_code` and `bash` still match.

docs/providers/comparison.md
hertznsk: Applied — split the two apart so the row now reads `native SDK spill (large_output)`, making clear that what this PR forwards only handles spill-to-file and is unrelated to the SDK's separate context-window compaction.

jrob5756 (approved):
LGTM. Approved!

PR #317, opened 2026-07-20
fix(providers): align Copilot and Hermes prompt schemas on the shared recursive builder

hertznsk:
## Summary

Builds on #311, which extracted the shared output-schema builders into `_schema.py`. This PR flips the two prompt-injection providers (Copilot and Hermes) onto the single shared recursive builder and deletes the legacy Hermes-only code path — one builder, one behavior, no dead code.

Two parity fixes:

1. **Copilot and Hermes prompt schemas no longer fall back to `The {field} field` for fields without an explicit `description`.** The shared schema builder already omitted the fallback for Claude; Copilot and Hermes now use the same behavior, so an absent `description` produces a JSON Schema object without a synthetic description. This keeps prompts shorter and consistent across providers.

2. **Hermes now uses the shared recursive prompt-schema builder** (the same one Copilot uses). Previously Hermes used its own hand-rolled builder that only recursed through object properties and collapsed nested arrays. It now delegates to the shared recursive builder, which correctly handles scalar types, arrays of arrays, and arbitrary nesting: `array<object>` items include both `properties` AND `required`, `array<array>` items recurse, and explicit descriptions survive at every nesting level.

Deleted: `build_hermes_legacy_prompt_schema`, `_build_hermes_legacy_item_schema`, and the `description_fallback` / `field_name` parameters from the shared prompt-schema builders. Claude and Claude Agent SDK are untouched (already aligned, JSON-Schema flavor).

## Migration note

Hermes prompt schemas now use the shared recursive builder. The shared depth limit now applies to every array item: scalar/array-item nesting that legacy Hermes never depth-checked at all (it only checked object-item properties, and collapsed array-of-arrays) may now raise `ValidationError: Schema nesting depth exceeds maximum of 10 levels` at depth >10; reduce nesting or split the schema. If model adherence degrades on fields without descriptions, add an explicit `description`.

## Test plan

- TDD two-phase: focused semantic tests pinned the new behavior red first (no fallback, `required` in `array<object>` items, recursive `array<array>`, exact depth-10/11 boundary counting), then the implementation, then golden literals regenerated from actual builder output.
- Golden literals for Copilot/Hermes updated and structurally verified; Claude/SDK literals byte-identical to main.
- `TestGoldenMutationGuard` retargeted to mutate an explicit description (its old mutation target disappeared with the fallback).
- Mutation checks: restoring the pre-PR source makes the new tests fail; the legacy builder import no longer exists.
- `make check` ✅ `make validate-examples` ✅ full suite green except one pre-existing environmental failure (`test_large_create_tool_call_does_not_truncate`, requires a Copilot enterprise entitlement — untouched by this PR).

### DIFF
--- a/src/conductor/providers/_schema.py
+++ b/src/conductor/providers/_schema.py
@@ -107,26 +107,18 @@ def build_json_schema_properties(
 def build_prompt_schema_field(
     field: OutputField,
     *,
-    field_name: str | None,
     depth: int = 0,
     max_depth: int = 10,
-    description_fallback: bool = False,
 ) -> dict[str, Any]:
     """Build a prompt-facing schema fragment for a single ``OutputField``.
 
-    When ``description_fallback`` is true and the field has no explicit
-    description and ``field_name`` is provided, the description is filled
-    with ``"The {field_name} field"``. Array items are built with
-    ``field_name=None`` so they do not gain a fallback description.
+    The fragment contains ``type`` and optionally ``description``,
+    ``properties`` + ``required`` (for objects), or ``items`` (for arrays).
 
     Args:
         field: The output field definition.
-        field_name: The field name used for the fallback description, or
-            ``None`` to suppress the fallback.
         depth: Current nesting depth.
         max_depth: Maximum allowed nesting depth.
-        description_fallback: Whether to synthesize a description when one
-            is not explicitly set.
 
     Returns:
         A prompt-facing schema fragment dictionary.
@@ -136,30 +128,19 @@ def build_prompt_schema_field(
     """
     _check_depth(depth, max_depth)
 
-    description = field.description
-    if description_fallback and description is None and field_name is not None:
-        description = f"The {field_name} field"
-
     schema: dict[str, Any] = {"type": field.type}
-    if description:
-        schema["description"] = description
+    if field.description:
+        schema["description"] = field.description
 
     if field.type == "object" and field.properties:
         schema["properties"] = build_prompt_schema_properties(
-            field.properties,
-            depth=depth + 1,
-            max_depth=max_depth,
-            description_fallback=description_fallback,
+            field.properties, depth=depth + 1, max_depth=max_depth
         )
         schema["required"] = list(field.properties.keys())
 
     if field.type == "array" and field.items:
         schema["items"] = build_prompt_schema_field(
-            field.items,
-            field_name=None,
-            depth=depth + 1,
-            max_depth=max_depth,
-            description_fallback=description_fallback,
+            field.items, depth=depth + 1, max_depth=max_depth
         )
 
     return schema
@@ -170,7 +151,6 @@ def build_prompt_schema_properties(
     *,
     depth: int = 0,
     max_depth: int = 10,
-    description_fallback: bool = False,
 ) -> dict[str, Any]:
     """Build a prompt-facing schema mapping from named ``OutputField`` definitions.
 
@@ -178,8 +158,6 @@ def build_prompt_schema_properties(
         fields: Mapping from field name to output field definition.
         depth: Current nesting depth.
         max_depth: Maximum allowed nesting depth.
-        description_fallback: Whether to synthesize a description when one
-            is not explicitly set.
 
     Returns:
         A prompt-facing schema mapping.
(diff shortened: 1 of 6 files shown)


src/conductor/providers/_schema.py
```diff
@@ -107,26 +107,18 @@ def build_json_schema_properties(
 def build_prompt_schema_field(
```
jrob5756, line 107: Now that `field_name`/`description_fallback` are gone, `build_prompt_schema_field`/`build_prompt_schema_properties` are identical to `build_json_schema_field`/`build_json_schema_properties` above (lines 45-104) - same depth check, same shape, same recursion. Only two call sites are left (`copilot.py:1658`, `hermes.py:652`). Given the PR title is literally about aligning on one shared builder, it seems worth finishing the job: drop this pair and point both callers at `build_json_schema_properties` instead. Not asking for it in this PR necessarily, just flagging since the diff makes the duplication obvious.

tests/test_providers/test_hermes.py
```diff
@@ -919,34 +921,169 @@ def chain_in_array_item(levels: int) -> dict[str, OutputField]:
...
+            for _ in range(levels):
+                leaf = OutputField(type="array", items=leaf)
+            return leaf
+
+        _build_prompt_schema({"matrix": nested_array_chain(_MAX_SCHEMA_DEPTH)})
+        with pytest.raises(ValidationError, match="exceeds maximum"):
+            _build_prompt_schema({"matrix": nested_array_chain(_MAX_SCHEMA_DEPTH + 1)})
+
+
+class TestHermesSharedPromptSchema:
```
jrob5756: This new class and `TestHermesBuildPromptSchema` above (line 839) cover almost the same ground - same `_chain_schema` helper copy-pasted, and 4-5 of the 6 tests are near-identical bodies (no-fallback description, array<object> required, array-of-arrays recursion, the depth-8/9 boundary). The class docstring even calls out that it's "complementing" the other suite, so this looks intentional rather than an oversight, but it means two places to update if this behavior ever changes again. Worth folding into one class and keeping only the two tests that are actually unique (`test_build_prompt_schema_exceeds_max_depth`'s message/suggestion pin, and `test_nested_array_object_descriptions_preserved`).

tests/test_providers/test_output_schema.py
```diff
@@ -794,12 +787,12 @@ class TestGoldenMutationGuard:
...
+        must fail so regressions are caught."""
         provider = _make_copilot_provider()
         actual = _serialize(provider._build_prompt_schema(rich_schema))
         mutated = EXPECTED_COPILOT_RICH_SCHEMA.replace(
-            '"description": "The number_scalar field"',
-            '"description": "The number_scalar field (mutated)"',
+            '"description": "A string scalar field with a description"',
+            '"description": "A string scalar field with a description (mutated)"',
         )
         assert actual != mutated
```
jrob5756: `mutated` here is built by string-replacing the *expected* literal, not by touching the real builder or its output. So `actual != mutated` will always be true, no matter what `_build_prompt_schema` actually does - it's checking that two different strings are different, not that the golden test would catch a regression. As written it doesn't add coverage beyond what `TestCopilotOutputSchemaGolden` already gives you.

If the goal is a real mutation guard, it'd need to run the builder against a schema that's deliberately different from `rich_schema` and confirm *that* output no longer matches the golden string, e.g.:

```suggestion
    def test_golden_assertion_detects_missing_description(self) -> None:
        """A schema whose only difference from `rich_schema` is a dropped description
        must not satisfy the rich-schema golden assertion."""
        provider = _make_copilot_provider()
        schema_without_description = {
            **rich_schema,
            "string_scalar": OutputField(type="string"),
        }
        actual = _serialize(provider._build_prompt_schema(schema_without_description))
        assert actual != EXPECTED_COPILOT_RICH_SCHEMA
```

Or just drop the class if that's more effort than it's worth for what it buys you.

tests/test_providers/test_hermes.py
```diff
@@ -839,10 +839,12 @@ def test_real_capabilities_tuple_excludes_max(self) -> None:
 class TestHermesBuildPromptSchema:
     """Tests for the Hermes prompt-schema builder wrapper.
 
-    The wrapper delegates to conductor.providers._schema.build_hermes_legacy_prompt_schema
-    and converts SchemaDepthError into ValidationError with the exact legacy message and
-    suggestion. These tests pin the legacy behavior (description fallback, no required
-    inside array-item objects, collapsed array-of-array items) and the depth boundary.
+    The wrapper delegates to the shared recursive builder in
+    conductor.providers._schema and converts SchemaDepthError into ValidationError
+    with the exact legacy message and suggestion. These tests pin the shared
```
jrob5756: "legacy" is stale here - there's no legacy-specific message anymore, `hermes.py:648` was already updated in this same PR to drop the word. Small thing, but it'll read as if a separate legacy format still exists.

```suggestion
    with the exact message and suggestion. These tests pin the shared
```

tests/test_providers/test_output_schema.py
```diff
@@ -752,7 +743,8 @@ def test_build_prompt_schema_rich_matches_baseline(self) -> None:
 
     def test_build_prompt_schema_missing_descriptions_matches_baseline(self) -> None:
         """Copilot prompt schema wrapper must be byte-for-byte identical to the pre-refactor
```
jrob5756: This still claims byte-for-byte identity with the "pre-refactor baseline," but the whole point of this test is that the output *changed* from that baseline (the fallback descriptions are gone). Might be clearer as something like "must match the current prompt schema output for a schema without descriptions - fields without an explicit description emit no description key," so a future reader doesn't assume this is a frozen behavior-preservation check.

tests/test_providers/test_output_schema.py
```diff
@@ -763,13 +755,14 @@ class TestHermesOutputSchemaGolden:
 
     def test_build_prompt_schema_rich_matches_baseline(self) -> None:
         """Hermes prompt schema wrapper must be byte-for-byte identical to the pre-refactor
```
jrob5756: Same issue as the Copilot one above (line 745) - this and the missing-descriptions test right below it (line 763) both say "identical to the pre-refactor baseline" while describing exactly the behavior this PR changed (recursive array-of-arrays, no fallback description). Worth rewording both to describe current behavior rather than parity with the old implementation.

jrob5756 (commented):
Thanks for contributing once again! Everything looks great, nothing really blocking. But a few things to look at. The comments below are mostly cleanup you exposed by finishing the job so thoroughly. If you take a look I can merge right away!

hertznsk:
Thank you for your thorough review! I will make corrections tomorrow or the day after tomorrow.

src/conductor/providers/_schema.py
hertznsk: Good catch - they're identical right now, but deliberately so, and I'd like to keep the pair. I'm planning a follow-up PR that adds `additionalProperties: false` to the **json_schema** flavor only (Claude's `emit_output` tool schema and claude-agent-sdk's `output_format`), where the API actually enforces it. The prompt flavor is serialized as text into Copilot/Hermes prompts, where the extra key would just add noise with no enforcement value - so the two flavors will intentionally diverge, and collapsing them now would mean re-splitting them in that follow-up. The `build_prompt_schema_*` names also document the intent at the two call sites. Happy to revisit the structure in that follow-up if you'd prefer something like one parameterized private builder with two thin public wrappers.

tests/test_providers/test_hermes.py
hertznsk: Done - folded into `TestHermesBuildPromptSchema`. Kept the two unique tests you flagged (the exact message/suggestion pin already lived in the surviving class, and `test_nested_array_object_descriptions_preserved` moved over); the near-identical bodies and the duplicated `_chain_schema` helper are gone.

tests/test_providers/test_output_schema.py
hertznsk: Rewritten as you suggested - the guard now runs the builder against a schema whose only difference from `rich_schema` is a dropped description and asserts the output no longer matches `EXPECTED_COPILOT_RICH_SCHEMA`. Much better as a real negative test, thanks for the snippet.

tests/test_providers/test_hermes.py
hertznsk: Fixed - dropped "legacy" here and in `test_build_prompt_schema_exceeds_max_depth` below; both now say "shared".

tests/test_providers/test_output_schema.py
hertznsk: Reworded - the docstring now describes the current behavior (no description key for fields without an explicit description) instead of claiming identity with the old baseline. Also caught the same stale claim in the module docstring and the rich-schema test.

tests/test_providers/test_output_schema.py
hertznsk: Both reworded to describe current behavior rather than parity with the old implementation.

jrob5756 (approved):
LGTM. Approved!

PR #325, opened 2026-07-21
Honor configured session timeout for Copilot dialog turns

PolyphonyRequiem:
## Summary
- replace the hard-coded 120-second Copilot dialog wait with the provider's configured `max_session_seconds`
- classify dialog timeout as retryable so callers can continue the existing dialog
- add regression coverage and an Unreleased changelog entry

## Problem

`CopilotProvider.execute_dialog_turn()` ignored `runtime.max_session_seconds` and always timed out after 120 seconds. Long reasoning responses could therefore fail even when the workflow explicitly allowed a longer session.

## Validation

- `PYTHONPATH= ./.venv/Scripts/pytest.exe tests/test_providers/test_copilot.py` — 128 passed
- Ruff check and format check passed for changed files
- `ty check src/conductor/providers/copilot.py` passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/microsoft/conductor/compare/v0.1.23...HEAD)
 
+### Fixed
+
+- **Copilot dialog turns no longer fail after a hidden 120-second deadline** —
+  lightweight dialog sessions now honor `runtime.max_session_seconds` (or the
+  Copilot provider's 1800-second default), and timeout errors are classified as
+  retryable so console and web dialogs can continue without losing history.
+
 ## [0.1.23](https://github.com/microsoft/conductor/compare/v0.1.22...v0.1.23) - 2026-07-20
 
 ### Added
--- a/src/conductor/providers/copilot.py
+++ b/src/conductor/providers/copilot.py
@@ -2468,12 +2468,13 @@ class CopilotProvider(AgentProvider):
             session.on(on_event)
             await session.send(full_prompt)
 
+            dialog_timeout = self._idle_recovery_config.max_session_seconds
             try:
-                await asyncio.wait_for(done.wait(), timeout=120.0)
+                await asyncio.wait_for(done.wait(), timeout=dialog_timeout)
             except TimeoutError as exc:
                 raise ProviderError(
-                    "Dialog turn timed out after 120s",
-                    is_retryable=False,
+                    f"Dialog turn timed out after {dialog_timeout:g}s",
+                    is_retryable=True,
                 ) from exc
 
             if error_message:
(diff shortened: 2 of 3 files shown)


src/conductor/providers/copilot.py
```diff
@@ -2468,12 +2468,13 @@ def on_event(event: Any) -> None:
             session.on(on_event)
             await session.send(full_prompt)
 
+            dialog_timeout = self._idle_recovery_config.max_session_seconds
```
jrob5756, line 2471: This bumps the default dialog-turn timeout from 120s to 1800s for anyone who hasn't set `runtime.max_session_seconds`, and it's a single flat `wait_for` with none of the periodic idle-check/recovery logic the main agent loop has around this same config value. If a dialog session genuinely hangs, a console or web user could be staring at nothing for 30 minutes before seeing an error.

Would a shorter, dialog-specific default (or at least a log line once `idle_timeout_seconds` has elapsed) be worth adding here, so a stall isn't completely silent until the full deadline hits?

src/conductor/providers/copilot.py
```diff
@@ -2468,12 +2468,13 @@ def on_event(event: Any) -> None:
...
             try:
-                await asyncio.wait_for(done.wait(), timeout=120.0)
+                await asyncio.wait_for(done.wait(), timeout=dialog_timeout)
             except TimeoutError as exc:
                 raise ProviderError(
-                    "Dialog turn timed out after 120s",
-                    is_retryable=False,
+                    f"Dialog turn timed out after {dialog_timeout:g}s",
+                    is_retryable=True,
```
jrob5756, line 2477: Neither `gates/dialog.py` nor `dialog_evaluator.py` look at `.is_retryable` on this error — both just catch a bare `Exception` and move on. So this flip doesn't change any observable behavior today, and it puts this raise at odds with the main execution loop, which treats the identical `max_session_seconds` timeout as non-retryable on purpose (see the `# Don't retry — same root cause will recur` comment a bit further up in this file).

If there's a future retry path planned for dialog turns, a short comment here explaining the divergence would help the next person who reads this. Otherwise I'd lean toward keeping this `False` to match the main loop.

```suggestion
                    is_retryable=False,
```

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/microsoft/conductor/compare/v0.1.23...HEAD)
 
+### Fixed
+
+- **Copilot dialog turns no longer fail after a hidden 120-second deadline** —
+  lightweight dialog sessions now honor `runtime.max_session_seconds` (or the
+  Copilot provider's 1800-second default), and timeout errors are classified as
+  retryable so console and web dialogs can continue without losing history.
```
jrob5756, line 15: Minor: the "so console and web dialogs can continue without losing history" part already happens today regardless of `is_retryable` — both dialog call sites roll back the user turn and continue on any exception, not just retryable ones. Might be worth tightening this to just describe the timeout fix itself.

jrob5756 (commented):
Nice to see your name again 😄. I like the change. Just two minor inline notes worth a look before merging: the is_retryable flip doesn't seem to reach any code that checks it, and reusing max_session_seconds for a single flat wait removes the fast-failure behavior without putting anything back in its place. Lint, types, and the full suite all pass, and the new regression test is solid (I reverted the fix locally and confirmed it fails).

Issue #118, opened 2026-04-24
feat: allow script agents to declare output schemas for typed downstream access


PolyphonyRequiem:
### Current behavior

Script agents (\	ype: script\) cannot declare an \output:\ block. Validation fails with:

\\\
script agents cannot have 'output' schema (output is always stdout/stderr/exit_code)
\\\

### Desired behavior

Script agents should be able to declare typed output schemas, just like LLM agents. The script writes JSON to stdout; conductor should parse it according to the declared schema and make it available as \gent_name.output.field_name\ for downstream agents and routing conditions.

### Why this matters

When converting LLM agents to deterministic scripts (for routing, validation, state detection), the output contract is the same — downstream agents need typed fields. Currently the workaround is:

1. Route via exit codes (e.g., \xit_code == 10\ → planning, \xit_code == 20\ → implementation)
2. Downstream agents read the raw stdout JSON

This works but has drawbacks:
- Exit codes are opaque — no self-documenting schema
- Routing conditions can't reference output fields (\output.phase == 'done'\)
- No validation that the script actually outputs the expected JSON structure
- The exit code workaround diverges from how LLM agent routing works

### Proposed solution

Allow \	ype: script\ agents to optionally declare \output:\ with the same schema syntax as LLM agents. Conductor would:
1. Parse stdout as JSON
2. Validate it against the declared schema (if present)
3. Make fields available as \gent_name.output.field_name\
4. Allow \when:\ conditions to reference output fields

If no \output:\ is declared, current behavior (raw stdout/stderr/exit_code) is preserved.

### Workaround

Currently using exit codes for routing and raw stdout for context. Functional but less ergonomic than typed outputs.

### Environment

- Conductor v0.1.9
- Python 3.14.2
- Windows 11

PR #337, opened 2026-07-26
fix(executor): validate nested array item output schemas recursively

hertznsk:
## Summary

Continues the output-schema tightening series started in #311 (shared builders extracted into `_schema.py`) and #317 (Copilot and Hermes aligned on the shared recursive prompt-schema builder). Those PRs made the schema that providers *send to the model* fully recursive — but `validate_output` (`executor/output.py`), which checks the model's response against the declared `output:` contract, still only type-checked top-level `array` items. An `array<object>` (or deeper) output with missing or mistyped nested fields therefore passed validation silently, even though the model had received the full nested schema.

This PR closes that gap: the per-field validation logic is extracted into a recursive `_validate_field` helper that recurses into both `object.properties` (existing behavior, moved verbatim) and `array.items` (new), so nested array item schemas are enforced at every depth — in all five `validate_output` call sites at once (LLM agents, Claude, Hermes, `set` steps, `script` steps).

## Behavior change

This is a **tightening bugfix**: workflows whose agents, `set` steps, or `script` steps emit `array<object>` outputs that violate the declared nested schema will now fail with `ValidationError` instead of passing silently. The declared `output:` contract is finally enforced the same way the schema builders already advertise it to the model.

## Compatibility

- **Flat schemas and `object` nesting: byte-identical behavior and error messages.** All pre-existing validation tests pass unmodified, pinning the message shapes (`Missing required output field: ...`, `Output field '...' has wrong type: ...`, `Array item N in '...' has wrong type: ...`).
- No changes to providers, engine, CLI, or the workflow YAML schema — the fix lives entirely in `executor/output.py` and propagates through the existing call sites.
- No new depth limit in `validate_output`: schemas deeper than the builders' `max_depth=10` are still rejected at schema-build time, before validation ever runs.
- Arrays declared without `items` keep their historical passthrough.

## Explicitly out of scope

- `additionalProperties: false` enforcement (follow-up PR).
- Self-healing / validation retry in the Claude agentic loop (follow-up PR).
- New `OutputField` fields (`enum`, `pattern`, `required`, `nullable`, min/max).
- Path prefixes in error messages (`findings[2].severity`) — messages stay byte-compatible here; a separate diagnostics PR will add paths.

## Test plan

- 7 new regression tests in `TestValidateOutputArrayRecursion`: valid `array<object>` passes, missing nested field raises, wrong nested type raises, `array<array<object>>` deep error raises, no-`items` passthrough preserved, builder-boundary depth-10 schema (5 array/object pairs, verified acceptable to `build_json_schema_field`) rejects invalid content without `RecursionError`, and a source-inspection guard that `validate_output` never calls `_check_type` directly (all value checks flow through `_validate_field`).
- Mutation check: commenting out the recursive array-item call makes exactly the 4 nested-validation tests fail; restoring it returns the suite to green.
- One integration-test mock (`test_workflows.py`) that violated its own declared `array<object>` schema (missing required `action` inside a `plan` item) is brought into conformance — a one-line mock fix, no test-semantics change. This was the only in-repo consumer relying on the previously silent acceptance.
- `make test`: 4338 passed. `make check` (ruff + format + ty): clean. `make validate-examples`: green.

### DIFF
--- a/src/conductor/executor/output.py
+++ b/src/conductor/executor/output.py
@@ -19,6 +19,7 @@ def validate_output(
     """Validate agent output against declared schema.
 
     Checks that all required fields are present and have the correct types.
+    Nested object properties and array items are validated recursively.
 
     Args:
         content: Agent's output content as a dictionary.
@@ -40,29 +41,44 @@ def validate_output(
                 suggestion=f"Ensure agent returns '{field_name}' in output",
             )
 
-        value = content[field_name]
-        expected_type = field_def.type
+        _validate_field(field_name, content[field_name], field_def)
 
-        # Type checking
-        if not _check_type(value, expected_type):
-            raise ValidationError(
-                f"Output field '{field_name}' has wrong type: "
-                f"expected {expected_type}, got {type(value).__name__}",
-                suggestion=f"Ensure agent returns correct type for '{field_name}'",
-            )
 
-        # Recursively validate nested structures
-        if expected_type == "object" and field_def.properties and isinstance(value, dict):
-            validate_output(value, field_def.properties)
+def _validate_field(field_name: str, value: Any, field_def: OutputField) -> None:
+    """Validate a single value against its output field definition.
 
-        if expected_type == "array" and field_def.items and isinstance(value, list):
-            for i, item in enumerate(value):
-                if not _check_type(item, field_def.items.type):
-                    raise ValidationError(
-                        f"Array item {i} in '{field_name}' has wrong type: "
-                        f"expected {field_def.items.type}, got {type(item).__name__}",
-                        suggestion=f"Ensure all items in '{field_name}' have correct type",
-                    )
+    Recursively validates nested object properties and array items so
+    ``array<object>`` and deeper combinations are checked at every depth,
+    matching the recursion the object branch has always had.
+
+    Args:
+        field_name: Field name used in error messages (array items keep the
+            parent array's name, matching the existing message style).
+        value: The value to validate.
+        field_def: The field definition to validate against.
+
+    Raises:
+        ValidationError: If the value or any nested value doesn't match.
+    """
+    if not _check_type(value, field_def.type):
+        raise ValidationError(
+            f"Output field '{field_name}' has wrong type: "
+            f"expected {field_def.type}, got {type(value).__name__}",
+            suggestion=f"Ensure agent returns correct type for '{field_name}'",
+        )
+
+    if field_def.type == "object" and field_def.properties and isinstance(value, dict):
+        validate_output(value, field_def.properties)
+
+    if field_def.type == "array" and field_def.items and isinstance(value, list):
+        for i, item in enumerate(value):
+            if not _check_type(item, field_def.items.type):
+                raise ValidationError(
+                    f"Array item {i} in '{field_name}' has wrong type: "
+                    f"expected {field_def.items.type}, got {type(item).__name__}",
+                    suggestion=f"Ensure all items in '{field_name}' have correct type",
+                )
+            _validate_field(field_name, item, field_def.items)
 
 
 def _check_type(value: Any, expected: str) -> bool:
--- a/tests/test_executor/test_output.py
+++ b/tests/test_executor/test_output.py
@@ -8,6 +8,8 @@ Tests cover:
 - JSON parsing from raw responses
 """
 
+from typing import Any
+
 import pytest
 
 from conductor.config.schema import OutputField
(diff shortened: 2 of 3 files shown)


src/conductor/executor/output.py
```diff
@@ -40,29 +41,44 @@ def validate_output(
...
+
+    if field_def.type == "array" and field_def.items and isinstance(value, list):
+        for i, item in enumerate(value):
+            if not _check_type(item, field_def.items.type):
+                raise ValidationError(
+                    f"Array item {i} in '{field_name}' has wrong type: "
+                    f"expected {field_def.items.type}, got {type(item).__name__}",
+                    suggestion=f"Ensure all items in '{field_name}' have correct type",
+                )
+            _validate_field(field_name, item, field_def.items)
```
jrob5756, line 81: Passing `field_name` down unchanged means the index and the name end up describing different levels. For `array<array<object>>` with the bad element at `matrix[1][0]`, you get `Array item 0 in 'matrix'`, and `matrix[0]` is perfectly valid. This path was unreachable before the PR, so there is no message to preserve here.

The one-line change below addresses the array case. The object case still reports a bare `Missing required output field: score` with no array or index, because `validate_output` re-derives names from the nested schema keys. Threading an optional `_path` prefix through `validate_output` fixes both and lets the double type-check collapse into one call.

```suggestion
    if field_def.type == "array" and field_def.items and isinstance(value, list):
        for i, item in enumerate(value):
            # Checked here as well as in the recursive call below, so an element type
            # mismatch keeps the indexed "Array item {i}" wording instead of falling
            # through to the generic "Output field" one.
            if not _check_type(item, field_def.items.type):
                raise ValidationError(
                    f"Array item {i} in '{field_name}' has wrong type: "
                    f"expected {field_def.items.type}, got {type(item).__name__}",
                    suggestion=f"Ensure all items in '{field_name}' have correct type",
                )
            _validate_field(f"{field_name}[{i}]", item, field_def.items)
```

src/conductor/executor/output.py
```diff
@@ -40,29 +41,44 @@ def validate_output(
...
+def _validate_field(field_name: str, value: Any, field_def: OutputField) -> None:
+    """Validate a single value against its output field definition.
+
+    Recursively validates nested object properties and array items so
+    ``array<object>`` and deeper combinations are checked at every depth,
+    matching the recursion the object branch has always had.
+
+    Args:
+        field_name: Field name used in error messages (array items keep the
+            parent array's name, matching the existing message style).
```
jrob5756, line 56: "matching the recursion the object branch has always had" describes the change rather than the contract, and it will read as stale once nobody remembers what the object branch used to do. The two branches also do not match: one recurses through `validate_output`, the other through `_validate_field`.

The `field_name` note is also not accurate for the paths this PR adds. Nested object properties are renamed from the nested schema keys, so the parent name is dropped entirely.

```suggestion
    Checks ``value`` against ``field_def.type``, then recurses: object values are
    validated against ``field_def.properties`` and array elements against
    ``field_def.items``. Fields declared ``array`` without ``items`` or ``object``
    without ``properties`` are checked only at their own level; their contents are
    accepted unexamined.

    Args:
        field_name: Name used in error messages. Array elements are addressed as
            ``name[i]``; nested object properties are renamed by ``validate_output``
            from the nested schema keys.
```

src/conductor/executor/output.py
```diff
@@ -19,6 +19,7 @@ def validate_output(
     """Validate agent output against declared schema.
 
     Checks that all required fields are present and have the correct types.
+    Nested object properties and array items are validated recursively.
```
jrob5756, line 22: Worth spelling out the two things a caller cannot infer here, since both decide whether an existing workflow starts failing. `OutputField` has no optional or nullable flag, so declaring `items.properties` makes every one of those keys mandatory on every element. And an array declared without `items` still accepts anything, including nulls and mixed types.

```suggestion
    Nested object properties and array items are validated recursively, at every
    depth. Every field a schema declares is required (``OutputField`` has no optional
    flag), so declaring ``items.properties`` requires those keys on every element.
    Arrays without ``items`` and objects without ``properties`` pass their contents
    through unchecked.
```

tests/test_executor/test_output.py
```diff
@@ -202,6 +204,148 @@ def test_array_items_wrong_type(self) -> None:
...
+        """After refactor all value checks must flow through _validate_field."""
+        import inspect
+
+        from conductor.executor.output import _validate_field, validate_output
+
+        validate_source = inspect.getsource(validate_output)
+        validate_field_source = inspect.getsource(_validate_field)
+
+        assert "_check_type(" not in validate_source
+        assert "_check_type(" in validate_field_source
```
jrob5756, line 346: I ran the mutation check: comment out the recursive `_validate_field` call in `output.py` and four of the seven new tests fail, but this one still passes. It greps source text, so it stays green against a validator with the fix removed, and it goes red on a rename or on the path refactor suggested above. The behavioural tests above it already cover the invariant.

```suggestion

```

tests/test_executor/test_output.py
```diff
@@ -202,6 +204,148 @@ def test_array_items_wrong_type(self) -> None:
...
+                ),
+            )
+        }
+        content = {
+            "matrix": [
+                [{"v": 1.0}, {"v": "x"}],
+            ]
+        }
+
+        with pytest.raises(ValidationError, match="wrong type"):
```
jrob5756, line 300: `match="wrong type"` matches both message forms at every depth, so this test cannot tell you which message fired or at what level. That matters for a PR whose stated invariant is byte-for-byte message preservation. Tightening it also turns this test into a real detector for the index/name mismatch flagged in `output.py`.

```suggestion
        with pytest.raises(
            ValidationError, match=r"Output field 'v' has wrong type: expected number, got str"
        ):
```

tests/test_executor/test_output.py
```diff
@@ -202,6 +204,148 @@ def test_array_items_wrong_type(self) -> None:
...
+
+    def test_array_without_items_unchanged(self) -> None:
+        """Arrays declared without items keep historical passthrough behavior."""
+        schema = {"tags": OutputField(type="array")}
+        content = {"tags": [1, "two", {"three": 3}]}
+
+        validate_output(content, schema)
+
+    def test_builder_shaped_deep_schema_invalid_content_raises(self) -> None:
+        """A schema at the builder boundary (max_depth=10) must be
```
jrob5756, line 311: Depth 10 against CPython's recursion limit of 1000 is not really a recursion test. A `RecursionError` would fail the test (it is not a `ValidationError`), so the claim is not false, but a reader will believe depth safety is covered here when it is not. The genuinely useful fact is the boundary itself: 5 array/object pairs build, 6 raise `SchemaDepthError`.

```suggestion
    def test_builder_shaped_deep_schema_invalid_content_raises(self) -> None:
        """Deepest provider-buildable schema (5 array/object pairs) still validates."""
```

tests/test_executor/test_output.py
```diff
@@ -202,6 +204,148 @@ def test_array_items_wrong_type(self) -> None:
...
+                type="array",
+                items=OutputField(
+                    type="object",
+                    properties={"x": inner},
+                ),
+            )
+        schema_root = {"data": inner}
+
+        built = build_json_schema_field(inner)
+        assert isinstance(built, dict)
```
jrob5756, line 328: `isinstance(built, dict)` is trivially true against the `-> dict[str, Any]` annotation. The build call earns its keep, the assertion does not, and the reason for building at all is invisible.

```suggestion
        # Building first asserts this depth is legal for a provider schema: 5
        # array/object pairs sit exactly on _schema.py's max_depth=10 limit.
        build_json_schema_field(inner)
```

tests/test_executor/test_output.py
```diff
@@ -202,6 +204,148 @@ def test_array_items_wrong_type(self) -> None:
...
+
+        with pytest.raises(ValidationError, match="wrong type"):
+            validate_output(content, schema)
+
+    def test_array_without_items_unchanged(self) -> None:
+        """Arrays declared without items keep historical passthrough behavior."""
+        schema = {"tags": OutputField(type="array")}
+        content = {"tags": [1, "two", {"three": 3}]}
+
+        validate_output(content, schema)
```
jrob5756, line 308: Good test, and the right guardrail for a tightening change. Two siblings are worth adding next to it. A JSON `null` inside an array item is the likeliest real-world break from this PR, since models emit null for unknown fields constantly, and it currently produces a type error rather than a missing-field error:

```python
def test_null_inside_array_item_field_raises(self) -> None:
    """JSON null in an array item field is a type error, not an absent field."""
    schema = {
        "findings": OutputField(
            type="array",
            items=OutputField(
                type="object", properties={"title": OutputField(type="string")}
            ),
        )
    }
    with pytest.raises(ValidationError, match=r"expected string, got NoneType"):
        validate_output({"findings": [{"title": None}]}, schema)
```

An empty list and `items: {type: object}` with no `properties` both pass today and are also unpinned.

tests/test_executor/test_output.py
```diff
@@ -202,6 +204,148 @@ def test_array_items_wrong_type(self) -> None:
             validate_output(content, schema)
 
 
+class TestValidateOutputArrayRecursion:
```
jrob5756, line 207: "(issue regression)" names no issue. The convention elsewhere in the repo is to cite the number, for example `engine/workflow.py:1111` ("issue #118"), which is actually followable.

```suggestion
    """Tests for recursive validation of array item schemas (issue #NNN)."""
```

tests/test_executor/test_output.py
```diff
@@ -202,6 +204,148 @@ def test_array_items_wrong_type(self) -> None:
...
+            "findings": [
+                {"title": "first", "score": 1.0},
+                {"title": "second", "score": 2.0},
+            ]
+        }
+
+        validate_output(content, schema)
+
+    def test_array_of_objects_missing_nested_field_raises(self) -> None:
+        """Missing required field inside an array item must raise (previously silently accepted)."""
```
jrob5756, line 234: "previously silently accepted" is true relative to one commit and will not be in a year. Same pattern on line 281 ("pinning general recursion").

```suggestion
        """Missing required field inside an array item must raise."""
```

tests/test_integration/test_workflows.py
```diff
@@ -744,7 +744,7 @@ def mock_handler(agent, prompt, context):
             agent_calls.append(agent.name)
 
             if agent.name == "planner":
-                return {"plan": [{"step": "1"}], "confidence": 0.9}
+                return {"plan": [{"step": "1", "action": "Research"}], "confidence": 0.9}
```
jrob5756, line 747: I reverted this and watched `test_existing_full_workflow_unchanged` fail, so the fix is real. But it reads as arbitrary test data, and the test name points away from the cause. One line of context saves the next person trimming this fixture.

```suggestion
            if agent.name == "planner":
                # valid_full.yaml declares plan.items.properties = {step, action}, and
                # every declared property is required, so both keys must be present.
                return {"plan": [{"step": "1", "action": "Research"}], "confidence": 0.9}
```

jrob5756 (approved):
The recursive validation fix is correct and closes a real contract gap. The remaining comments are non-blocking suggestions around diagnostics, documentation, and broader call-site coverage. Approved.

Issue #335, opened 2026-07-22
claude-agent-sdk provider: implement MCP server translation (mcp_tools=False is a gap, not an SDK limit)


feltroidprime:
## Summary

The `claude-agent-sdk` provider (`src/conductor/providers/claude_agent_sdk.py`) declares `mcp_tools=False` in its `CAPABILITIES` and rejects workflow-level `runtime.mcp_servers` at the factory. The comment at `claude_agent_sdk.py:115-116` says why:

> MCP servers are rejected at the factory (no translation from Conductor's MCP config to the SDK's MCP dict is implemented).

This is an implementation gap, not an upstream SDK limitation. `claude-agent-sdk` (locked at `0.2.87` in `uv.lock`) genuinely supports MCP servers via `ClaudeAgentOptions.mcp_servers: dict[str, McpServerConfig]`, where `McpServerConfig` is one of `McpStdioServerConfig` (`command`/`args`/`env`), `McpSSEServerConfig` (`url`/`headers`), or `McpHttpServerConfig` (`url`/`headers`).

Conductor's own `MCPServerDef` (`src/conductor/config/schema.py:1605`) already has an almost 1:1 shape: `type: "stdio"|"http"|"sse"`, `command`, `args`, `env`, `url`, `headers`, `timeout`, `tools` (allowlist, `["*"]` = all).

## Proposal

- Translate `MCPServerDef` → the SDK's `McpStdioServerConfig` / `McpSSEServerConfig` / `McpHttpServerConfig` shapes and pass them via `ClaudeAgentOptions.mcp_servers`.
- Use the SDK's `allowed_tools` (or equivalent scoping) to honor each server's per-tool `tools:` allowlist and the per-agent resolved `tools:` list from `resolve_agent_tools()`, mirroring how `claude.py` filters MCP tools today (`_convert_mcp_tools_to_claude`'s `tool_filter`).
- Flip `CAPABILITIES.mcp_tools` to `True` once wired, and update the `_resolve_tool_config` docstring/behavior in `claude_agent_sdk.py` (it currently refuses any non-empty per-agent `tools:` allowlist because there's no MCP-name → CLI-tool-ID mapping — that refusal reasoning goes away once MCP servers route through the SDK's own MCP support instead of native CLI tool IDs).
- Update `docs/providers/experimental.md` and the provider-parity notes in `AGENTS.md` accordingly.

## Why it matters

[...]

Issue #348, opened 2026-07-30
claude-agent-sdk: forward the resolved working_dir to ClaudeAgentOptions.cwd


jrob5756:
## Summary

`ClaudeAgentSdkProvider.CAPABILITIES` declares `working_dir=False`, so `conductor validate` rejects any workflow setting `agent.working_dir` or `runtime.working_dir` against this provider.

Until #335 the stated reason was that MCP servers were rejected at the factory, so there was "nothing to scope by directory". That premise is gone — the provider now spawns MCP servers — and the comment was rewritten in #346 to give the real reason: **the provider never forwards the engine-resolved working directory to `ClaudeAgentOptions.cwd`.** The capability is still accurate, but it is now a plain gap rather than a consequence of another one.

## Why it's worth closing

`ClaudeAgentOptions` has a `cwd` field (`types.py:1699`), so the plumbing is straightforward. Two things depend on it:

- The `claude` CLI itself runs in the resolved directory, matching what `copilot` and `claude` already do (#297).
- Any stdio MCP server the CLI spawns inherits that cwd, so a server declared with a relative path behaves consistently across providers.

Today both silently use whatever directory the Conductor process happens to be in.

## Scope

- Forward the resolved `working_dir` to `ClaudeAgentOptions.cwd` in `execute()`.
- Flip `CAPABILITIES.working_dir` to `True` and update the comment at `providers/claude_agent_sdk.py`.
- Update the carve-out row in `docs/providers/experimental.md` and the parity notes in `AGENTS.md`.
- Mirror the existing working-dir tests for this provider (see #297 for the pattern).

Worth confirming how `cwd` interacts with `add_dirs` and whether a non-existent directory should fail before the SDK call, matching the engine's existing behavior of raising an execution error for a missing target directory (`docs/mcp-tools.md:259`).

## Related

- #297 — `working_dir` for LLM agents and their MCP servers
- #335 / #346 — MCP server support, which invalidated the old justification
- #304 / #305 — documenting the `working_dir` carve-out

jrob5756:
Dug into `claude-agent-sdk` 0.2.82 to answer the two open questions at the bottom.

[...]

PR #354, opened 2026-07-31
fix(claude-agent-sdk): load skills natively and stop inheriting ambient ones

### LINKED ISSUE #352: claude-agent-sdk: use native skills option, and stop loading ambient skills implicitly
`claude_agent_sdk.py` takes the eager preamble-injection path for skills, on the stated grounds that the upstream SDK has no native surface. That is out of date — the native options exist and are older than our pin. Separately, because the provider never constrains `setting_sources`, the underlying `claude` CLI is already loading skills from the user's machine that the workflow never declared.

Both verified against the real installed package (`claude-agent-sdk` 0.2.128; our pin is `>=0.2.82`).

## 1. The native surface exists

`claude_agent_sdk.py` currently leaves `supports_native_skills` at the default `False`, with this comment on `CAPABILITIES`:

> Skill content is eagerly injected into the rendered prompt by AgentExecutor (the claude-agent-sdk surfaces no `skill_directories` kwarg today; once it does we can flip to native via `supports_native_skills`).

There is indeed no kwarg named `skill_directories` — that is a Copilot name. The equivalent is called `skills`:

```python
>>> dataclasses.fields(ClaudeAgentOptions)   # claude-agent-sdk 0.2.128
skills:          list[str] | Literal['all'] | None      = None
setting_sources: list[Literal['user','project','local']] | None = None
plugins:         list[SdkPluginConfig]                  = <factory>
```

`skills` landed in v0.1.62 and `plugins` in v0.1.5, both well below our `>=0.2.82` pin. From the SDK's own docstring:

> **`skills`** — Skills to enable for the main session. This is the single place to turn skills on; you do not need to add `"Skill"` to `allowed_tools` or set `setting_sources` yourself — the SDK does both when this is set.
>
> - `"all"`: enable every discovered skill.
> - `list[str]`: enable only the listed skills. Names match the SKILL.md `name` / directory name, or `plugin:skill` for plugin-qualified skills.

Flipping to native would also make this provider **progressive** rather than eager: Claude Code loads only `name` + `description` (~100 tokens/skill) up front and reads the body on demand. Today we prepend the full `SKILL.md` plus the entire `references/` tree on every call — for the bundled `conductor` skill that is 109KB (~27K tokens) per agent invocation, per retry, per validator call.

## 2. `skills: []` does not actually disable skills here

This is the correctness problem. The provider never sets `setting_sources`, and the SDK docstring is explicit about what the default means:

> **`setting_sources`** — Control which filesystem settings to load. When `None`, **all sources are loaded** (matches CLI defaults). Pass `[]` to disable filesystem settings (SDK isolation mode).

and for `skills`:

> `None` (default): no SDK auto-configuration. The CLI's own defaults still apply, so this is **not** "skills off" — to suppress every skill from the listing, use `[]`.

So with both left at `None`, the `claude` CLI discovers and enables skills from `~/.claude/skills/`, `.claude/skills/` (cwd and every parent up to the repo root), and enabled plugins. Consequences:

- An agent receives skills the workflow never declared, varying by developer machine and by which directory the run happens to start in.
- Conductor documents `skills: []` as an **explicit opt-out**. On this provider it is not — `AgentExecutor` injects nothing, but the CLI still loads the ambient set. The opt-out silently fails to opt out.
- `CAPABILITIES.skills=True` is therefore describing only half of what the provider actually does.

This is the same class of issue that `strict_mcp_config=True` was introduced to prevent, a few lines away in the same file:

> `strict_mcp_config=True` is set **unconditionally**, including when the workflow declares no servers: otherwise the CLI loads project `.mcp.json`, user-global, and plugin-provided servers, and `permission_mode` bypasses approval for whatever they expose.

We close that door for MCP servers and leave it open for skills. Note the default `setting_sources` also pulls in `CLAUDE.md`, `.claude/rules/*.md`, project `settings.json`, and hooks — worth deciding deliberately rather than by omission.

## Proposed change

1. Set `supports_native_skills = True` on the provider and forward resolved skills via `ClaudeAgentOptions.skills` instead of eager injection. Names are `SKILL.md` `name` / directory name, or `plugin:skill`.
2. Set `setting_sources` explicitly rather than inheriting the CLI default, so ambient discovery is a deliberate choice. Pair with `plugins=` for any skill directory Conductor resolves itself.
3. Pass `skills=[]` when the workflow opts out, so `skills: []` means what the docs claim on every provider.
4. Update the `CAPABILITIES` comment and the `claude_agent_sdk.py` parity notes in `AGENTS.md`.

Worth confirming with a live run against the `claude` CLI before implementing — the reasoning above is from the SDK source and docs, not an executed end-to-end test (`claude-agent-sdk` is an optional extra and is not installed in the default dev venv).

## Related

- #350 — skill discovery across providers; contains the full provider matrix and the token-cost measurements
- #215 — introduced the skills mechanism and this provider's eager-injection path
- #351 — added the `capabilities.skills` validator cross-check

jrob5756:
Closes #352

## What

`claude_agent_sdk.py` declared `supports_native_skills=False` on the grounds that the upstream SDK had no skill surface. That is out of date, and expensive: the full `SKILL.md` plus the entire `references/` tree was injected into every rendered prompt, on every call, every retry, and every validator pass (~27K tokens for the bundled `conductor` skill).

It also never set `setting_sources`, so the `claude` CLI discovered and enabled skills from `~/.claude/skills/`, every `.claude/skills/` up the directory tree, and enabled plugins — none of which the workflow declared, all of which varied by developer machine and launch directory. Conductor documents `skills: []` as an explicit opt-out; on this provider it opted out of nothing.

Same class of hole `strict_mcp_config=True` already closes for MCP servers, a few lines away in the same file.

## How

- **Native loading** — the Claude Code plugin owning each skill is registered via `ClaudeAgentOptions.plugins` and the skill enabled by its `<plugin>:<skill>` name, so the CLI reads only the `SKILL.md` frontmatter up front and loads the body on demand. `_resolve_skill_plugins` maps each resolved skill directory back to its plugin (`skills/registry.py::resolve_skill_plugin` walks up for `.claude-plugin/plugin.json`); a directory with no manifest raises `ProviderError` rather than being silently dropped.
- **`setting_sources=[]` unconditionally** — the skills counterpart to the unconditional `strict_mcp_config`. Note `skills=[]` and `skills=None` are *not* interchangeable upstream: `None` means "CLI defaults apply", so only the explicit empty list makes the opt-out real.
- **`tools: []` + skills** now grants back the single `Skill` tool. An explicit `tools: []` sends `--tools ""` (empty base tool set), which would otherwise leave the declared skill unreachable. No permission bypass needed — the SDK auto-allows it via `Skill(<name>)` in `allowed_tools`.
- **Packaging fix (not in the issue)** — the wheel `force-include` shipped only the skill body, not `plugins/conductor/.claude-plugin/`. Without the manifest, a non-editable install resolved a plugin root the CLI cannot load. Verified by building the wheel and resolving the skill from an actual wheel install.

## ⚠️ Behaviour change

Agents on this provider no longer inherit ambient `CLAUDE.md`, `.claude/rules/*.md`, project `settings.json`, or hooks. Use `--workspace-instructions` / `--instructions` to supply that content explicitly. Called out in the CHANGELOG.

## Verification

The SDK surface was checked against the **real installed package at the pin floor `0.2.82`** as well as current `0.2.128` — `skills`, `setting_sources`, and `plugins` all exist at both, so **no pin bump is needed**.

Tests assert the actual argv the SDK builds rather than stopping at the options object, across all four skills x tools combinations:

| Case | argv |
|---|---|
| skills, `tools: []` | `--tools Skill --allowedTools Skill(conductor:conductor) --setting-sources= --plugin-dir <root>` |
| skills, tools omitted | `--tools default --allowedTools Skill(conductor:conductor) --setting-sources= --plugin-dir <root>` |
| no skills, `tools: []` | `--tools "" --setting-sources=` (no `--plugin-dir`) |
| no skills, tools omitted | `--tools default --setting-sources=` (no `--plugin-dir`) |

| Check | Result |
|---|---|
| Full suite **with** `claude-agent-sdk` extra (ci.yml:109) | 4731 passed |
| Full suite **without** the extra (default CI job) | 4580 passed |
| `make check` (ruff + ty, CI config) | clean |
| `make validate-examples` | exit 0 |
| Wheel build + install | manifest packaged, `conductor:conductor` resolves |

## Review notes

- The one thing not verifiable offline is whether the CLI wants the qualified `conductor:conductor` or the bare `conductor`. We send the documented plugin-qualified form (and the repo's own local-dev flow is `claude --plugin-dir plugins/conductor`). It is a one-line change in `_resolve_skill_plugins` if a live run says otherwise.
- Draft: worth a live smoke test against a real `claude` CLI before merging.
- Unrelated stale doc spotted and left alone: `references/yaml-schema.md` still says `mcp_servers` is "ignored by claude-agent-sdk", which #335 made false. Worth a follow-up.


jrob5756:
Addressed review feedback in d6a70ed. Seven review passes; no blocking bugs, but several findings converged on one theme: the first pass was loud about the single case where it couldn't build a skill name, and silent about every case where the name it built was **wrong** — the same failure mode #352 exists to fix.

**Resolution now refuses, with the real reason attached**
- A manifest found above a skill no longer makes that plugin its owner — the skill must live under the candidate's `skills/`, or the walk continues.
- `SKILL.md` must exist and its frontmatter `name` must equal the directory name. The CLI resolves by frontmatter name while we send the directory name, so drift *hides* the skill rather than failing.
- Names restricted to `[A-Za-z0-9_.-]+` — they're joined into a comma-delimited `--allowedTools` value, where `,` or `:` splits into extra permission rules.
- Two plugins claiming one qualified name are refused, not deduped (deduping drops a declared skill).
- `resolve_skill_plugin` returns `None` only for "no owning plugin root" and raises `SkillPluginError` when a plugin is present but unusable, so a broken manifest stops being reported as a missing one.
- All resulting `ProviderError`s set `is_retryable=False`, matching sibling config errors — the message interpolates a filesystem path, and the default heuristic sniffs for "connection"/"timeout".

**Test gap that mattered most**

The executor→provider seam had *zero* coverage. Four mutations there — including `skill_directories=None` — suppressed every skill with the entire suite green. Eager injection used to backstop that seam; this PR removed the backstop. Now covered, and I verified all three previously-silent regressions fail under mutation:

| Mutation | Before | After |
|---|---|---|
| `skill_directories=None` in executor | green | **fails** |
| `supports_native_skills` check removed | green | **fails** |
| `SKILL.md` frontmatter `name` renamed | green | **fails** |

**Docs corrections** — several claims were wrong, including mine:
- A missing wheel manifest fails loudly; it does not "silently resolve to nothing".
- The `skills: []` opt-out needs `setting_sources=[]` **and** an explicit `skills` list — the SDK re-defaults `setting_sources` to `["user","project"]` when `skills` is set without it. The changelog credited only one, the source comments only the other; either could have been "simplified" away, re-opening the issue with tests green.
- The skill list is a context filter, not a sandbox — undeclared skills are hidden from the listing, but their files stay readable.
- `docs/providers/comparison.md` had no mention of skills or the settings isolation at all; a user's only notice that `CLAUDE.md` stopped loading was the changelog.

A confirmation pass then caught one bug introduced by the fixes: `SkillPlugin.__post_init__` raised a bare `ValueError` that slipped past the provider's `except SkillPluginError`. Fixed and regression-tested.

**Verification:** 4760 passed with the `claude-agent-sdk` extra / 4606 without (+29 and +26), `make check` clean in CI's config, `make validate-examples` exit 0.

Still draft — the qualified `conductor:conductor` vs bare `conductor` name remains the one thing not verifiable without a live `claude` CLI.

jrob5756:
Rebased onto `main` to pick up #349 (`working_dir` via `ClaudeAgentOptions.cwd`), which had landed in the meantime and touched the same four files.

The two changes are independent — `cwd` versus `skills`/`setting_sources`/`plugins` — so the conflicts were mechanical: both sides of the `CAPABILITIES` block, the options construction, the AGENTS.md parity notes, and two test classes that git had interleaved because both were appended at the end of the file. Both `TestWorkingDirectory` and `TestSkillsWiring` are intact.

Worth flagging one thing the rebase surfaced that a clean auto-merge would have hidden: #348 documented, correctly at the time, that cwd drags the CLI's ambient `CLAUDE.md`, settings, and hooks along with it — including the line *"Conductor never sets `setting_sources`, so the CLI's load-everything default applies."* This PR makes that false. After the rebase, AGENTS.md asserted it three lines above the bullet explaining why `setting_sources` is now always `[]`. Corrected in AGENTS.md, CHANGELOG.md, and `docs/providers/experimental.md`.

The two features compose rather than collide: `cwd` still selects the transcript directory and where a project `.mcp.json`/`.claude/` tree *would* be looked for, while `strict_mcp_config` and `setting_sources=[]` stop anything there being loaded. "Point `working_dir` only at trees you trust" is no longer the caveat it was.

Verification on the rebased tip: 4773 passed with the `claude-agent-sdk` extra / 4608 without, `make check` clean, `make validate-examples` exit 0.

Marking ready for review.

Issue #353, opened 2026-07-30
MCP cleanup exits AnyIO cancel scopes from a different asyncio task


hertznsk:
## Summary

`MCPManager` keeps MCP stdio and session contexts in one shared `AsyncExitStack`. A connection may be opened from a parallel-agent or for-each worker task, while provider shutdown later closes that stack from the workflow/root task.

AnyIO cancel scopes are task-affine, so cleanup can fail with:

```text
RuntimeError: Attempted to exit cancel scope in a different task than it was entered in
```

The exception is currently caught and logged by `MCPManager.close()`, which means shutdown continues but MCP sessions/processes may not be cleaned up correctly.

## Public `main` references

Verified against `origin/main` at `c8fad5c95fdae0e748455041ed1eca706329989d`:

- The manager owns one shared `AsyncExitStack`: https://github.com/microsoft/conductor/blob/c8fad5c95fdae0e748455041ed1eca706329989d/src/conductor/mcp/manager.py#L122-L127
- `connect_server()` enters both AnyIO-backed contexts in the calling task: https://github.com/microsoft/conductor/blob/c8fad5c95fdae0e748455041ed1eca706329989d/src/conductor/mcp/manager.py#L177-L186
- `close()` exits the stack in whichever task performs provider cleanup: https://github.com/microsoft/conductor/blob/c8fad5c95fdae0e748455041ed1eca706329989d/src/conductor/mcp/manager.py#L494-L508
- Claude pools managers used by concurrent agents: https://github.com/microsoft/conductor/blob/c8fad5c95fdae0e748455041ed1eca706329989d/src/conductor/providers/claude.py#L627-L634
- Provider shutdown closes those managers later: https://github.com/microsoft/conductor/blob/c8fad5c95fdae0e748455041ed1eca706329989d/src/conductor/providers/claude.py#L755-L770

## Minimal reproduction of the lifecycle pattern

This does not require an API call or Pydantic AI; it isolates the same task-ownership mismatch:

```python
import asyncio
from contextlib import AsyncExitStack, asynccontextmanager

import anyio


@asynccontextmanager
async def task_bound_context():
    with anyio.CancelScope():
        yield


async def main():
    stack = AsyncExitStack()

    async def connect():
        await stack.enter_async_context(task_bound_context())

    await asyncio.create_task(connect())
    await stack.aclose()


asyncio.run(main())
```

Observed with Python 3.12.13 and AnyIO 4.12.1:

```text
[...]

Issue #315, opened 2026-07-17
Idea: evaluate Pydantic AI as a shared inner agent runtime for direct OpenAI and Anthropic providers


hertznsk:
## Summary

Would it be worth evaluating [Pydantic AI](https://ai.pydantic.dev/) as a shared inner agent runtime for direct OpenAI and Anthropic providers?

This is an exploratory idea rather than a migration proposal. I do not know whether it aligns with the maintainers' provider strategy, and I am not suggesting that the existing providers should be replaced without a focused technical evaluation.

> Status: `idea` — no architectural decision or implementation commitment is proposed by this issue.

## Motivation

Conductor currently has several different provider integration models:

- `CopilotProvider` delegates most of the agentic loop to the GitHub Copilot SDK.
- `ClaudeProvider` implements its own Anthropic tool/MCP loop, including retries, output recovery, usage accounting, interrupts, events, and partial output.
- `ClaudeAgentSdkProvider` delegates execution to the Claude Agent SDK and is classified as experimental due to its capability carve-outs.
- `HermesProvider` provides another model/runtime integration with its own constraints.
- `openai-agents` is present in the provider type vocabulary, but is not currently implemented.

The direct Claude implementation provides a high degree of control, but adding a direct OpenAI provider with equivalent behavior would likely require either:

1. another provider-specific agent loop;
2. a new Conductor-owned shared loop and normalized model protocol; or
3. adoption of a provider-neutral inner agent runtime.

Options 1 and 2 may duplicate functionality already maintained by an upstream agent framework.

Pydantic AI may be worth evaluating because Conductor already uses Pydantic v2, and Pydantic AI provides:

- OpenAI and Anthropic model integrations;
- a shared tool-calling loop;
- MCP toolsets and tool-call interception;
- typed and dynamically generated structured outputs;
- output validators with model retry;
- streaming text, thinking, tool-call, and tool-result events;
- usage limits and message history;
- a relatively narrow agent-runtime abstraction that may fit underneath Conductor's existing workflow engine.

## Proposed architectural boundary

The possible boundary would be:

```text
Conductor WorkflowEngine
  - YAML workflow graph
  - deterministic routing
  - parallel / for_each execution
  - workflow context
  - checkpoints and resume
  - events and dashboard
  - provider capability validation
             |
             v
Conductor AgentProvider adapter
             |
             v
Pydantic AI inner agent runtime
  - OpenAI / Anthropic model protocol
  - tool loop
  - MCP invocation
[...]

PR #215, opened 2026-05-20
feat: Conductor Expert — opt-in knowledge base for Conductor-aware agents (#180)

brrusino:
## Summary

Implements Phase 1 of issue #180: bundles the existing plugin reference docs as package data and wires them into the instructions pipeline with an opt-in flag. This gives agents deep understanding of Conductor's YAML schema, execution model, authoring patterns, and CLI commands.

### New: `src/conductor/expert/` package
- **`loader.py`** — loads and caches (~70KB) bundled reference docs via `importlib.resources` + `lru_cache`, wraps in `<conductor_knowledge>` tags
- **`knowledge/`** — bundled markdown reference docs (yaml-schema.md, authoring.md, execution.md)

### Schema (`config/schema.py`)
- **`AgentDef.conductor_expert: bool | None`** — tri-state: None = inherit workflow default, True = force enable, False = force disable. Forbidden on script/workflow/human_gate agents.
- **`RuntimeConfig.conductor_expert: bool`** — workflow-wide default (False)

### Executor (`executor/agent.py`)
- New `_build_prompt_prefix()` helper shared by `execute()` and `render_prompt()`
- New `_should_inject_expert()` — resolves tri-state agent flag vs workflow default

### Engine (`engine/workflow.py`)
- Passes `conductor_expert_default` to `AgentExecutor` in both single-provider and multi-provider paths

### Tests & Docs
- 25 new tests in `tests/test_expert/` covering loader, schema validation, and executor integration
- Updated `AGENTS.md` with expert package documentation
- Added `examples/conductor-expert.yaml` example workflow

Closes #180 (Phase 1)

Issue #343, opened 2026-07-28
copilot provider: schema-shape failures bypass the parse-recovery loop and kill the workflow (hermes recovers from the identical response)


franklixuefei:
## Summary

When an agent returns **syntactically valid JSON** whose fields have the **wrong shape**, the `copilot` provider fails the entire workflow immediately with zero recovery attempts — even though the provider has a fully-built parse-recovery mechanism (`max_parse_recovery_attempts`, default **5**) that is designed for exactly this class of "the model didn't follow the output contract" failure.

The cause is that the `copilot` provider **never calls `validate_output()` inside its recovery loop**. Schema validation happens one layer up, in `executor/agent.py`, *after* the provider has already returned — so the `ValidationError` is raised outside the loop that could have re-prompted the model.

The `hermes` provider does this correctly. `copilot` — the default provider — does not.

## Impact

A ~3.5 hour, ~$2.30 multi-agent run died at the **final** agent, after all prior agents had succeeded and a git commit had already been made. The failure was a single field on a single response, recoverable by one re-prompt that the codebase already knows how to construct (`_build_parse_recovery_prompt`).

```
Workflow Failed
Output field 'decision' has wrong type: expected string, got dict
💡 Suggestion: Ensure agent returns correct type for 'decision'
```

The agent declared:

```yaml
output:
  decision:
    type: string
    description: Review decision - APPROVE or REQUEST_CHANGES
```

…and the model returned `decision` as a nested object rather than the bare string. Everything else about the response was well-formed JSON.

This is materially worse than a parse failure, because it tends to surface with **non-Anthropic models** whose structured-output conventions differ, and it strands work that is otherwise complete.

## Environment

| | |
|---|---|
| Conductor | **v0.1.26** (latest release; also verified against `main`) |
| Provider | `copilot` |
| Affected agent model | `gpt-5.6-sol` (`reasoning: xhigh`, `context_tier: long_context`) |
| OS | Windows 11, PowerShell 7 |

Reproduced with a workflow whose other five agents ran on `claude-opus-5` / `claude-sonnet-5` without incident — only the agent on a non-Anthropic model tripped it.

## Root cause

Verified against `main` (paths below are current):

**1. Validation lives outside the provider.** `src/conductor/executor/agent.py:318`

```python
# Validate output against schema (skip for partial output from interrupts)
if agent.output and not output.partial:
    validate_output(output.content, agent.output)
```

[...]

PR #355, opened 2026-07-31
refactor(providers): run Claude through Pydantic AI

hertznsk:
## Summary

This PR replaces the hand-written inner agent loop in `ClaudeProvider` with a Pydantic AI-based runtime while preserving Conductor's public provider contract and workflow-level behavior.

It is primarily intended as a concrete way to move the architectural discussion in #315 forward. It is **not presented with an expectation that it should be merged as-is**. Whether this direction belongs in Conductor, and whether this implementation should be taken further, is entirely up to the maintainers.

## What changed

- delegates Claude agent execution, structured output, and tool iteration to Pydantic AI;
- keeps `AgentProvider`, `AgentOutput`, capability declarations, Conductor events, retries, interrupts, usage accounting, MCP policy, and output validation at the Conductor boundary;
- adds focused adapters for dynamic output models, event mapping, MCP tools, retries, interrupts, structured output, and usage;
- adapts the Claude provider and integration test suites to the new runtime seams;
- documents the new internal architecture and migration implications.

## Practical validation

I have spent several days using builds from this branch on real Conductor workflows and fixed the defects I encountered along the way. Apart from the issue described below, it has worked without problems for my tasks. Those tasks do not cover Conductor's complete feature surface, so this should not be read as exhaustive production validation.

## Known issue

One remaining defect should be fixed separately: #353. MCP connections can be opened in a worker task and later closed from the root task, which conflicts with AnyIO cancel-scope task affinity. The problem is in the shared `MCPManager` lifecycle and is not specific to Pydantic AI, so I deliberately did not expand this PR to address it.

## Skills follow-up

I also noticed that skills capabilities were added recently in #215. The Pydantic AI harness offers an opportunity to support this somewhat better than the current eager-preamble implementation. I did not expand the scope again to redesign that integration, but I can follow up with a separate PR if the maintainers are interested.

## Verification

- `make check`
- `make test`: 4556 passed; 27 environment/optional-integration tests skipped; 8 install-script tests deselected
- several days of use on real workflows, with the coverage limitation noted above

Related: #315, #353, #215


src/conductor/providers/_pydantic_ai/retry.py
```diff
@@ -0,0 +1,353 @@
...
+
+    if last_error is not None and type(last_error).__name__ == "UnexpectedModelBehavior":
+        suggestion = (
+            "Model repeatedly failed to produce valid structured output (output tool not called); "
+            "retry later or simplify the output schema"
+        )
+    else:
+        suggestion = f"Check API connectivity and rate limits. Last error: {last_error}"
+
+    raise ProviderError(
```
jrob5756, line 366: This breaks the parity contract from issue #343 (see AGENTS.md): once the retry budget is exhausted on a structured-output failure, we're supposed to re-raise the original `ValidationError` (it names the field and expected type) and reserve `ProviderError` for syntax failures.

Here `UnexpectedModelBehavior` is treated as retryable and, once exhausted, gets wrapped in a generic `ProviderError` with a vague "output tool not called" message. But pydantic-ai actually preserves the real `ValidationError` as `__cause__` on that exception - we're just not looking at it. A workflow author with a schema mismatch now gets "retry later or simplify the schema" instead of the actual field/type error, and it's misleading since retrying won't fix a shape mismatch.

Roughly:
```python
from pydantic import ValidationError as PydanticValidationError

cause = getattr(last_error, "__cause__", None)
if isinstance(cause, PydanticValidationError):
    raise ValidationError(str(cause)) from cause
```
before falling through to the existing `ProviderError` branch for everything else.

src/conductor/providers/_pydantic_ai/interrupt.py
```diff
@@ -0,0 +1,338 @@
...
+
+    This performs a hard cut: tools are not available on this run so the model
+    must answer directly.  The returned content is marked partial and is not
+    validated against the Conductor output schema.
+    """
+    emit_agent_turn_start(event_callback, "awaiting_model")
+    interrupt_message = _make_interrupt_message(has_output_schema)
+    partial_history = list(history)
+    partial_history.append(ModelRequest(parts=[interrupt_message]))
+    result = await agent.run(user_prompt=None, message_history=partial_history)
```
jrob5756: The docstring above says "tools are not available on this run so the model must answer directly," but `agent.run` here reuses the same `agent` built with `toolsets=` and a `ToolOutput` output type - nothing overrides either for this call. Pydantic AI's `run()` supports passing `toolsets=[]` / `output_type=str` per-call, which would actually enforce what the docstring claims.

As written, an interrupted structured-output agent can still try to call tools or get stuck on `ToolOutput` validation retries on this "final" call, and could raise `UnexpectedModelBehavior` instead of returning a partial result - which somewhat defeats the point of this code path.

src/conductor/providers/_pydantic_ai/agent_builder.py
```diff
@@ -0,0 +1,374 @@
...
+    """
+    model = _resolve_anthropic_model(
+        agent, default_model, api_key, base_url, auth_token=auth_token, timeout=timeout
+    )
+
+    output_type = _build_output_type(agent)
+    if output_type is None:
+        output_type = str
+
+    model_settings = _build_model_settings(
```
jrob5756, line 358: `_build_model_settings` never gets `default_model` here, even though `_resolve_anthropic_model` a few lines up does. Inside `_build_model_settings`, `model_name = (agent.model or None) or None` (line 269) means the thinking-support check falls back to the hardcoded `DEFAULT_ANTHROPIC_MODEL` whenever `agent.model` is unset - not the workflow/provider's actual configured default.

That means a workflow relying on `runtime.provider`'s default model together with `default_reasoning_effort` gets its thinking-budget validated against the wrong model, which can produce spurious `ValidationError`s (or miss a real one) depending on which models support extended thinking.

```suggestion
    model_settings = _build_model_settings(
        agent,
        default_temperature,
        default_max_tokens,
        default_reasoning_effort,
        default_model=default_model,
        timeout=timeout,
    )
```

src/conductor/providers/_pydantic_ai/structured_output.py
```diff
@@ -0,0 +1,112 @@
...
+
+    Returns:
+        Validated content dict.
+
+    Raises:
+        ProviderError: If no valid JSON matching the schema can be extracted.
+        ValidationError: If extracted JSON fails schema validation.
+    """
+    try:
+        content = parse_json_output(text)
```
jrob5756, line 106: This path parses and validates directly, but skips `unwrap_scalar_wrappers`/`normalize_agent_output` from `providers/_output_shape.py`, which `copilot.py` and `hermes.py` both apply before validating. That's the step that recovers single-field wrapper shapes like `{"value": ...}` or `{"result": ...}`.

Might be low-impact in practice since `_build_output_type` wraps schemas in `ToolOutput`, so this fallback may rarely get hit - but worth confirming, and if it can be reached, it should go through the same normalization the other providers use for parity.

src/conductor/providers/claude.py
```diff
@@ -921,536 +779,172 @@ async def execute(
...
+                try:
+                    return parse_json_output(partial_output)
+                except ValidationError:
+                    pass
+            return {"result": partial_output}
 
-        return None
+        return {"result": partial_output}
 
     def _coerce_for_thinking(
```
jrob5756: This method looks unused - `execute()` and `execute_dialog_turn()` call the module-level `_coerce_for_thinking` imported from `agent_builder.py` instead (see line 699), not `self._coerce_for_thinking`. Looks like leftover from the refactor. If it's truly dead, removing it would also let you drop the now-unnecessary `CLAUDE_ANSWER_HEADROOM_TOKENS`/`CLAUDE_EXTENDED_THINKING_OUTPUT_CAP` imports at the top of the file (lines 42-43).

src/conductor/providers/claude.py
```diff
@@ -921,536 +779,172 @@ async def execute(
...
+                    has_output_schema=bool(agent.output),
+                    usage_limits=UsageLimits(request_limit=max_iterations),
+                    max_session_seconds=max_session,
+                    max_parse_recovery_attempts=retry_cfg.max_parse_recovery_attempts,
+                ),
+                retry_config=retry_cfg,
+                event_callback=intercepting_callback,
+                agent_name=agent.name,
+            )
+        finally:
```
jrob5756:
```suggestion
```
No-op `finally: pass` - looks like refactor debris, safe to remove.

src/conductor/providers/_pydantic_ai/retry.py
```diff
@@ -0,0 +1,353 @@
...
+except ImportError:
+    anthropic = None  # type: ignore[assignment]
+
+from conductor.config.schema import RetryPolicy
+from conductor.exceptions import ProviderError, ValidationError
+
+logger = logging.getLogger(__name__)
+
+EventCallback = Callable[[str, dict[str, Any]], None]
+
```
jrob5756, line 32: `EventCallback` is defined twice in a row here (identical to line 30). Probably a merge artifact - one can be dropped.

src/conductor/providers/_pydantic_ai/mcp_toolset.py
```diff
@@ -0,0 +1,161 @@
...
+
+        Raises:
+            ToolFailed: When the underlying MCP tool call fails, so the agent loop
+                records an ``outcome='failed'`` tool return that Anthropic maps to
+                ``is_error=True``.
+        """
+        try:
+            result = await self._manager.call_tool(name, tool_args)
+        except Exception as e:
+            raise ToolFailed(f"Error executing tool {name!r}: {e}") from e
```
jrob5756, line 132: This re-raises as `ToolFailed` so the model sees the error, which matches the old provider's behavior - but there's no Conductor-side log here. A genuine adapter bug (vs. a legitimate tool-level failure) will only ever show up in the model's transcript, which most subscribers/dashboards don't surface. Worth a `logger.warning` before re-raising so operators can tell the two apart without digging through transcripts.

src/conductor/providers/_pydantic_ai/agent_builder.py
```diff
@@ -0,0 +1,374 @@
...
+        max_parse_recovery_attempts: Output correction retries handled inside
+            the Pydantic AI agent.
+        api_key: Anthropic API key. Falls back to ``ANTHROPIC_API_KEY`` env var.
+        auth_token: Optional bearer-auth token for gateway / LiteLLM endpoints.
+        base_url: Optional custom API endpoint.
+        timeout: Request timeout in seconds. ``None`` lets the Anthropic SDK use
+            its own default.
+        toolsets: Optional Pydantic AI toolsets to register. Used by Phase 4 to
+            inject MCP toolsets without changing this signature.
+        tools: Optional plain Pydantic AI tools to register. Also reserved for
```
jrob5756: "Reserved for Phase 4" is already inaccurate - toolsets are wired up in this same PR (`claude.py` passes `MCPManagerToolset` into `build_agent(toolsets=...)`). Also worth dropping the internal phase reference and just describing the current contract, e.g. "Pydantic AI toolsets to register (e.g. the MCP tool bridge)."

jrob5756 (commented):
Reviewed the Claude/Pydantic AI refactor. Overall this is a clean split of responsibilities across the new _pydantic_ai package, and the test suite is much better organized than the old monolithic test_claude.py. I found three issues worth blocking on (mainly the structured-output recovery contract from issue https://github.com/microsoft/conductor/issues/343 no longer holding), plus some smaller cleanup items inline. Happy to discuss any of these.

src/conductor/providers/_pydantic_ai/retry.py
hertznsk: Fixed. On retry-budget exhaustion, when `UnexpectedModelBehavior` carries a pydantic `ValidationError` as `__cause__`, we now re-raise it as Conductor `ValidationError` (naming the field and expected type) via `raise ... from cause`, so the original error stays available for introspection. `ProviderError` with the "output tool not called" suggestion is kept only for the case where no schema-validation cause is preserved. A new test asserts both the message content and `__cause__ is schema_error`.

src/conductor/providers/_pydantic_ai/interrupt.py
hertznsk: Good catch — enforced rather than documented. The partial-result call now runs inside `agent.override(toolsets=[], tools=[])` with `output_type=str`. One pydantic-ai subtlety worth noting: the `toolsets=` run kwarg is additive in 2.18 (it never clears construction-time toolsets — verified empirically that an MCP-style toolset stays callable through `run(toolsets=[])`), so `agent.override` is the only mechanism that actually removes them. Since `output_type=str` removes the output tool from the request, the structured interrupt prompt was also reworded from "call the final_result tool" to "respond with plain text; do not call any tools" — asking for a tool that no longer exists would just produce a failed tool call instead of a partial result. Trade-off rationale: this path is an escape hatch whose contract is best-effort delivery after Esc; partial outputs are never schema-validated downstream (`_build_partial_content` handles both str with lenient JSON extraction and BaseModel), so guaranteed delivery beats shape fidelity here. Covered by a semantic test where the scripted model attempts a tool call on the partial run and the tool's side effect must not happen.

src/conductor/providers/_pydantic_ai/agent_builder.py
hertznsk: Fixed exactly as suggested: `_build_model_settings` now takes `default_model` and resolves `(agent.model or default_model) or DEFAULT_ANTHROPIC_MODEL`, with `build_agent` forwarding it. Two new tests cover both directions — a thinking-capable default model is accepted when `agent.model` is unset, and a non-thinking default raises `ValidationError` instead of silently validating against the library default.

src/conductor/providers/_pydantic_ai/structured_output.py
hertznsk: Fixed — applied `normalize_agent_output` before `validate_output` in both places where provider output is validated: `parse_text_fallback` (your comment target) and the `BaseModel` path in `extract_content`, which had the same gap for test-constructed/edge outputs. On reachability: you're right that ToolOutput rejects wrapper shapes upstream most of the time, but the fallback is reachable whenever the model answers a tool-output schema with plain text containing JSON, so parity is worth having. Added tests for both paths (wrapper object unwrapped to the scalar in each).

src/conductor/providers/claude.py
hertznsk: Confirmed dead — removed the method along with the `CLAUDE_ANSWER_HEADROOM_TOKENS`/`CLAUDE_EXTENDED_THINKING_OUTPUT_CAP` imports (verified no remaining references; the live call site at the dialog-turn path uses the module-level `_coerce_for_thinking` from `agent_builder.py`).

src/conductor/providers/claude.py
hertznsk: Removed — the `try/finally: pass` is gone and `execute_with_retry` is awaited directly (the finally block was empty, so no cleanup semantics changed).

src/conductor/providers/_pydantic_ai/retry.py
hertznsk: Dropped the duplicate definition.

src/conductor/providers/_pydantic_ai/mcp_toolset.py
hertznsk: Added `logger.warning("MCP tool call %r failed: %s: %s", name, type(e).__name__, e)` immediately before the `ToolFailed` raise, so operator logs distinguish adapter/tool failures from model behavior without digging through transcripts.

src/conductor/providers/_pydantic_ai/agent_builder.py
hertznsk: Reworded — no more phase references: "Optional Pydantic AI toolsets to register (e.g. the MCP tool bridge)" / "Optional plain Pydantic AI tools to register."

src/conductor/providers/_pydantic_ai/interrupt.py
hertznsk: To be honest, I'm not sure if this is the right solution, but I don't have any other suggestions yet.

jrob5756 (approved):
LGTM. Approved!

PR #358, opened 2026-08-01
fix(mcp): keep connection cleanup in owner tasks

### LINKED ISSUE #353: MCP cleanup exits AnyIO cancel scopes from a different asyncio task

hertznsk:
## Summary

- move each MCP stdio/session context lifecycle into a persistent owner task
- signal owner tasks during manager shutdown so AnyIO cancel scopes exit in the task that entered them
- preserve caller cancellation while consuming and logging cleanup failures
- protect owner-task teardown when manager shutdown is cancelled
- reject duplicate server names without orphaning the existing connection

## Testing

- `PYTHONASYNCIODEBUG=1 uv run pytest tests/test_mcp -q` (51 passed)
- `make check`
- direct stdio MCP lifecycle driver covering connect, tool call, duplicate names, failure, cancellation, concurrency, and idempotent close

Fixes #353

### DIFF
--- a/src/conductor/mcp/manager.py
+++ b/src/conductor/mcp/manager.py
@@ -125,7 +125,6 @@ class MCPManager:
         self.tool_to_server: dict[str, str] = {}  # prefixed_name -> server
         self._connection_tasks: dict[str, asyncio.Task[None]] = {}
         self._connection_stops: dict[str, asyncio.Event] = {}
-        self._initialized = False
         self._tool_output = tool_output or ToolOutputConfig()
 
     async def connect_server(
@@ -214,7 +213,6 @@ class MCPManager:
 
                     self.sessions[name] = session
                     self.tools[name] = tools
-                    self._initialized = True
                     ready.set_result(tools)
                     await stop.wait()
             except asyncio.CancelledError:
@@ -240,12 +238,19 @@ class MCPManager:
                 ready.cancel()
             task.cancel()
             cleanup = asyncio.gather(task, return_exceptions=True)
-            try:
-                results = await asyncio.shield(cleanup)
-            except asyncio.CancelledError:
-                results = await cleanup
+            # Await the owner task's teardown under a re-shielding loop: a
+            # repeated cancellation of connect_server() lands on the shield
+            # instead of the gather, so cleanup still completes and the
+            # bookkeeping below always runs.
+            results: list[BaseException | None] | None = None
+            while results is None:
+                try:
+                    results = list(await asyncio.shield(cleanup))
+                except asyncio.CancelledError:
+                    if cleanup.done():
+                        results = list(cleanup.result())
             for result in results:
-                if isinstance(result, Exception) and not isinstance(result, asyncio.CancelledError):
+                if isinstance(result, Exception):
                     logger.warning(f"Error cancelling MCP connection '{name}': {result}")
             self._connection_tasks.pop(name, None)
             self._connection_stops.pop(name, None)
@@ -256,7 +261,7 @@ class MCPManager:
             self._connection_tasks.pop(name, None)
             self._connection_stops.pop(name, None)
             self._discard_server_state(name)
-            logger.error(f"Failed to connect to MCP server '{name}': {exc}")
+            logger.error(f"Failed to connect to MCP server '{name}': {exc}", exc_info=exc)
             raise RuntimeError(f"Failed to connect to MCP server '{name}': {exc}") from exc
 
         logger.info(
@@ -555,36 +560,36 @@ class MCPManager:
 
         This method should be called when the manager is no longer needed.
         It properly closes all stdio connections and cleans up internal state.
+        Cancellation is deliberately absorbed until owner-task teardown
+        completes, so task-affine MCP contexts are never orphaned.
         """
         if not self._connection_tasks:
             return
 
-        logger.debug(f"Closing {len(self.sessions)} MCP server connection(s)")
+        logger.debug(f"Closing {len(self._connection_tasks)} MCP server connection(s)")
 
         for stop in self._connection_stops.values():
             stop.set()
 
         cleanup = asyncio.gather(*self._connection_tasks.values(), return_exceptions=True)
-        try:
+        # Await teardown under a re-shielding loop: a repeated cancellation of
+        # close() lands on the shield instead of the gather, so cleanup still
+        # completes before bookkeeping is cleared.
+        results: list[BaseException | None] | None = None
+        while results is None:
             try:
-                results = await asyncio.shield(cleanup)
+                results = list(await asyncio.shield(cleanup))
             except asyncio.CancelledError:
-                results = await cleanup
-                for result in results:
-                    if isinstance(result, Exception):
-                        logger.warning(f"Error closing MCP connections: {result}")
-                raise
+                if cleanup.done():
+                    results = list(cleanup.result())
 
-            for result in results:
-                if isinstance(result, Exception):
-                    logger.warning(f"Error closing MCP connections: {result}")
-        finally:
-            if cleanup.done():
-                self.sessions.clear()
-                self.tools.clear()
-                self.tool_to_server.clear()
-                self._connection_tasks.clear()
-                self._connection_stops.clear()
-                self._initialized = False
+        self.sessions.clear()
+        self.tools.clear()
+        self.tool_to_server.clear()
+        self._connection_tasks.clear()
+        self._connection_stops.clear()
+        for result in results:
+            if isinstance(result, Exception):
+                logger.warning(f"Error closing MCP connections: {result}")
 
         logger.debug("MCP manager closed")
--- a/src/conductor/mcp/manager.py
+++ b/src/conductor/mcp/manager.py
@@ -15,6 +15,7 @@ part of the returned string.
 
 from __future__ import annotations
 
+import asyncio
 import logging
 import os
 import re
@@ -35,13 +36,13 @@ try:
     from mcp.client.stdio import StdioServerParameters, stdio_client
     from mcp.types import TextContent
 
-    MCP_SDK_AVAILABLE = True
 except ImportError:
-    MCP_SDK_AVAILABLE = False
     ClientSession = None  # type: ignore[misc, assignment]
     StdioServerParameters = None  # type: ignore[misc, assignment]
     stdio_client = None  # type: ignore[misc, assignment]
     TextContent = None  # type: ignore[misc, assignment]
+
+MCP_SDK_AVAILABLE = ClientSession is not None
 
 
 # Marker constants. The generic hint is embedded by the manager and replaced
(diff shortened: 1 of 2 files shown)


src/conductor/mcp/manager.py
```diff
@@ -491,25 +539,52 @@ def has_servers(self) -> bool:
...
+                        logger.warning(f"Error closing MCP connections: {result}")
+                raise
 
-        self.sessions.clear()
-        self.tools.clear()
-        self.tool_to_server.clear()
-        self._initialized = False
+            for result in results:
+                if isinstance(result, Exception):
+                    logger.warning(f"Error closing MCP connections: {result}")
```
jrob5756: If close() gets cancelled a second time while it's sitting at the unshielded `await cleanup` on line 573, that cancellation propagates into the still-running owner tasks and can interrupt teardown mid-flight (e.g. while a subprocess is being terminated). But `cleanup.done()` is true even for a cancelled future, so this `finally` block clears all bookkeeping as if everything shut down cleanly, and there's no log line for this path at all. Worth tracking whether `results` was actually obtained before treating state as cleared, and logging when a repeated cancellation prevents confirming cleanup completed.

src/conductor/mcp/manager.py
```diff
@@ -163,59 +165,105 @@ async def connect_server(
...
+        try:
+            tools = await asyncio.shield(ready)
+        except asyncio.CancelledError:
+            if not ready.done():
+                ready.cancel()
+            task.cancel()
+            cleanup = asyncio.gather(task, return_exceptions=True)
+            try:
+                results = await asyncio.shield(cleanup)
+            except asyncio.CancelledError:
```
jrob5756: Same gap here as in close(): if a second cancellation lands while awaiting the unshielded `await cleanup` on this line, it propagates straight out and skips the pop()/`_discard_server_state()` calls below. That leaves `_connection_tasks[name]` populated with a task that may not have finished cleanup, so subsequent `connect_server(name=...)` calls for that name will incorrectly raise "already connected or connecting" until the whole manager is torn down, with no log explaining why.

src/conductor/mcp/manager.py
```diff
@@ -163,59 +165,105 @@ async def connect_server(
...
+            self._connection_tasks.pop(name, None)
+            self._connection_stops.pop(name, None)
+            self._discard_server_state(name)
+            raise
+        except Exception as exc:
+            await asyncio.gather(task, return_exceptions=True)
+            self._connection_tasks.pop(name, None)
+            self._connection_stops.pop(name, None)
+            self._discard_server_state(name)
+            logger.error(f"Failed to connect to MCP server '{name}': {exc}")
```
jrob5756: `{exc}` only calls `str()`. If the `AsyncExitStack` teardown raises a different exception while unwinding after the real connect failure (e.g. a broken pipe during `stdio_client`/`ClientSession` exit), that secondary exception replaces the original as what's caught and logged here, masking the actual root cause.

```suggestion
            logger.error(f"Failed to connect to MCP server '{name}': {exc}", exc_info=exc)
```

src/conductor/mcp/manager.py
```diff
@@ -163,59 +165,105 @@ async def connect_server(
...
+            if not ready.done():
+                ready.cancel()
+            task.cancel()
+            cleanup = asyncio.gather(task, return_exceptions=True)
+            try:
+                results = await asyncio.shield(cleanup)
+            except asyncio.CancelledError:
+                results = await cleanup
+            for result in results:
+                if isinstance(result, Exception) and not isinstance(result, asyncio.CancelledError):
```
jrob5756: `asyncio.CancelledError` derives from `BaseException`, not `Exception`, since Python 3.8, so `isinstance(result, Exception)` already excludes it. The trailing check is dead code.

```suggestion
                if isinstance(result, Exception):
```

src/conductor/mcp/manager.py
```diff
@@ -119,10 +120,11 @@ def __init__(self, tool_output: ToolOutputConfig | None = None) -> None:
...
             raise ImportError("MCP SDK not installed. Install with: uv add 'mcp>=1.0.0'")
 
-        self.sessions: dict[str, ClientSession] = {}
+        self.sessions: dict[str, Any] = {}
         self.tools: dict[str, list[dict[str, Any]]] = {}  # server -> tools
         self.tool_to_server: dict[str, str] = {}  # prefixed_name -> server
-        self._exit_stack = AsyncExitStack()
+        self._connection_tasks: dict[str, asyncio.Task[None]] = {}
+        self._connection_stops: dict[str, asyncio.Event] = {}
         self._initialized = False
```
jrob5756: `_initialized` is set here and at two other spots but nothing reads it anymore — `close()`'s guard now checks `self._connection_tasks` instead. Worth removing it so a future reader doesn't assume it reflects live connection state.

src/conductor/mcp/manager.py
```diff
@@ -491,25 +539,52 @@ def has_servers(self) -> bool:
...
+
     async def close(self) -> None:
         """Close all server connections and clean up resources.
 
         This method should be called when the manager is no longer needed.
         It properly closes all stdio connections and cleans up internal state.
         """
-        if not self._initialized:
+        if not self._connection_tasks:
             return
```
jrob5756, line 568: This counts `len(self.sessions)`, which is only populated once a connection's `ready` future resolves. If `close()` runs while a `connect_server()` call is still mid-handshake, this understates how many owner tasks are actually being torn down. `len(self._connection_tasks)` would be accurate.

jrob5756 (commented):
Good fix overall. The owner-task-per-connection model is the right way to solve the anyio cancel-scope task affinity problem, and the new task-affinity tests are a solid way to prove it. 

Left a few small comments on double-cancellation edge cases and a couple of small cleanups, but nothing blocking here.

hertznsk (commented):
All six review comments are addressed in ca8551d. Per-thread replies via the review-comment API return 404 for my token (fork-based PR), so summarizing here:

1. close() double-cancellation — the unshielded `await cleanup` fallback is replaced with a re-shielding loop: repeated cancellations land on a throwaway `asyncio.shield(cleanup)` while the gather keeps running, and bookkeeping is cleared only after `results` is actually obtained. close() now deliberately absorbs cancellation until owner-task teardown completes (documented in the docstring); covered by `test_double_cancelled_close_still_finishes_cleanup`.
2. connect_server() double-cancellation — same re-shielding loop: cleanup completes and the `pop()`/`_discard_server_state()` calls always run before the cancellation is re-raised, so `_connection_tasks[name]` can no longer be left populated and block reconnects; covered by `test_double_cancelled_connection_clears_bookkeeping`.
3. Connect-failure logging — now passes `exc_info=exc`, so a secondary AsyncExitStack teardown exception cannot mask the original root cause.
4. Dead `not isinstance(result, asyncio.CancelledError)` check removed.
5. Unused `_initialized` flag removed from all three spots.
6. Shutdown log now counts `len(self._connection_tasks)`, including mid-handshake connections.

Full test suite passes (4580 tests), ruff and ty clean.

src/conductor/mcp/manager.py
hertznsk: Fixed in ca8551d - the debug log now counts len(self._connection_tasks), which includes connections still mid-handshake.

src/conductor/mcp/manager.py
hertznsk: Fixed in ca8551d. The unshielded `await cleanup` fallback is replaced with a re-shielding loop: repeated cancellations now land on a throwaway `asyncio.shield(cleanup)` while the gather keeps running, and bookkeeping is cleared only after `results` is actually obtained (a cancelled gather result is never read). The behavior goes slightly further than suggested: close() now deliberately absorbs cancellation until owner-task teardown completes (documented in the docstring), so teardown is never interrupted mid-flight and state is always consistent. Covered by test_double_cancelled_close_still_finishes_cleanup.

src/conductor/mcp/manager.py
hertznsk: Fixed in ca8551d with the same re-shielding loop as in close(): a repeated cancellation lands on the shield instead of the gather, so cleanup completes and the pop()/_discard_server_state() calls always run before the cancellation is re-raised. _connection_tasks[name] can no longer be left populated, so a later connect_server(name=...) will not be spuriously rejected. Covered by test_double_cancelled_connection_clears_bookkeeping.

src/conductor/mcp/manager.py
hertznsk: Applied in ca8551d - the log call now passes exc_info=exc, so the original connect failure traceback is preserved even when AsyncExitStack teardown raises a secondary exception while unwinding.

src/conductor/mcp/manager.py
hertznsk: Applied in ca8551d - the dead CancelledError check is removed; the condition is now plain isinstance(result, Exception).

src/conductor/mcp/manager.py
hertznsk: Removed in ca8551d - _initialized is gone from all three spots; close() keeps guarding on _connection_tasks.

jrob5756 (approved):
LGTM. Approved!

PR #359, opened 2026-08-03
fix(engine): serialize validator and dialog-evaluator prompts with ensure_ascii=False

### LINKED ISSUE #356: Validator truncation disproportionately reduces context for non-ASCII output
## Description

`OutputValidator.validate()` serializes the primary agent output with `json.dumps()` using the default `ensure_ascii=True`, and only then applies the fixed `_OUTPUT_LIMIT = 8000` character budget.

This expands non-ASCII text into JSON Unicode escape sequences before truncation. For example:

```text
A  -> A       (1 character)
П  -> \u041f  (6 characters)
你 -> \u4f60  (6 characters)
😀 -> \ud83d\ude00 (12 characters)
```

As a result, validator calls receive substantially less source material for Cyrillic, CJK, and other non-Latin output than for equivalent ASCII output. This can affect the grading decision when relevant content is truncated solely because of the language or script used. Truncation can also split a `\uXXXX` escape sequence, leaving malformed JSON in the embedded `AGENT OUTPUT` section.

The Unicode data itself is not corrupted when the complete string is decoded; the issue is the language-dependent effective validation budget caused by escaping before truncation.

## Location

Confirmed on current `main` at commit `f39bd63c3f8fe540f164102ebedb598526b2f686`:

```python
output_str = json.dumps(primary_output, indent=2, default=str)
# ...
primary_output=_truncate(output_str, _OUTPUT_LIMIT),
```

https://github.com/microsoft/conductor/blob/f39bd63c3f8fe540f164102ebedb598526b2f686/src/conductor/engine/validator.py

## Minimal reproduction

```python
import json

ascii_output = json.dumps({"text": "A" * 2000})
cjk_output = json.dumps({"text": "你" * 2000})

print(len(ascii_output))  # approximately 2,012
print(len(cjk_output))    # approximately 12,012
```

The ASCII output fits within the 8,000-character validator budget, while the same number of CJK code points is truncated.

## Expected behavior

The effective validator input budget should not be disproportionately reduced for non-ASCII languages. The embedded output should remain readable and should not be truncated inside JSON escape sequences.

## Possible direction

Serialize the validator input with `ensure_ascii=False` before applying the character limit:

```python
json.dumps(primary_output, indent=2, default=str, ensure_ascii=False)
```

A regression test could compare equivalent ASCII and non-ASCII output and verify that non-ASCII characters remain unescaped in the rendered validator prompt.

hertznsk:
## Summary

Fixes #356.

`OutputValidator.validate()` serialized the primary agent output with `json.dumps(primary_output, indent=2, default=str)` (default `ensure_ascii=True`) and only then applied the fixed `_OUTPUT_LIMIT = 8000` character budget. Non-ASCII text inflates into `\uXXXX` escapes before truncation — 6 chars per Cyrillic/CJK code point, 12 per emoji — so non-English outputs reached the grader with up to ~6x less source material than an equivalent ASCII output, and the cut could land inside an escape sequence, leaving malformed JSON in the embedded `AGENT OUTPUT` section.

## Changes

- **`src/conductor/engine/validator.py`** — serialize with `ensure_ascii=False` so the 8000-char budget is measured in real characters for every language and truncation cannot split an escape sequence.
- **`src/conductor/engine/dialog_evaluator.py`** — a codebase scan found the identical serialize-then-truncate pattern feeding the dialog-trigger evaluator prompt (4000-char budget); fixed the same way.
- **Regression tests** (`test_validator.py`, `test_dialog_evaluator.py`) — CJK output that fits the budget unescaped (~2 KB / ~1 KB) but would be truncated escaped (~12 KB / ~6 KB) must reach the prompt unescaped, with no `\uXXXX` escapes and no truncation marker.

Other `json.dumps` call sites were reviewed and intentionally left alone: they are system serialization (checkpoints, NDJSON event logs, CLI argv, cache files) or prompt content without fixed-budget truncation, where ASCII escaping has no fairness or correctness impact.

## Test plan

- [x] `pytest tests/test_engine/test_validator.py tests/test_engine/test_validator_integration.py tests/test_engine/test_dialog_evaluator.py` — 61/61 pass
- [x] `ruff check` / `ruff format --check` / `ty check` — clean on all touched files

### DIFF
--- a/src/conductor/engine/dialog_evaluator.py
+++ b/src/conductor/engine/dialog_evaluator.py
@@ -123,7 +123,10 @@ class DialogEvaluator:
             DialogEvaluation with trigger decision and opening question.
         """
         try:
-            output_str = json.dumps(output, indent=2, default=str)
+            # ``ensure_ascii=False`` so the 4000-char budget below is measured
+            # in real characters for every language — same truncation-fairness
+            # fix as the output validator (issue #356).
+            output_str = json.dumps(output, indent=2, default=str, ensure_ascii=False)
         except (TypeError, ValueError):
             output_str = str(output)
 
--- a/src/conductor/engine/validator.py
+++ b/src/conductor/engine/validator.py
@@ -164,7 +164,13 @@ class OutputValidator:
             return ValidationOutcome(passed=True)
 
         try:
-            output_str = json.dumps(primary_output, indent=2, default=str)
+            # ``ensure_ascii=False`` keeps non-ASCII text literal so the fixed
+            # ``_OUTPUT_LIMIT`` budget is measured in real characters for every
+            # language — with the default ``ensure_ascii=True`` each Cyrillic /
+            # CJK code point inflates to 6 (and emoji to 12) ``\uXXXX`` chars
+            # before truncation, shrinking the effective budget ~6x and letting
+            # the cut land inside an escape sequence (issue #356).
+            output_str = json.dumps(primary_output, indent=2, default=str, ensure_ascii=False)
         except (TypeError, ValueError):
             output_str = str(primary_output)
 
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -2432,7 +2432,9 @@ class WorkflowEngine:
         last_output_preview: str | None = None
         if last_output is not None:
             try:
-                preview = json.dumps(last_output, indent=2, default=str)
+                # ``ensure_ascii=False`` so the preview shows real non-ASCII
+                # text instead of \uXXXX escapes (issue #356).
+                preview = json.dumps(last_output, indent=2, default=str, ensure_ascii=False)
                 last_output_preview = preview[:500]
             except (TypeError, ValueError):
                 last_output_preview = str(last_output)[:500]
@@ -2502,7 +2504,11 @@ class WorkflowEngine:
             return False
 
         try:
-            preview = json.dumps(partial_output.content, indent=2, default=str)[:500]
+            # ``ensure_ascii=False`` so the preview shows real non-ASCII
+            # text instead of \uXXXX escapes (issue #356).
+            preview = json.dumps(partial_output.content, indent=2, default=str, ensure_ascii=False)[
+                :500
+            ]
         except (TypeError, ValueError):
             preview = str(partial_output.content)[:500]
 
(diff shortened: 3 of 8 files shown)


jrob5756 (commented):
Nice fix, and the tests actually prove it, I reverted ensure_ascii=False locally and both new tests fail as expected!

One gap worth a follow-up: the same escaping bug still lives in `src/conductor/gates/dialog.py` (lines 222, 388, 612), all doing `json.dumps(agent_output, indent=2, default=str)` with no `ensure_ascii=False`. Lines 222 and 388 feed the same output into the dialog-mode LLM system prompt, so non-ASCII agent output reaches the model escaped. Line 612 is worse in a different way: it renders straight into the console panel a person is looking at during a dialog session, so someone running a workflow on Cyrillic or CJK output would see literal `\u4f60\u597d` instead of the real text. None of these have a fixed truncation budget, so it's not the exact mid-escape-cut bug from #356, but it's the same root cause and the same one-line fix.

Lower priority but same pattern: `src/conductor/engine/workflow.py` lines 2435, 2505, 2630 build `[:500]` log/console previews the same escaped way, so non-ASCII previews can come out garbled. Cosmetic, but worth folding in while you're touching this.

Neither is blocking for this PR's stated scope, just flagging so #356 doesn't need a part two.

hertznsk (commented):
Thanks for the thorough check — great catch on the remaining call sites. I've folded them into this PR in c910c2b rather than leaving them for a follow-up:

- `gates/dialog.py` (all three): the CLI + web dialog system prompts and the console "Agent Output" panel now serialize with `ensure_ascii=False`, so the dialog-mode LLM and the person at the terminal both see real non-ASCII text.
- `engine/workflow.py` (all three): the `[:500]` previews (interrupt handler preview and the `agent_paused` partial-content previews) now serialize the same way.

Regression tests added for each surface — CLI dialog system prompt, web dialog system prompt, the console panel renderable, and the interrupt preview — with mixed Cyrillic/CJK output, asserting the literal text is present and no `\\u4f60`-style escapes leak. I also verified each new test fails when the fix is reverted, same as you did for the original pair.

`make check` (ruff + ty) and the engine/gates/executor test suites are green (1247 passed, 1 skipped).

jrob5756 (approved):
LGTM. Approved!

PR #367, opened 2026-08-03
fix(engine): drop tracebacks from warning logs on handled fail-open paths

### LINKED ISSUE #357: Validator fail-open re-run failures print a full traceback
## Description

When a semantic validator rejects an agent output and the validator-triggered primary re-run fails, Conductor correctly fails open and keeps the original output. However, the expected fail-open path is logged with `exc_info=True`, so users see a full Python traceback even though the workflow continues successfully.

This makes a handled validator recovery failure look like an unhandled workflow crash and adds noisy implementation details to normal CLI output.

## Reproduction

On current `main` (`f39bd63c3f8fe540f164102ebedb598526b2f686`):

1. Configure an agent with `validator.max_retries: 1`.
2. Return `passed: false` from the validator.
3. Make the validator-triggered primary re-run raise a provider/runtime exception.
4. Run the workflow with warning logs visible.

The existing regression scenario demonstrates the behavior directly:

```bash
uv run pytest tests/test_engine/test_validator_integration.py::TestValidatorCostAndFailurePaths::test_rerun_failure_keeps_original_no_double_count_and_emits_event -q -s --log-cli-level=WARNING
```

Observed output includes:

```text
WARNING conductor.engine.workflow: Validator re-run failed for 'reviewer'; using original output
Traceback (most recent call last):
  ...
RuntimeError: rerun boom
PASSED
```

## Expected behavior

The workflow should still fail open and retain the original output, but the handled re-run failure should be reported as a concise warning without a full traceback in normal user-facing logs. Diagnostic exception details can remain available at debug level or in structured diagnostic metadata/logs.

## Actual behavior

`WorkflowEngine._apply_validator()` catches the exception and returns the original output, but logs the warning with `exc_info=True` in `src/conductor/engine/workflow.py`. This prints the entire traceback at WARNING level despite the failure being handled.

## Additional context

The existing test verifies output preservation, usage accounting, and the `rerun_errored` event, but does not assert that the user-facing fail-open path avoids traceback noise.

hertznsk:
## Summary

Fixes #357.

When a semantic validator rejects an agent output and the validator-triggered primary re-run fails, Conductor correctly fails open and keeps the original output — but logged the warning with `exc_info=True`, so users saw a full Python traceback at WARNING level even though the workflow continued successfully. This made a handled validator recovery failure look like an unhandled workflow crash.

The warning is now concise (exception type and message included), and the full traceback is kept at DEBUG level for diagnosis.

## Changes

- **`engine/workflow.py` — `_apply_validator`**: the re-run fail-open path (the exact scenario from the issue) now logs a concise warning without `exc_info`; the traceback moves to `logger.debug(..., exc_info=True)`.
- **Sibling handled/best-effort paths** brought to the same pattern, found by a codebase sweep for `logger.warning(..., exc_info=True)` on fail-open paths:
  - `engine/validator.py` and `engine/workflow.py` — validator call failure / timeout (also documented fail-open behavior, see `docs/workflow-syntax.md`)
  - `engine/workflow.py` — provider session-ID collection for checkpoints
  - `engine/checkpoint.py` — checkpoint save failure and rotation/cleanup listing failure
- **`providers/hermes.py` and `providers/_event_format.py`**: event-callback errors move from `logger.warning(..., exc_info=True)` to `logger.debug(..., exc_info=True)`, matching the existing convention in `claude_agent_sdk.py::_safe_callback` and `_pydantic_ai/events.py` (provider parity).

No documentation or frontend changes were needed: `docs/workflow-syntax.md` already describes validator fail-open as "treated as a pass (with a logged warning)", and the `rerun_errored` event handling in the dashboard is unchanged.

## Test plan

- Extended `test_rerun_failure_keeps_original_no_double_count_and_emits_event` with `caplog` assertions: exactly one WARNING record (no `exc_info`, message contains the exception type and text), and the traceback present at DEBUG level.
- The reproduction from the issue now prints a single concise line instead of a traceback:
  ```
  WARNING conductor.engine.workflow: Validator re-run failed for 'reviewer'; using original output (RuntimeError: rerun boom)
  ```
- Full suite: **4593 passed, 37 skipped**; `ruff check` and `ty check` clean on all touched files (remaining diagnostics pre-exist on `main`).

### DIFF
--- a/src/conductor/engine/checkpoint.py
+++ b/src/conductor/engine/checkpoint.py
@@ -295,8 +295,9 @@ class CheckpointManager:
 
             return checkpoint_path
 
-        except Exception:
-            logger.warning("Failed to save checkpoint", exc_info=True)
+        except Exception as exc:
+            logger.warning("Failed to save checkpoint (%s: %s)", type(exc).__name__, exc)
+            logger.debug("Checkpoint save traceback", exc_info=True)
             return None
 
     @staticmethod
@@ -510,8 +511,11 @@ class CheckpointManager:
         """
         try:
             candidates = CheckpointManager._periodic_checkpoints_for_run(workflow_path, run_id)
-        except Exception:
-            logger.warning("Failed to list checkpoints for %s", action, exc_info=True)
+        except Exception as exc:
+            logger.warning(
+                "Failed to list checkpoints for %s (%s: %s)", action, type(exc).__name__, exc
+            )
+            logger.debug("Checkpoint listing traceback", exc_info=True)
             return
         # list_checkpoints sorts newest-first, so anything past keep_last is old.
         for cp in candidates[keep_last:]:
--- a/src/conductor/engine/checkpoint.py
+++ b/src/conductor/engine/checkpoint.py
@@ -513,7 +513,12 @@ class CheckpointManager:
             candidates = CheckpointManager._periodic_checkpoints_for_run(workflow_path, run_id)
         except Exception as exc:
             logger.warning(
-                "Failed to list checkpoints for %s (%s: %s)", action, type(exc).__name__, exc
+                "Failed to list checkpoints for %s (run_id=%s, workflow=%s) (%s: %s)",
+                action,
+                run_id,
+                workflow_path,
+                type(exc).__name__,
+                exc,
             )
             logger.debug("Checkpoint listing traceback", exc_info=True)
             return
--- a/src/conductor/engine/validator.py
+++ b/src/conductor/engine/validator.py
@@ -192,12 +192,14 @@ class OutputValidator:
         except asyncio.CancelledError:
             # Interrupt / cancellation must propagate — never silently pass.
             raise
-        except Exception:
+        except Exception as exc:
             logger.warning(
-                "Validator call failed for agent '%s'; treating as pass",
+                "Validator call failed for agent '%s'; treating as pass (%s: %s)",
                 agent.name,
-                exc_info=True,
+                type(exc).__name__,
+                exc,
             )
+            logger.debug("Validator call traceback for agent '%s'", agent.name, exc_info=True)
             return ValidationOutcome(passed=True, errored=True)
 
         passed, issues, parse_ok = self._parse(output.content)
--- a/src/conductor/engine/workflow.py
+++ b/src/conductor/engine/workflow.py
@@ -2138,8 +2138,13 @@ class WorkflowEngine:
                         if hasattr(p, "get_session_cwds"):
                             copilot_session_cwds = p.get_session_cwds()  # type: ignore[union-attr]
                         break
-        except Exception:
-            logger.warning("Failed to collect provider session IDs for checkpoint", exc_info=True)
+        except Exception as exc:
+            logger.warning(
+                "Failed to collect provider session IDs for checkpoint (%s: %s)",
+                type(exc).__name__,
+                exc,
+            )
+            logger.debug("Session ID collection traceback", exc_info=True)
             copilot_session_ids = None
             copilot_session_cwds = None
 
@@ -3036,12 +3041,14 @@ class WorkflowEngine:
                 outcome = await validate_coro
         except asyncio.CancelledError:
             raise
-        except Exception:
+        except Exception as exc:
             logger.warning(
-                "Validator call for '%s' timed out or failed; treating as pass",
+                "Validator call for '%s' timed out or failed; treating as pass (%s: %s)",
                 agent.name,
-                exc_info=True,
+                type(exc).__name__,
+                exc,
             )
+            logger.debug("Validator call traceback for '%s'", agent.name, exc_info=True)
             outcome = ValidationOutcome(passed=True, errored=True)
         _v_elapsed = _time.time() - _v_start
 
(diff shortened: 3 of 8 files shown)


src/conductor/engine/checkpoint.py
```diff
@@ -510,8 +511,11 @@ def _delete_periodic_checkpoints(
         """
         try:
             candidates = CheckpointManager._periodic_checkpoints_for_run(workflow_path, run_id)
-        except Exception:
-            logger.warning("Failed to list checkpoints for %s", action, exc_info=True)
+        except Exception as exc:
+            logger.warning(
+                "Failed to list checkpoints for %s (%s: %s)", action, type(exc).__name__, exc
```
jrob5756: This message only carries `action` ("rotation"/"cleanup"), not which run or workflow failed. Since a failure here means checkpoints silently pile up over time, it'd help to include `run_id` or `workflow_path` at WARNING level so this is actionable without turning on DEBUG.

```suggestion
                "Failed to list checkpoints for %s (run_id=%s, workflow=%s) (%s: %s)",
                action,
                run_id,
                workflow_path,
                type(exc).__name__,
                exc,
```

src/conductor/engine/checkpoint.py
```diff
@@ -295,8 +295,9 @@ def save_checkpoint(
 
             return checkpoint_path
 
-        except Exception:
-            logger.warning("Failed to save checkpoint", exc_info=True)
+        except Exception as exc:
+            logger.warning("Failed to save checkpoint (%s: %s)", type(exc).__name__, exc)
+            logger.debug("Checkpoint save traceback", exc_info=True)
```
jrob5756, line 300: This site (and `_delete_periodic_checkpoints` below) already has a behavioral test for the failure path (`test_never_raises_on_failure`), but nothing pins the new logging contract the way `test_rerun_failure_keeps_original_no_double_count_and_emits_event` does for the validator re-run. Worth extending with the same `caplog` pattern so a future refactor can't quietly put `exc_info=True` back on the warning.

src/conductor/engine/validator.py
```diff
@@ -192,12 +192,14 @@ async def validate(
...
+        except Exception as exc:
             logger.warning(
-                "Validator call failed for agent '%s'; treating as pass",
+                "Validator call failed for agent '%s'; treating as pass (%s: %s)",
                 agent.name,
-                exc_info=True,
+                type(exc).__name__,
+                exc,
             )
+            logger.debug("Validator call traceback for agent '%s'", agent.name, exc_info=True)
```
jrob5756, line 202: This is the closest sibling to the workflow.py re-run path that already has a caplog regression test — same message-formatting pattern, same risk of a copy/paste slip going untested. A quick caplog assertion here (WARNING has no exc_info, DEBUG does) would close the gap cheaply.

jrob5756 (commented):
Solid, well-scoped fix!  The concise-warning/debug-traceback split is applied consistently, cancellation still propagates everywhere, and the hermes/_event_format downgrade actually closes an existing parity gap with claude_agent_sdk and _pydantic_ai rather than opening one. A couple of small suggestions below, nothing blocking at all.

src/conductor/engine/checkpoint.py
hertznsk: Done in 54d4a60 — the WARNING now names both `run_id` and `workflow_path`:

```
Failed to list checkpoints for rotation (run_id=run-42, workflow=/path/to/wf.yaml) (OSError: disk gone)
```

Also added `test_listing_failure_logs_run_context_and_debug_traceback` in `TestRotatePeriodicCheckpoints` asserting the WARNING carries the run context and that the traceback is DEBUG-only.

src/conductor/engine/checkpoint.py
hertznsk: Done in 54d4a60 — added two caplog tests in `tests/test_engine/test_checkpoint.py` mirroring the validator re-run pattern:

- `test_save_checkpoint_failure_logs_concise_warning_and_debug_traceback` (`TestSaveCheckpoint`) — pins `save_checkpoint`'s fail-open path: exactly one WARNING with `exc_info is None`, asserts the exception type (`FileNotFoundError`) appears in the message, and a DEBUG record with `exc_info` present.
- `test_listing_failure_logs_run_context_and_debug_traceback` (`TestRotatePeriodicCheckpoints`) — same contract for the rotation/cleanup listing failure, plus assertions that `run_id` and the workflow path appear in the WARNING text.

src/conductor/engine/validator.py
hertznsk: Done in 54d4a60 — added `test_provider_error_logs_concise_warning_and_debug_traceback` in `tests/test_engine/test_validator.py::TestValidatorValidate` with the same caplog shape as the workflow.py re-run test: `AsyncMock(side_effect=RuntimeError("boom"))`, asserts exactly one WARNING with `exc_info is None` containing "Validator call failed", "RuntimeError", "boom", and the agent name, plus a DEBUG record with `exc_info` present. A copy/paste slip reintroducing `exc_info=True` on the WARNING now fails this test.

jrob5756 (approved):
LGTM. Approved!

PR #372, opened 2026-08-05
feat: add output field constraints (enum, pattern, range, length, optional, nullable)

hertznsk:
## Summary

Workflow authors can now declare real constraints in their agents' `output:` blocks — allowed value lists (`enum`), regex patterns (`pattern`), numeric ranges (`minimum`/`maximum`), string length limits (`minLength`/`maxLength`), optional fields (`required: false`), and nullable fields (`nullable: true`). Every provider (Claude, Copilot, Hermes, Claude Agent SDK, ACA) enforces them the same way, and on Claude the model gets an automatic in-session chance to fix a constraint violation before the workflow fails.

## What changed

- **`OutputField` schema model** gains eight optional fields (`enum`, `pattern`, `minimum`, `maximum`, `minLength`, `maxLength`, `required`, `nullable`) with Pydantic validators rejecting type-inappropriate combinations (e.g. `pattern` on `type: number`, `None` entries in `enum` — use `nullable: true` instead).
- **Shared schema builders** (`_schema.py`) emit the new keywords in both flavours (JSON schema and prompt schema); `required` arrays list only required fields; `nullable` renders as `type: [<T>, "null"]`; no `default` keys ever appear in generated schemas.
- **Pydantic-ai converter** (Claude path) enforces constraints via `AfterValidator` (strict Python semantics identical to the final validator — no `Literal[...]` / Rust-regex surprises) while advertising them to the model via `json_schema_extra`. Constraint violations raise inside the pydantic-ai output-retry loop, giving the model a free in-session self-correction. Optional fields use `exclude_unset=True` so an omitted key is never materialized as `None`.
- **Strict validator** (`validate_output`) checks all eight constraints after the type check, recursion-aware (object properties and array items). Legacy error messages remain byte-identical when no new fields are used.
- **YAML-load gate**: root-level `required: false` is rejected at load time via a `WorkflowConfig` model validator (optional output fields are only allowed inside object properties).
- **Provider parity tests** prove one shared YAML schema produces the same keywords in Copilot/Hermes prompt schemas, the claude_agent_sdk `output_format` payload, pydantic constraints on the Claude dynamic model, and identical `ValidationError` from `validate_output`. The ACA wire boundary is covered too (constraint fields survive the round trip).
- **Docs**: `docs/workflow-syntax.md` gains a "Field constraints" subsection; new example `examples/output-constraints.yaml` exercising every field; CHANGELOG entry under Unreleased.

## Deliberately out of scope

- No `additionalProperties: false` / extra-key rejection (deferred to a follow-up).
- No default values for optional fields — an omitted optional key stays omitted.
- No new JSON Schema keywords beyond the agreed eight (no exclusive bounds, `multipleOf`, `format`, `const`, etc.).
- Workflows that do not use the new fields behave byte-identically (golden schema tests unchanged).

## Test plan

- `uv run pytest -m "not performance" -q` — full suite green
- `make lint` / `make typecheck` — clean
- `make validate-examples` — green
- New unit tests cover every rejection combination, parity across providers, and the Claude in-session retry path (via a `FunctionModel` returning a violating payload then a valid one)


src/conductor/providers/_pydantic_ai/converters.py
```diff
@@ -34,6 +36,156 @@ def _reject_bool(value: Any) -> Any:
...
+        elif isinstance(branch, dict):
+            value_branch = branch
+
+    if null_branch is None or value_branch is None:
+        return node
+
+    merged = dict(value_branch)
+    value_type = merged.pop("type", None)
+    if value_type is not None:
+        merged["type"] = [value_type, "null"]
```
jrob5756, line 84: `NumberType` is `int | float`, so a nullable number nests its union: `anyOf: [{anyOf: [integer, number]}, {null}]`. The value branch carries no `type` key, so `merged.pop("type", None)` returns `None`, line 83 is skipped, and the null branch is dropped without a word.

I ran a real `build_agent` for `OutputField(type="number", nullable=True, minimum=0, maximum=10)`. The schema actually sent to Anthropic:

```json
{"anyOf": [{"type":"integer"},{"type":"number"}], "ge":0.0, "le":10.0, "maximum":10.0, "minimum":0.0}
```

No `"null"` anywhere, while `build_json_schema_field` advertises `{"type": ["number","null"], ...}` to every other provider. Same YAML, opposite contract. Without a range pydantic flattens to a 3-branch union and the null survives, which is exactly why the tests pass.

```suggestion
    merged = dict(value_branch)
    value_type = merged.pop("type", None)
    if value_type is not None:
        merged["type"] = [value_type, "null"]
    elif "anyOf" in merged:
        inner = merged.pop("anyOf")
        types = [b["type"] for b in inner if isinstance(b, dict) and "type" in b]
        if len(types) == len(inner):
            merged["type"] = [*types, "null"]
        else:
            merged["anyOf"] = [*inner, null_branch]
    else:
        merged["anyOf"] = [value_branch, null_branch]
```

src/conductor/providers/_pydantic_ai/converters.py
```diff
@@ -49,6 +201,116 @@ def _to_pascal(snake: str) -> str:
...
+        depth=depth,
+        max_depth=max_depth,
+    )
+
+    if field.type in ("string", "number", "integer", "boolean"):
+        base_type = _wrap_scalar_field_type(field, base_type)
+        field_info = _build_field_info(field)
+        return Annotated[base_type, field_info]
+
+    return base_type
```
jrob5756: Nullability is applied only to scalars here, but `_build_pydantic_model` applies it unconditionally. A nullable object or array item therefore diverges:

- `validate_output({"a": [None]}, ...)` accepts it (`output.py:138`)
- `output_schema_to_pydantic_model(...).model_validate({"a": [None]})` raises

I confirmed both. The shared builder advertises `items: {"type": ["object","null"]}`, so this workflow runs on copilot, hermes, aca and claude_agent_sdk and hard-fails on claude.

`"integer"` is also unreachable: `OutputField.type` is `Literal["string","number","boolean","array","object"]`. Dropping it here is behaviour preserving.

```suggestion
    if field.type in ("string", "number", "boolean"):
        base_type = _wrap_scalar_field_type(field, base_type)
        field_info = _build_field_info(field)
        return Annotated[base_type, field_info]

    if field.nullable:
        base_type = base_type | None

    return base_type
```

src/conductor/executor/output.py
```diff
@@ -44,6 +47,55 @@ def validate_output(
...
+                    f"Output field '{field_name}' must be one of {field_def.enum!r}, got {value!r}",
+                    suggestion=f"Ensure '{field_name}' is one of {field_def.enum!r}",
+                )
+        elif value not in field_def.enum:
+            raise ValidationError(
+                f"Output field '{field_name}' must be one of {field_def.enum!r}, got {value!r}",
+                suggestion=f"Ensure '{field_name}' is one of {field_def.enum!r}",
+            )
+
+    if field_def.pattern is not None and re.search(field_def.pattern, value) is None:
```
jrob5756: `pattern` runs before the `minLength`/`maxLength` block below, so a bounded field is not actually bounded when the regex sees it. Model output is untrusted input, and Python's `re` has no timeout.

`OutputField(type="string", pattern="^(a+)+$", maxLength=40)` loads without complaint. A 32 character input then never returns. I let it run 40 seconds before killing it.

The part that makes this more than a slow function: `re` holds the GIL while it backtracks. My watchdog thread could not even print its timeout message. `validate_output` is called synchronously from `executor/agent.py`, so this stalls the event loop, every agent in a `parallel:` or `for_each:` group, and the Esc-to-pause listener. The run does not fail, it just stops.

Worth noting that reordering alone is not enough. 32 characters is already under `maxLength: 40`. Options that do work: reject nested quantifiers at load time, or run `search` against a deadline and raise `ValidationError`.

src/conductor/providers/_schema.py
```diff
@@ -65,14 +65,35 @@ def build_json_schema_field(
...
     schema: dict[str, Any] = {"type": field.type}
 
+    if field.nullable:
+        schema["type"] = [field.type, "null"]
+
     if field.description:
         schema["description"] = field.description
 
+    if field.enum is not None:
+        schema["enum"] = field.enum
```
jrob5756: `type` and `enum` are conjunctive in JSON Schema, so `{"type": ["string","null"], "enum": ["a","b"]}` forbids null regardless of the type list. Meanwhile `output.py:115` returns early for `None` and the pydantic model accepts it too, so every enforcement layer permits a value every published schema rejects.

This combination is not hypothetical. `schema.py:161` refuses `null` inside `enum` and tells the author to use `nullable: true` instead, so the validator points people straight at it.

```suggestion
    if field.enum is not None:
        schema["enum"] = [*field.enum, None] if field.nullable else field.enum
```

Same change is needed in `build_prompt_schema_field` and in `_field_json_schema_extra`.

src/conductor/providers/_pydantic_ai/converters.py
```diff
@@ -34,6 +36,156 @@ def _reject_bool(value: Any) -> Any:
...
+    flattens nullable unions into ``{type: [T, "null"]}``.
+    """
+
+    model_config = ConfigDict(extra="allow")
+
+    @classmethod
+    def model_json_schema(cls, *args: Any, **kwargs: Any) -> dict[str, Any]:
+        schema = super().model_json_schema(*args, **kwargs)
+        _sanitize_json_schema(schema)
+        return schema
```
jrob5756: This override is never called. I wrapped it in a spy and ran a real `build_agent` with a rich schema: zero calls. Deleting it leaves the final tool schema byte identical, and so does replacing the whole class with `BaseModel` plus `ConfigDict(extra="allow")`.

`_build_output_type` always wraps the model in `ToolOutput`, and pydantic-ai builds tool schemas through its own `GenerateJsonSchema` pipeline, which does not route through `Model.model_json_schema`. The real sanitisation is the pass in `agent_builder.py`.

Two knock-on effects worth deciding on together. The docstring above describes stripping `default`, `$ref`, `$defs` and flattening `anyOf`, none of which happens here. And because `agent_builder.py` is the only production caller, the unresolvable-`$ref` branch below can never be reached by the second pass its own docstring promises.

src/conductor/providers/_pydantic_ai/structured_output.py
```diff
@@ -55,7 +55,7 @@ def extract_content(
         return _wrap_text_output(output)
 
     if isinstance(output, BaseModel):
-        content = output.model_dump()
+        content = output.model_dump(exclude_unset=True)
```
jrob5756, line 58: `exclude_unset=True` combined with `extra="allow"` on the dynamic model turns a misspelled optional key into silent data loss. Reproduced with `verdict` (required, enum) plus `summary` (`required: false`):

```
model returns   {"verdict": "pass", "summry": "typo'd value"}
model_dump()    {"verdict": "pass", "summry": "typo'd value"}
validate_output PASSED
```

`validate_output` iterates `schema.items()` and never looks at `content.keys()`, so extras are structurally invisible to it. That was safe while every field was required, because the missing key produced a pydantic `missing` error and drove the issue #343 recovery loop. Optional fields remove that backstop, and the failure resurfaces two agents later as `TemplateError: Undefined variable 'summary'`, pointing at the consumer rather than the producer.

A warning naming the unknown keys would keep the "extras are tolerated" contract while making the near miss visible. Emitting it as an event would be better still, since a typo is exactly what a retry could fix.

tests/test_providers/test_output_constraints_parity.py
```diff
@@ -0,0 +1,307 @@
...
+
+def _stub_handler(agent: AgentDef, prompt: str, context: dict[str, Any]) -> dict[str, Any]:
+    """Return a minimal dict for CopilotProvider mock-handler construction."""
+    return {}
+
+
+# Shared schema exercising every constraint keyword supported by OutputField.
+# It intentionally mixes nullable, optional, and required fields so the tests
+# can verify that each keyword survives its provider-specific transformation.
+SHARED_OUTPUT_SCHEMA = {
```
jrob5756, line 57: This fixture is why the two nullable defects above survived 4931 tests. Every field is a flat scalar, so no array or object shape is exercised at all, and the combinations that break are never co-located:

- `code` is nullable but is a string; `score` has the range but is not nullable, so a nullable number with a range never appears
- `enum` and `nullable` are never on the same field
- `required: false` never appears on an object property, only at the root

I mutation tested it. Deleting every `min_length`/`max_length`/`ge`/`le` from `_build_field_info` leaves this file green. Deleting the `nullable` handling from `_wrap_scalar_field_type` also leaves it green. A parity file that passes while one provider drops half the feature is not yet doing its job.

Cases that would have caught both on the first run:

```python
"ratio": OutputField(type="number", minimum=0, maximum=10, nullable=True),
"nullable_enum": OutputField(type="string", enum=["A", "B"], nullable=True),
"rows": OutputField(type="array", items=OutputField(
    type="object", nullable=True, properties={"name": OutputField(type="string")})),
"nested": OutputField(type="object", properties={
    "opt": OutputField(type="string", required=False)}),
```

The bigger fix is the assertion style. Right now copilot, hermes and claude_agent_sdk each re-assert the same literals by hand, claude is checked on accept/reject, and aca on wire keys, so nothing ever compares one provider's schema to another or to the shared builder. `test_output_schema.py` already does the right thing with full-dict equality. Parameterising a single normalising comparison against `build_json_schema_field` would replace all three literal blocks and catch this class of bug by construction.

src/conductor/config/schema.py
```diff
@@ -99,15 +100,99 @@ class OutputField(BaseModel):
     properties: dict[str, OutputField] | None = None
     """For object types, the schema of object properties."""
 
+    enum: list[Any] | None = None
```
jrob5756, line 112: `OutputField` has no `model_config`, so unknown keys are accepted and ignored. `OutputField.model_validate({"type": "string", "minlength": 5})` loads happily and the field ships unconstrained, with `conductor validate` reporting success.

That was survivable with four well-known keys. With twelve, two of them camelCase in a schema whose every other YAML key is snake_case, a typo now quietly downgrades a declared contract to no contract. `RouteDef`, `ParallelGroup` and the other models in this file already set `extra="forbid"`.

```suggestion
    model_config = ConfigDict(extra="forbid")

    enum: list[Any] | None = None
```

Worth flagging that `aca.py` relies on the current tolerance for forward compatibility, so this needs checking against the runner before it lands.

src/conductor/config/schema.py
```diff
@@ -99,15 +100,99 @@ class OutputField(BaseModel):
...
+    """Allowed values for scalar types."""
+
+    pattern: str | None = None
+    """Regular expression pattern for string types."""
+
+    minimum: float | None = None
```
jrob5756: `minimum: 0` in YAML becomes `0.0`, which then shows up in generated schemas and in error text as `is below minimum 0.0`. `int | float` round-trips integral bounds unchanged.

```suggestion
    minimum: int | float | None = None
```

src/conductor/config/schema.py
```diff
@@ -99,15 +100,99 @@ class OutputField(BaseModel):
...
+            if self.type in ("array", "object"):
+                raise ValueError("enum can only be set for scalar types")
+
+            if len(self.enum) == 0:
+                raise ValueError("enum must contain at least one value")
+
+            if any(value is None for value in self.enum):
+                raise ValueError(
+                    "enum cannot contain null; use nullable: true to allow null values"
+                )
```
jrob5756, line 184: This message is good, but it currently routes authors into a combination that does not work. `enum: ["a","b"], nullable: true` produces a schema where null can never validate, while all three enforcement paths accept it. Worth resolving alongside the `_schema.py` change so the advice and the emitted schema agree.

src/conductor/config/schema.py
```diff
@@ -99,15 +100,99 @@ class OutputField(BaseModel):
...
+            raise ValueError("minLength cannot be greater than maxLength")
+
+        # Number range validation.
+        if self.minimum is not None and self.maximum is not None and self.minimum > self.maximum:
+            raise ValueError("minimum cannot be greater than maximum")
+
+        # Pattern compilation.
+        if self.pattern is not None:
+            try:
+                re.compile(self.pattern)
```
jrob5756: The compiled pattern is discarded, so every `re.search` at validation time re-looks-up the string in the `re` module cache. That cache holds 512 entries and is cleared wholesale on overflow, so a workflow with many patterns will thrash.

Keeping the compiled object also gives the ReDoS fix somewhere natural to live.

src/conductor/executor/output.py
```diff
@@ -44,6 +47,55 @@ def validate_output(
...
+            if not (field_def.type == "boolean" and value in field_def.enum):
+                raise ValidationError(
+                    f"Output field '{field_name}' must be one of {field_def.enum!r}, got {value!r}",
+                    suggestion=f"Ensure '{field_name}' is one of {field_def.enum!r}",
+                )
+        elif value not in field_def.enum:
+            raise ValidationError(
+                f"Output field '{field_name}' must be one of {field_def.enum!r}, got {value!r}",
+                suggestion=f"Ensure '{field_name}' is one of {field_def.enum!r}",
+            )
```
jrob5756: `_check_constraints` runs only after `check_type` passed, and `check_type` admits a `bool` solely when `type == "boolean"`. The first conjunct on line 58 is therefore always true, making this branch an exact duplicate of the `elif` below, same message and all.

```suggestion
        if value not in field_def.enum:
            raise ValidationError(
                f"Output field '{field_name}' must be one of {field_def.enum!r}, got {value!r}",
                suggestion=f"Ensure '{field_name}' is one of {field_def.enum!r}",
            )
```

`_make_enum_validator` in `converters.py` carries the same mirrored branch. Worth collapsing both together so the two enforcement paths stay in step.

src/conductor/executor/output.py
```diff
@@ -68,11 +128,15 @@ def _validate_field(field_name: str, value: Any, field_def: OutputField) -> None
...
 
+    _check_constraints(field_name, value, field_def)
+
     if field_def.type == "object" and field_def.properties and isinstance(value, dict):
         validate_output(value, field_def.properties)
 
     if field_def.type == "array" and field_def.items and isinstance(value, list):
         for i, item in enumerate(value):
+            if item is None and field_def.items.nullable:
+                continue
```
jrob5756, line 183: The recursive call on line 147 passes the parent array's name through, so a constraint failure on element 37 reports `Output field 'tags' must be one of [...]` with no index, while the type error four lines above correctly says `Array item 37 in 'tags'`. Two different conventions inside one loop.

None of the six `_check_constraints` messages use `_describe_value` either, so the offending value never appears. That helper was added precisely because naming only the types makes log-only diagnosis impractical, and the same reasoning applies here.

Passing `f"array item {i} in '{field_name}'"` into the recursive call would line both up.

src/conductor/providers/_pydantic_ai/converters.py
```diff
@@ -49,6 +201,116 @@ def _to_pascal(snake: str) -> str:
...
+    """
+    extra: dict[str, Any] = {}
+    if field.enum is not None:
+        extra["enum"] = field.enum
+    if field.pattern is not None:
+        extra["pattern"] = field.pattern
+    if field.minLength is not None:
+        extra["minLength"] = field.minLength
+    if field.maxLength is not None:
+        extra["maxLength"] = field.maxLength
```
jrob5756: These two are already emitted by pydantic from the `min_length`/`max_length` kwargs in `_build_field_info`. I ablated `_field_json_schema_extra` to `lambda field: None` and diffed: the length keys were identical across plain, nullable, array-item and nested-object shapes, contributing nothing.

`enum`, `pattern`, `minimum` and `maximum` are load bearing and must stay, so this is a small trim rather than a rewrite.

Related, and more visible to users: `Field(ge=..., le=...)` cannot attach to the `int | float` union, so pydantic emits raw `ge`/`le` alongside the correct `minimum`/`maximum`. They are pydantic-isms rather than JSON Schema keywords and they reach the wire.

src/conductor/providers/_pydantic_ai/converters.py
```diff
@@ -34,6 +36,156 @@ def _reject_bool(value: Any) -> Any:
...
+
+    - A boolean value only passes when the field type is ``boolean`` and the
+      value is present in the enum.
+    - For all other values, plain Python membership is used (so ``1.0``
+      satisfies a number enum ``[1]``).
+    """
+
+    def _validate(value: Any) -> Any:
+        if value is None:
+            return value
```
jrob5756: This guard cannot fire. `_wrap_scalar_field_type` attaches the `AfterValidator` to `base_type` before the `| None` union is added, so in `Annotated[str, AfterValidator(...)] | None` a `None` matches the `NoneType` member and the validator is never entered. I counted 396 validations across enum and pattern, string, number and boolean, nullable and not: `value is None` was reached zero times.

The same applies to the `None` guard in `_make_pattern_validator`.

src/conductor/providers/aca.py
```diff
@@ -694,7 +694,10 @@ def _build_request(
         reasoning_effort = resolve_reasoning_effort(agent, self._default_reasoning_effort)
         working_dir = agent.sandbox.working_dir if agent.sandbox is not None else None
         output_schema = (
-            {name: field.model_dump(exclude_none=True) for name, field in agent.output.items()}
+            {
+                name: field.model_dump(mode="json", exclude_none=True, exclude_defaults=True)
+                for name, field in agent.output.items()
+            }
```
jrob5756, line 704: `exclude_defaults=True` gives the wire format the meaning "absent equals whatever the receiver's default is". Both sides agree today and I checked the nested round trip is lossless, so nothing is broken right now.

The part worth a decision is version skew. The in-container runner reconstructs through `OutputField.model_validate`, which has no `extra="forbid"`, so an older runner silently drops `enum`, `pattern`, `minimum` and the rest. The inner Copilot session then builds a prompt schema with no constraints, the model is never told about them, the session closes, and the host rejects the output afterwards with no chance at the in-session recovery the #343 contract is built around. The only signal is `_warn_on_version_skew`, whose text never mentions output.

An `output_constraints` capability on `/health`, with a clear `ProviderError` when an agent declares constraints the runner cannot honour, would turn a quiet degradation into something actionable.

jrob5756 (commented):
Read through the constraint plumbing end to end and ran the new code against this branch to check the behaviour below.

The direction here is great. Extending the two shared builders in `_schema.py` and keeping a single `validate_output` is what makes cross-provider parity reachable at all, and routing `pattern` through Python `re` on every provider rather than letting pydantic's Rust engine diverge is a good call. Load-time validation is thorough, and `test_output_schema.py` asserting full-dict equality is the right instinct.

Two things stop it working as advertised, both on the Claude path, both invisible to the current tests. A nullable number that also carries `minimum`/`maximum` loses its null branch entirely. And a nullable object or array item is accepted by `validate_output` but rejected by the dynamic model, so the same YAML runs on four providers and fails on the fifth. I verified both against a real `build_agent` tool schema rather than by reading.

Separately, `pattern` is evaluated before `minLength`/`maxLength`, which puts a catastrophic regex within reach of model output. `^(a+)+$` under `maxLength: 40` never returns on a 32 character input, and because `re` holds the GIL while backtracking it takes the event loop and every neighbouring agent with it.

If you only change one thing first, make it the parity fixture. It is what let the other two through: deleting all range and length enforcement from the Claude path leaves `test_output_constraints_parity.py` green, and so does deleting nullable handling outright.

Details inline. Several are small cleanups rather than problems, and I have marked which is which. Happy to talk any of them through.

Thanks again for your contributions!

src/conductor/providers/_pydantic_ai/converters.py
hertznsk: Fixed. `_convert_anyof_nullable` now flattens the nested `int | float` union — when the value branch has no `type` but an `anyOf` whose branches all declare one, it emits `type: ["integer", "number", "null"]`; any unrecognized shape returns the node unchanged, so the null branch is never silently dropped. The regression test goes through a real `build_agent` and inspects the attached tool schema: a nullable ranged number now advertises `{"type": ["integer", "number", "null"], "minimum": 0, "maximum": 10}`.

src/conductor/providers/_pydantic_ai/converters.py
hertznsk: Fixed. `_build_array_item_type` now unions non-scalar item types with `None` when `nullable: true`, so `{"rows": [None]}` validates identically on the pydantic path and in `validate_output` (pinned by the new payload-matrix parity test). The unreachable `"integer"` entries are dropped from the new tuples; the pre-existing `IntegerType` branch in `_map_output_field_type` predates this PR and is left alone.

src/conductor/executor/output.py
hertznsk: Fixed on both axes. Length checks now run before the pattern check, and matching moved from stdlib `re` to the `regex` engine under a 1s wall-clock deadline (`PATTERN_MATCH_TIMEOUT_SECONDS`). A timeout raises `ValidationError` here, and a `ValueError` inside the pydantic AfterValidator, so it drives the in-session output retry instead of hanging the run. Verified empirically: `^(a|aa)+$` on a 61-char input raises in ~1s; your `^(a+)+$` example is optimized by the regex engine and returns instantly. The compiled pattern is cached on `OutputField.compiled_pattern`, so the deadline has a single home.

src/conductor/providers/_schema.py
hertznsk: Fixed in all three emission points — both `_schema.py` flavours and `_field_json_schema_extra`. A nullable enum now emits `[*enum, None]`, so the advertised schema no longer forbids the null that every enforcement layer accepts. The author-facing rule is unchanged: YAML `enum` still cannot contain null (rejected at load), `nullable: true` remains the sole mechanism.

src/conductor/providers/_pydantic_ai/converters.py
hertznsk: Confirmed with a spy through a real `build_agent` — zero calls, and deleting the override leaves the attached tool schema byte-identical. Removed it, renamed the class to `_OutputBaseModel` (its remaining job is `extra="allow"`), and corrected both docstrings, including the phantom final-inlining-pass promise on `_sanitize_schema_node` (the `agent_builder.py` pass is the only one).

src/conductor/providers/_pydantic_ai/structured_output.py
hertznsk: Went with the warning: `validate_output` now logs the undeclared keys by name. It is the single choke point, so the warning covers every provider plus `set`/`script` outputs, not just the pydantic path. I kept the event-based variant out of this PR since it touches the event vocabulary and the dashboard — happy to file a follow-up if you think the retry angle is worth it.

tests/test_providers/test_output_constraints_parity.py
hertznsk: Done, and done first per your priority. The fixture now co-locates the breaking combinations: `ratio` (nullable + range), `nullable_enum` (enum + nullable), `rows` (nullable object items), `nested` (optional property inside an object). On top of that: a 16-payload matrix asserting the Claude dynamic model and `validate_output` agree on accept/reject for every payload, and a tool-schema hygiene test through the real `build_agent` seam asserting no `ge`/`le`/`default` keys anywhere. Deleting range enforcement or nullable handling now fails immediately. I did not do the full normalize-against-`build_json_schema_field` parameterization — the agreement matrix catches the same bug class more directly — but I am open to it if you prefer that shape.

One correction worth surfacing: the previous `claude_agent_sdk` assertion (`required == [category, code, score]`) never actually ran, because the test skips without the SDK installed. `_build_output_format` marks all declared fields required via `list(output.keys())` — correct in practice since the load-time gate forbids root-level `required: false` — so the updated test asserts that and the docstring now says so.

src/conductor/config/schema.py
hertznsk: Done. The runner reconstructs through the same `OutputField` class and the wire only carries dumped known fields, so this is wire-safe today; the ACA and config suites are green. Side benefit for the version-skew concern in your other comment: once a runner carries this change, future unknown fields fail loudly at reconstruction instead of being silently dropped.

src/conductor/config/schema.py
hertznsk: Done — `int | float | None` for both bounds. `minimum: 0` now round-trips as `0` in generated schemas and error text instead of `0.0` (pinned by a round-trip test).

src/conductor/config/schema.py
hertznsk: Resolved together with the `_schema.py` change: with null appended to the emitted enum when `nullable: true`, the advice in this message and the generated schema now agree.

src/conductor/config/schema.py
hertznsk: Done — `OutputField.compiled_pattern` is a cached_property (excluded from `model_dump`, verified by test), shared by both enforcement paths, and it is where the 1s match deadline lives.

src/conductor/executor/output.py
hertznsk: Collapsed in both places — `_check_constraints` and the mirrored `_make_enum_validator` are now a single plain-membership check each. The pinned semantics are unchanged and tested on both paths: a number `enum: [1]` rejects `True` (via the type check / `_reject_bool` before the enum check) and accepts `1.0`.

src/conductor/executor/output.py
hertznsk: Done. The recursion now passes `array item {i} in '{field}'`, so a constraint failure on element 37 names the index, lining up with the existing type-error convention (which is unchanged). The pattern/length/range messages also include the offending value via `_describe_value` (containers reduced to shape, scalars repr-truncated at 200 chars).

src/conductor/providers/_pydantic_ai/converters.py
hertznsk: Both done. `minLength`/`maxLength` are removed from `json_schema_extra` (pydantic already emits them from the Field kwargs — your ablation finding holds), and the raw `ge`/`le` keys are stripped in the sanitize pass while the standard `minimum`/`maximum` remain; a tool-schema hygiene test asserts recursively that neither `ge`/`le` nor `default` reaches the attached schema.

src/conductor/providers/_pydantic_ai/converters.py
hertznsk: Removed from both `_make_enum_validator` and `_make_pattern_validator` — the NoneType union member short-circuits before the AfterValidator runs, exactly as you measured.

src/conductor/providers/aca.py
hertznsk: Agreed this deserves a proper design, and I would rather not grow this PR further — an `output_constraints` capability on `/health` with a clear `ProviderError` is a protocol change on both host and runner. Partial mitigation landed here: `extra="forbid"` on `OutputField` makes a newer-host/older-runner skew fail loudly at reconstruction instead of silently degrading. Want me to open a follow-up issue for the capability handshake?

hertznsk:
Thank you for the thorough review — the mutation testing of the parity fixture in particular caught real gaps.

Both blockers are fixed and verified through a real `build_agent` tool schema: the nullable ranged number now advertises `type: ["integer", "number", "null"]` with `minimum`/`maximum`, and nullable object/array items validate identically on the pydantic path and in `validate_output`.

The ReDoS vector is closed by moving pattern matching to the `regex` engine with a 1s wall-clock deadline, plus length-before-pattern ordering; a pathological pattern now raises a validation error that drives the normal retry loop instead of hanging the run.

The parity fixture is rebuilt around your four cases, plus a 16-payload accept/reject agreement matrix between the dynamic model and `validate_output`, plus a tool-schema hygiene test pinning the absence of `ge`/`le`/`default` keys. The deletions you mutation-tested now fail loudly.

Two items intentionally not in this PR: the event-based retry for undeclared-key typos (a warning naming the keys landed instead), and the ACA `/health` capability handshake (partial mitigation here: `extra="forbid"` makes the skew loud at the runner). Happy to file follow-up issues for both.

Everything else is addressed inline in the threads. Full suite green: 4969 passed, lint/typecheck/examples clean.

hertznsk:
Holding this PR for now — while dogfooding this branch on my own workflows I ran into a problem with the undeclared-fields warning, and I want to think it through before merge.

1. **Script steps warn on Conductor's own baseline keys.** `_validate_script_output_schema` validates the merged `output_content` — the `stdout`/`stderr`/`exit_code` baseline plus the parsed-JSON overlay — so every `type: script` step with a declared `output:` schema now logs `Output contains undeclared fields ... ['stdout', 'stderr', 'exit_code']`. These keys are injected by the executor itself, and schema authors cannot declare them away: root-level `required: false` is rejected by the load gate added in this PR.
2. **The warning also punishes a legitimate reuse pattern.** A shared script may emit a superset JSON payload where one workflow declares only the fields it consumes and another workflow declares a different subset. Fields that are *intentionally* undeclared in a given workflow now produce typo-looking noise, which trains users to ignore the warning entirely.

The soft alternative to extra-key rejection clearly needs more thought. Options I see: scope the warning to agent/provider steps only, exclude the script baseline keys, or drop the warning here and leave extra-key handling to the follow-up where it was planned. @jrob5756 curious about your take before I push anything.

hertznsk:
Pushed a resolution for the warning scoping problem (9c76781), taking option A from my earlier comment.

**What changed:** `validate_output` now takes `warn_undeclared_keys: bool = False` (keyword-only). The five provider-backed call sites pass `True` — the executor backstop, both parse-recovery loops (copilot, hermes), and both paths in `_pydantic_ai/structured_output` — while `set`/`script` step validation leaves it off. The flag threads through nested-object and array-item recursion; the warning text itself is unchanged.

**Why this shape:** the silent-data-loss repro above was specific to a non-deterministic producer — a model misspelling an optional key under `extra="allow"` + `exclude_unset=True`. That path still warns and names the keys, right after `model_dump`. Script steps differ in kind: the validated dict always carries Conductor's own `stdout`/`stderr`/`exit_code` baseline, and a shared script can legitimately emit a superset payload — both correct-by-construction output, so warning there was 100% noise and would have trained users to ignore the warning everywhere. The new load-time gate also means authors cannot declare the baseline keys away, which made that noise unfixable from YAML.

Tests pin all three behaviors: a provider-path typo warns; script baseline keys and superset payloads stay silent; set-step extras stay silent. Full suite green: 4974 passed, lint/typecheck clean.

This also sharpens the follow-up picture for real extra-key handling: given baseline keys and the superset-reuse pattern, rejection cannot be a default — it would have to be opt-in closed-object semantics (`additionalProperties: false`-style), enforced inside each provider's retry path rather than at the fatal backstops. I will fold that framing into the follow-up issue, and I am still happy to file the event-based retry variant alongside it if you think it is worth tracking.

@jrob5756 — ready for another pass when you have a chance.

hertznsk:
Heads up @jrob5756 — I noticed the CI test job failed on the latest run. That was my mistake.

The new `TestAcaWireBoundary` tests call `AcaRuntimeProvider._build_request`, which resolves the inner Copilot credential from the host environment (`COPILOT_PROVIDER_BASE_URL` → GitHub token env vars → `gh auth token`). They passed on my machine because my `gh` is signed in, but CI has no credential available, so the resolver raised `ProviderError`.

Fixed in 1b1dd08: an autouse fixture now pins `COPILOT_PROVIDER_BASE_URL`, so the resolver takes the BYOK branch and the tests no longer depend on machine auth state. Verified locally with all credential env vars unset; the full parity and ACA suites pass.

jrob5756 (approved):
LGTM. Approved!

PR #387, opened 2026-08-08
fix(cli): stop agent text being parsed as Rich markup

### LINKED ISSUE #382: MarkupError kills the run when agent text contains a bracketed closing-tag form (e.g. `[/nestedType]`) — reachable without --verbose

franklixuefei:
Fixes #382.

## The bug

Rich parses `[...]` as markup. Agent output is arbitrary text and regularly contains bracketed tokens — code, globs, regexes, ARM/URI patterns — so a closing-tag form in ordinary technical prose raises `MarkupError` and kills the run.

The trigger that found this was a plan document describing an Azure RBAC permission format:

```
{provider}/{type}[/{nestedType}...]/read
```

Nothing about that is malformed; it is what the format looks like.

```
rich.errors.MarkupError: closing tag '[/{nestedType}...]' at position 145705 doesn't match any open tag
```

It killed a long-running workflow mid-execution, then killed **both resume attempts at the identical byte offset**, because the offending text is checkpointed and replayed verbatim. Until I hand-edited the checkpoint the run was permanently unresumable.

## It is not verbose-only

Two sinks render agent-supplied content, and only one of them is gated on verbosity:

```python
should_console = is_verbose() and is_full()
should_file = _file_console is not None
...
if _file_console is not None:
    _file_console.print(Panel(content, ...))   # <-- no verbosity condition
```

`init_file_logging` is called from a bare `if log_file is not None:` in both `run` and `resume`. So **`--log-file` alone is sufficient to arm the crash**, including under `--silent`. Since `--log-file auto` is the natural way to keep a record of a long background run, this is reachable from the configuration you would most want for exactly the runs where losing the work hurts most.

The content is the **rendered prompt**, which embeds user-supplied plan/workflow text — so the crash is driven by user data, not by anything conductor controls.

## The change

**File console → `markup=False`.** It is already `no_color=True` plain text, so it has no use for markup interpretation.

I checked this is safe rather than assuming it: all 20 `_file_console.print(...)` call sites pass either a pre-built `rich.text.Text` (13 of them) or a plain string. `Text` is not markup-parsed, so the flag cannot change it. Every markup-bearing print in the module targets `_verbose_console` instead.

**Console path → `escape()` on the content.** That console deliberately keeps `markup=True` for conductor's own `[cyan]` title, so the fix there is to escape the untrusted part rather than de-feature the console. `title` is conductor-controlled and keeps its styling.

Both were verified against a standalone repro to raise nothing **and** preserve the literal text byte-for-byte.

## Tests

`tests/test_cli/test_markup_safety.py`, 15 tests. Two things I want to call out because they are what makes them worth having:

**The samples are pinned as genuine triggers.** `test_sample_raises_when_markup_is_enabled` asserts each fixture really does raise with markup on. Without that, a sample that happens not to trigger would make its cases pass vacuously — the same trap as the em-dash in #342, which turned out to be encodable in cp1252 and would have been a dud.

**The production consoles are asserted, not just hand-rolled ones.** The behavioural tests construct their own `Console` and would keep passing if someone dropped `markup=False` from `init_file_logging`. So there are assertions against the real `_file_console` and a test that drives `verbose_log_section` end-to-end through both sinks.

Verified by reverting both fixes: **3 of 15 fail**, and all 15 pass on the fixed tree.

Minimal triggers covered, all confirmed raising without the fix:

| input | result |
|---|---|
| `{provider}/{type}[/{nestedType}...]/read` | the real-world case |
| `[/nestedType]` | `doesn't match any open tag` |
| `[/]` | `has nothing to close` |
| `[/bold]` | `doesn't match any open tag` |
| `text [/foo] more` | raises at position 5 |

## Validation

`tests/test_cli` + `tests/test_web` matches the pre-existing baseline **exactly** — 48 failures before and after, verified by stashing the change and re-running. (My sandbox has no network access to PyPI so `pytest-asyncio` is unavailable and those async tests error identically either way; I diffed the failure sets rather than trusting the count.) `ruff check src tests` and `ruff format --check src tests` are clean.

## Not in this PR

- **A wider audit.** This is the instance that bit me, but any `Panel(...)` / `print(...)` fed agent- or user-derived text has the same shape. I fixed the two sinks I could demonstrate; a sweep is worth doing separately rather than bundled here.
- **Escaping at the source** (e.g. in the executor before the text reaches any renderer). That would be more thorough but also more invasive, and it would change what is stored in checkpoints — which is the thing that made this unresumable. Worth discussing if you would prefer it.

## Workaround, for anyone hitting this before it lands

The offending text is in the checkpoint, so a plain `resume` re-crashes at the same offset. Rewrite the bracket form in the checkpoint JSON (`[/x]` → `(/x)`) and then resume.

## Environment

```
conductor  v0.1.26 (main @ 4a4bd57)
rich       14.3.1
Python     3.14.5
OS         Windows 11
```

Not Windows-specific — nothing in the parse path is platform-dependent — but I have only reproduced it there.

## Related

Same run, different boundaries: #342 / #381 (`UnicodeEncodeError` at the byte-write boundary) and #344 / #383 (`conductor stop` orphaning runs). Three separate failures, three separate fixes; this one is a *parse* failure and neither of the others prevents it.


src/conductor/cli/run.py
```diff
@@ -124,7 +125,16 @@ def init_file_logging(log_path: Path) -> None:
...
     log_path.parent.mkdir(parents=True, exist_ok=True)
     _file_handle = open(log_path, "w", encoding="utf-8")  # noqa: SIM115
-    _file_console = Console(file=_file_handle, no_color=True, highlight=False, width=200)
+    # markup=False because this console renders agent-supplied text (prompts,
+    # tool output, model responses) verbatim. Rich would otherwise parse a
+    # bracketed token such as ``[/nestedType]`` -- ordinary technical prose --
+    # as a closing tag and raise MarkupError, killing the run (see #382). The
+    # log is already ``no_color=True`` plain text, so it has no use for markup;
+    # every styled write in this module goes to ``_verbose_console`` or passes a
+    # pre-built ``rich.text.Text``, neither of which is affected by this flag.
```
jrob5756: The last two lines of this comment are not true of the file as it stands, and they are the argument for why the change is safe.

`_maybe_print_experimental_banner` (~L906-918) builds a `Panel` from a markup-bearing plain string and prints that same object to `_verbose_console` *and* `_file_console`. I A/B'd it against `main`: the log file now contains the literal tags rather than styled text.

```
main:  Experimental provider in use: claude-agent-sdk (pin) maintained by @someone
here:  Experimental provider in use: [bold]claude-agent-sdk[/bold] ([dim]pin[/dim]) maintained by [dim]@someone[/dim]
```

Nothing catches it: `test_experimental_banner.py` swaps `_verbose_console` for a `MagicMock` and never calls `init_file_logging`, so it cannot observe the file sink.

Two separate things to fix. The banner should build its body with `Text.from_markup("\n".join(body_lines))`, which I verified renders identically on both consoles. And the comment should stop claiming an invariant that nothing enforces.

Also worth dropping the `no_color=True` clause. `no_color` suppresses colour only; bold and italic still emit ANSI:

```
force_terminal=True, no_color=True : '\x1b[1mB\x1b[0m \x1b[3mI\x1b[0m R'
```

Markup looked harmless in the log because the console writes to a file handle and `is_terminal` is False, not because of `no_color`. Anyone who later sets `force_terminal=True` would be misled.

```suggestion
    # markup=False because this console renders agent-supplied text (prompts,
    # tool output, model responses) verbatim. Rich would otherwise parse a
    # bracketed token such as ``[/nestedType]`` -- ordinary technical prose --
    # as a closing tag and raise MarkupError, killing the run (see #382).
    # Consequence: a markup-bearing renderable now prints its tags literally
    # here, so style file output with ``rich.text.Text`` rather than markup.
```

src/conductor/cli/run.py
```diff
@@ -376,7 +386,14 @@ def verbose_log_section(title: str, content: str) -> None:
         return
 
     if should_console:
```
jrob5756, line 348: Three more sinks in this same file still die on the identical input, and two of them are reachable without any flags at all.

```
run.py:202  _verbose_console.print(f"[{style}]{message}[/{style}]")
run.py:524  _verbose_console.print(error_msg, style="red dim")
run.py:763  same pattern
```

`style=` does not turn markup parsing off, which I checked directly. At 524 and 763 `error_msg` carries a provider exception message, so a workflow that is already failing has its real error replaced by a `MarkupError`.

I reproduced two of these as real `conductor run` failures on this branch with no flags set:

- a `type: script` step whose stderr contains the trigger reaches `executor/script.py:212` and exits 1 with `closing tag '[/nestedType]' at position 103`
- a `type: wait` step whose rendered `reason` contains it reaches `executor/wait.py:143` and exits 1

Related: `verbose_mode` and `full_mode` both default to `True` (`app.py:56,59`), so the console path here is live on a bare `conductor run`. The PR description and the test docstring both describe it as verbose-only and rank it as the safer of the two sinks, which is backwards.

Fixing two of five sites is a reasonable scope call, but the gap should be visible. Either escape at these three sites too, or add them as `xfail(strict=True)` against a follow-up issue so a green suite doesn't read as "#382 is closed".

src/conductor/cli/run.py
```diff
@@ -376,7 +386,14 @@ def verbose_log_section(title: str, content: str) -> None:
...
-        _verbose_console.print(Panel(content, title=f"[cyan]{title}[/cyan]", border_style="dim"))
+        # ``content`` is agent-supplied (rendered prompts, tool output, model
+        # responses) and must not be parsed as markup -- a bracketed token like
+        # ``[/nestedType]`` in ordinary technical prose would raise MarkupError
+        # and kill the run (#382). ``title`` is conductor-controlled, so it keeps
+        # its styling.
+        _verbose_console.print(
+            Panel(escape(content), title=f"[cyan]{title}[/cyan]", border_style="dim")
```
jrob5756: `Text(content)` would be a better fit here than `escape()`.

`escape()` does not round-trip input that already contains a backslash before a bracket:

```
in : 'pattern: \[0-9\]+'
out: 'pattern: [0-9\]+'
```

Low impact on its own, but `Text` is byte-exact, it is what the other 13 call sites in this module already do, and it removes the need to explain why escaping is safe. The `[cyan]` title keeps its styling either way. The `escape` import at L23 could then go.

```suggestion
            Panel(Text(content), title=f"[cyan]{title}[/cyan]", border_style="dim")
```

src/conductor/cli/run.py
```diff
@@ -376,7 +386,14 @@ def verbose_log_section(title: str, content: str) -> None:
...
+        )
 
     # File always gets full untruncated content
     if _file_console is not None:
```
jrob5756, line 361: The file path deliberately does not escape, and that asymmetry is load-bearing but unexplained.

Escaping into a `markup=False` console writes the backslash out literally, which I confirmed:

```
| fmt: {p}/{t}\[/{nestedType}...]/read
```

Sitting eight lines under a comment explaining why `content` must be escaped, this reads like an oversight, and the obvious "consistency fix" would corrupt every log file. One line prevents that.

```suggestion
    if _file_console is not None:
        # Not escaped: ``_file_console`` has ``markup=False``, so escaping here
        # would write literal backslashes into the log.
```

tests/test_cli/test_markup_safety.py
```diff
@@ -0,0 +1,157 @@
...
+# Bracketed forms that appear in ordinary technical prose and that rich parses
+# as closing tags. The first is the real-world trigger: an Azure RBAC permission
+# format, verbatim from a plan document.
+MARKUP_TRIGGERS = [
+    "Permission format: {provider}/{type}[/{nestedType}...]/read",
+    "[/nestedType]",
+    "[/]",
+    "[/bold]",
+    "text [/foo] more",
+]
```
jrob5756, line 45: Every trigger here is a closing-tag form, so the suite only pins the loud failure.

Opening tags are the quiet one. `[bold]` and `[red]` never raise; rich consumes them and the text is gone. Your fix handles this correctly already, which is worth locking in: on `main` the tags are dropped, and with both changes they survive.

```
markup=False  : [bold] preserved=True
escape()      : [bold] preserved=True
no fix        : [bold] preserved=False
```

Without a case for it, a future narrowing of the escaper to a closing-tag-only regex would pass all 15 tests while silently dropping agent text, and nothing would ever raise to say so.

```suggestion
MARKUP_TRIGGERS = [
    "Permission format: {provider}/{type}[/{nestedType}...]/read",
    "[/nestedType]",
    "[/]",
    "[/bold]",
    "text [/foo] more",
]

# Opening tags do not raise -- rich consumes them silently. A crash-only suite
# cannot tell a correct fix from one that quietly drops agent text.
OPENING_TAG_TRIGGERS = [
    "Use [bold] to emphasise and [red] for errors",
    "Config key [section] then [other]",
]
```

tests/test_cli/test_markup_safety.py
```diff
@@ -0,0 +1,157 @@
...
+        ``init_file_logging`` — the behavioural tests above construct their own
+        console and would keep passing.
+        """
+        try:
+            run_module.init_file_logging(tmp_path / "run.log")
+            console = run_module._file_console
+            assert console is not None
+            assert console.no_color is True
+            # The property that prevents #382.
+            assert console._markup is False
```
jrob5756: `_markup` is private and `rich` is pinned only `>=13.0.0`, with no public accessor on `Console` in 14.3.1.

It fails loudly rather than silently if rich renames it, so it is a maintenance cost rather than a hole. A behavioural equivalent survives a rich upgrade and pins the contract that actually matters:

```python
console.print("[bold]not a style[/bold]")
...
assert "[bold]not a style[/bold]" in log.read_text(encoding="utf-8")
```

tests/test_cli/test_markup_safety.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def test_verbose_section_escapes_before_rendering(self, tmp_path: Path) -> None:
+        """Drive the real console path and assert it does not raise.
+
+        Without ``escape()`` at the call site this raises ``MarkupError``,
+        because ``_verbose_console`` deliberately keeps ``markup=True`` for the
+        ``[cyan]`` title.
+        """
+        try:
+            run_module.init_file_logging(tmp_path / "run.log")
```
jrob5756: This test's docstring says it drives the console path, but the `init_file_logging` call pulls the file sink in with it.

That costs you the isolating signal. Reverting each fix separately:

```
revert markup=False only : 3 failed, 12 passed
revert escape() only     : 1 failed, 14 passed
```

The one test that catches `escape()` is this one, and under the first mutation it dies on the file console before it ever reaches its console assertion. So no test isolates `escape()`, and the "3 of 15 fail" figure in the description holds only when both changes are reverted together.

Dropping this line gives clean 1:1 discrimination. Asserting all five triggers rather than `MARKUP_TRIGGERS[0]` at L157 costs nothing either.

jrob5756 (commented):
Nice find, and the reproduction in the issue made this easy to confirm. I appreciate you digging in here! I reproduced the crash, and both sinks you changed are genuinely safe now. Gates are clean here too: 521 passed / 3 skipped in `tests/test_cli`, plus ruff and `ty`.

Three things I would want resolved before we merge.

1. `markup=False` regresses the experimental-provider banner in the log file, and the comment added alongside it states an invariant that this same file already violates.
2. Three more sinks in `run.py` still raise on the same input, two of which I reproduced as real `conductor run` failures with no flags set.
3. The test fixtures are all closing-tag forms, so the suite passes against an implementation that silently discards agent text.

Two smaller ones that don't fit on a line in the diff.

**CHANGELOG.** Every commit merged since v0.1.27 updated `CHANGELOG.md`, and a crash that kills a run and leaves it unresumable is squarely the kind of entry the existing `### Fixed` section covers.

**Resume coverage.** The unresumability is the heart of the issue and nothing exercises it. It would also pin something no current test does: that the checkpoint stores agent text verbatim and escaping happens at render time. A well-meaning alternative fix that escaped at checkpoint-write would pass all 15 tests today and double-escape stored context on every resume.

Worth saying what is good here, because some of it is unusual. `TestMarkupTriggersAreRealTriggers` guards its own fixtures against passing vacuously, and that earned its keep during review when a reviewer's own draft test passed for the wrong reason. Using the real Azure RBAC string rather than a synthetic `[/foo]` will age well. And the `title` versus `content` split is the right trust boundary, explained rather than just applied.

franklixuefei:
Thanks — all three blockers were real, and I verified each at source before changing anything. Addressed in `37e53da`.

## `style=` does not disable markup — three more sinks

Checked directly rather than taking it on faith:

```python
Console(...).print("Permission: {p}/{t}[/nestedType]/read", style="red dim")
# MarkupError: closing tag '[/nestedType]' at position 19 doesn't match any open tag
```

All three now pass `rich.text.Text`. The two `error_msg` sinks are the ones that sting: that string carries a provider exception, so a workflow **already failing** had its real error replaced by a `MarkupError`.

And you are right that I had the risk ranking backwards — `verbose_mode` and `full_mode` both default to `True` (`app.py:56,59`), so these are reachable on a bare `conductor run`. I described it as verbose-only; it is not. The CHANGELOG entry says so.

I escaped all three rather than `xfail`-ing them. Two of five fixed would have left a green suite reading as "#382 is closed" — your point exactly.

## The banner regression

Reproduced. `_maybe_print_experimental_banner` built a `Panel` from a markup-bearing string and printed that *same object* to both consoles:

```
verbose: │ Experimental provider in use: claude-agent-sdk (pin) │
file   : │ Experimental provider in use: [bold]claude-agent-sdk[/bold] ([dim]pin[/dim]) │
```

`Text.from_markup` resolves it once and renders identically on both — confirmed side by side. The comment now states the real consequence rather than an invariant nothing enforced.

Also dropped the `no_color` claim, which I verified is wrong:

```
force_terminal=True, no_color=True -> '\x1b[1mB\x1b[0m \x1b[3mI\x1b[0m R'
```

Markup looked harmless in the log because the sink is not a terminal, not because of `no_color`.

## `escape()` → `Text()`

Confirmed the round-trip loss:

```
in       : 'pattern: \[0-9\]+'
escape() : 'pattern: [0-9\]+'     <- backslash gone
Text()   : 'pattern: \[0-9\]+'     <- exact
```

Switched, dropped the now-unused `escape` import, and documented why the file path deliberately does **not** escape — that asymmetry sits eight lines below the comment explaining why `content` must be, so the obvious "consistency fix" would have corrupted every log.

## Tests

All three points taken:

- **Opening tags** now have their own triggers plus a negative control asserting the samples really are eaten with markup on. Without that, narrowing the escaper to a closing-tag-only regex would pass everything while silently dropping agent text.
- **`console._markup`** replaced with a behavioural assertion (write markup, read it back from the log). `rich` is pinned only `>=13.0.0` and exposes no accessor, so the private attribute was a maintenance cost for a contract that can be pinned properly.
- **`init_file_logging` dropped** from the console-path test — it cost exactly the isolating signal you describe. It now asserts all five triggers, not just the first.

Discrimination, checked by reverting `src` and re-running:

| Outcome | Tests |
|---|---|
| ❌ fail without these fixes | 6 — five `verbose_log` cases + the banner |
| ✅ pass either way | 3 — they guard the fix this PR already shipped, not anything added here |

Stating the second row plainly rather than quoting "9 new tests", since three of them are not evidence for this commit.

## Gates

554 passed in `tests/test_cli` (541 without the new tests). `ruff check` unchanged at 2 pre-existing findings; `ty check src` unchanged at 64; my files `ruff format` and `ty` clean. Merged `main` first this time, so the CHANGELOG entry did not conflict.

One observation worth recording: `test_plugin_commands.py::TestFetch::test_a_second_fetch_reports_a_cache_hit` failed once in a full-suite run and passed on an identical rerun, in isolation, and on a pristine tree. It looks flaky rather than related, but flagging it rather than quietly re-running until green.

jrob5756 (approved):
LGTM. Approved!

PR #391, opened 2026-08-09
fix(cli): show non-ASCII workflow inputs literally in verbose output

### LINKED ISSUE #371: perf(cli): verbose prompt rendering hangs / OOMs on large markdown prompts due to Rich markup parsing
## Summary

When `conductor run` logs a rendered prompt in verbose mode, it passes the raw prompt string to `rich.Panel`. If the prompt contains markdown with many `[...]` brackets (e.g. Renovate / Node.js changelogs with links like `[text](url)` or ``[`hash`](url)``), Rich interprets the brackets as markup tags. This creates thousands of unmatched style spans and triggers catastrophic CPU/memory scaling in `rich.markup.render` / `rich.text.Text.render`, eventually hanging the process or getting OOM-killed.

The same prompt with `--quiet` runs in seconds without any memory spike.

## Reproduction

A minimal reproduction renders a realistic markdown block through `rich.Panel`:

```python
import time
from rich.panel import Panel

# Realistic Node.js-style changelog line with many brackets
line = (
    "- **crypto**: harden KeyObject internal slots (Filip Skokan) "
    "[[`ce84aef37d`](https://github.com/nodejs/node/commit/ce84aef37d)] "
    "[#63111](https://github.com/nodejs/node/pull/63111)\n"
)

# 100 KB of such markdown
content = line * 1000

console = Console(force_terminal=False, width=80)
start = time.perf_counter()
console.print(Panel(content, title="Prompt", border_style="dim"))
print(f"rendered in {(time.perf_counter() - start) * 1000:.1f} ms")
```

In Conductor the problematic path is:

- `src/conductor/executor/agent.py:308` calls `_verbose_log_section(f"Prompt for '{agent.name}'", rendered_prompt)`
- `src/conductor/cli/run.py:379` passes the raw string to `_verbose_console.print(Panel(content, ...))`

## Observed behavior

Measurements on a 80-column console, output redirected to `/dev/null`:

### Synthetic markdown scaling

| size   | plain text | markdown with brackets | slowdown |
|--------|-----------:|-----------------------:|---------:|
| 10 KB  | 3 ms       | 26 ms                  | 9x       |
| 25 KB  | 4 ms       | 147 ms                 | 33x      |
| 50 KB  | 8 ms       | 642 ms                 | 78x      |
| 100 KB | 15 ms      | 3261 ms                | 215x     |

Peak RSS for 100 KB markdown reached **311 MB**, while plain text of the same size stayed at **3.6 MB**.

### Real changelog subsample (`/tmp/changelog.md`, ~1 MB total)

| lines | chars   | render time | peak RSS |
|------:|--------:|------------:|---------:|
| 100   | 18 427  | 66 ms       | 10 MB    |
| 500   | 93 494  | 2.1 s       | 208 MB   |
| 1 000 | 189 617 | 9.7 s       | 866 MB   |
| 2 000 | 354 497 | 38 s        | 3.1 GB   |
| 5 707 | 998 791 | >60 s / OOM | >60 s / OOM |

A 200 KB real-world prompt (~1 000 lines) already takes **10 seconds and 866 MB** just to render the verbose panel. The full 1 MB file exceeds the 60-second timeout and is killed.

## Root cause

`Console.print(Panel(str, ...))` invokes `rich.markup.render` on the string because the default `Console` has markup parsing enabled. The regex `RE_TAGS` (`\[(\\\*)?([#/@a-z][^[]*?)\]`) matches every `[...]` in markdown links and code spans. Unmatched tags are converted into style spans, and `Text.render` then performs expensive sorting/stack operations over those spans.

`cProfile` shows the hot path:

```
ncalls  tottime  cumtime  filename:lineno(function)
...     ...      11.490   rich/segment.py:309(split_and_crop_lines)
...     ...      10.497   {method 'extend' of 'list' objects}
...     ...      10.288   rich/console.py:1300(render)
...     ...       9.588   rich/panel.py:141(__rich_console__)
```

With bracket-heavy markdown the span count explodes, turning the per-line wrap/render loop into effectively quadratic or worse behavior.

## Expected behavior

Verbose logging should not make a workflow unusable just because the prompt contains markdown links. The console preview should render quickly and with bounded memory, regardless of prompt size.

## Possible fixes

1. **Disable markup parsing for prompt content** — the smallest fix. Wrap the content in `rich.text.Text(content)` before passing it to `Panel`. This bypasses `rich.markup.render` entirely and avoids the span explosion.

   ```python
   from rich.text import Text
   _verbose_console.print(Panel(Text(content), title=f"[cyan]{title}[/cyan]", border_style="dim"))
   ```

2. **Truncate console preview** — large prompts should not be dumped in full into the terminal scrollback. A preview of, say, 10 000 characters is enough for debugging and keeps rendering fast.

3. **Gate full prompt logging behind a separate debug flag** — keep the default verbose output short, and only write the complete prompt to the file log (`--log-file`) or when `CONDUCTOR_LOG_PROMPTS=1` is set.

4. **Use `markup=False` / `highlight=False`** on the content print call as a defensive default.

External alternatives such as manual box drawing or Textual's virtualized viewport were also considered, but option 1 is the minimal change.

## Workaround for users

Run with `--quiet`:

```bash
conductor --quiet run workflow.yaml
```

This skips the verbose panel rendering entirely and avoids the hang / memory spike.

hertznsk:
## Summary

Serialise the verbose "Workflow Inputs" panel with `ensure_ascii=False` so non-ASCII input values are shown as the text the user typed rather than `\uXXXX` escapes.

## Why

`json.dumps(inputs, indent=2)` used the default `ensure_ascii=True`, so running a workflow with a Cyrillic, CJK or emoji input rendered the panel as:

```
{
  "question": "\u043f\u043b\u0430\u043d \u4f60\u597d"
}
```

Every other JSON display path in the repo already passes `ensure_ascii=False` (issue #356, PR #359) with a comment explaining why: `gates/dialog.py`, `engine/workflow.py`, `engine/validator.py`, `engine/dialog_evaluator.py`. This call site was the last one that did not.

## Scope change

This branch originally fixed Rich parsing bracketed prompt text as console markup (#371). That problem was fixed independently on `main` by #387, which reached the same `Panel(Text(content), ...)` console fix and additionally set `markup=False` on the file-log console.

Measured on current `main`: the #371 reproduction (175 KB of link-heavy prompt) renders in 388 ms and 2.7 MB, down from 63 s and 993 MB, with the file log preserving content verbatim. The rest of this branch is therefore rebased away as redundant, and the duplicate test is dropped since `tests/test_cli/test_markup_safety.py` from #387 already covers both sinks more thoroughly.

## Validation

- New regression test `test_workflow_inputs_section_serializes_non_ascii_unescaped`, which fails without the change on `assert 'план 你好' in ...`
- Full suite: 5509 passed, 37 skipped
- `ruff check`, `ruff format --check` and `ty check` all clean

### DIFF
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -1698,7 +1698,9 @@ async def run_workflow_async(
         emitter.subscribe(console_subscriber.on_event)
 
         if inputs:
-            verbose_log_section("Workflow Inputs", json.dumps(inputs, indent=2))
+            # ``ensure_ascii=False`` so the panel shows real non-ASCII input values
+            # rather than ``\uXXXX`` escapes (issue #356).
+            verbose_log_section("Workflow Inputs", json.dumps(inputs, indent=2, ensure_ascii=False))
 
         # Apply provider override if specified.
         # Reassigning ``runtime.provider`` to a string re-triggers the
--- a/tests/test_cli/test_logging.py
+++ b/tests/test_cli/test_logging.py
@@ -480,6 +480,59 @@ class TestVerboseLogging:
             full_mode.reset(token_full)
             verbose_mode.reset(token_verbose)
 
+    async def test_workflow_inputs_section_serializes_non_ascii_unescaped(
+        self, tmp_path: Path
+    ) -> None:
+        """Test that the Workflow Inputs section renders non-ASCII values literally."""
+        from typing import Any
+
+        from conductor.cli.run import run_workflow_async
+        from conductor.config.schema import AgentDef
+        from conductor.providers.copilot import CopilotProvider
+
+        workflow_file = tmp_path / "wf.yaml"
+        workflow_file.write_text(
+            "workflow:\n"
+            "  name: inputs-encoding\n"
+            "  entry_point: answer\n"
+            "  input:\n"
+            "    question:\n"
+            "      type: string\n"
+            "      required: true\n"
+            "agents:\n"
+            "  - name: answer\n"
+            "    prompt: 'Echo {{ workflow.input.question }}'\n"
+            "    output:\n"
+            "      answer:\n"
+            "        type: string\n"
+            "    routes:\n"
+            "      - to: $end\n"
+            "output:\n"
+            "  answer: '{{ answer.output.answer }}'\n"
+        )
+
+        def mock_handler(agent: AgentDef, prompt: str, context: dict[str, Any]) -> dict[str, Any]:
+            return {"answer": "ok"}
+
+        sections: list[tuple[str, str]] = []
+
+        with (
+            patch("conductor.providers.registry.create_provider") as mock_factory,
+            patch(
+                "conductor.cli.run.verbose_log_section",
+                side_effect=lambda title, content: sections.append((title, content)),
+            ),
+        ):
+            mock_factory.return_value = CopilotProvider(mock_handler=mock_handler)
+            await run_workflow_async(workflow_file, {"question": "план 你好"})
+
+        # Requirement: the inputs panel must show non-ASCII input values as real
+        # text, not \uXXXX escapes (issue #356).
+        payload = next(content for title, content in sections if title == "Workflow Inputs")
+        assert "план 你好" in payload
+        assert "\\u4f60" not in payload
+        assert "\\u043f" not in payload
+
     def test_verbose_log_section_skipped_in_minimal_mode(self) -> None:
         """Test that verbose_log_section skips console output in MINIMAL mode."""
         from io import StringIO


src/conductor/cli/run.py
```diff
@@ -376,11 +379,13 @@ def verbose_log_section(title: str, content: str) -> None:
         return
 
     if should_console:
-        _verbose_console.print(Panel(content, title=f"[cyan]{title}[/cyan]", border_style="dim"))
+        _verbose_console.print(
+            Panel(Markdown(content), title=f"[cyan]{title}[/cyan]", border_style="dim")
+        )
```
jrob5756, line 358: Rich's Markdown renderer does not just neutralise the brackets, it rewrites the content. Confirmed to disappear from the panel: HTML/XML blocks, link URLs (`[text](url)` renders as bare `text`), leading indentation, and `_underscores_`. A tag-structured prompt with no blank lines renders as a completely empty box.

That last one bites in practice. `skills/loader.py:147` wraps injected skills in `<skills>`/`<skill name=...>`, and `executor/agent.py` prepends that to the prompt before this line logs it, so on `claude` and `hermes` the console silently shows the prompt minus every injected skill.

`Text` avoids markup parsing just as completely (0.22s / 2.9 MB against 1.15s / 9.0 MB for a 175 KB prompt) and shows what was actually sent. If you take this, also drop the now-unused `Markdown` import on line 370 and the fence on line 1713.

If you would rather keep Markdown, then the `content:` entry in the docstring needs to say so, because `content` has quietly become Markdown source and `executor/agent.py` passes arbitrary prompt text straight into it.

```suggestion
        _verbose_console.print(
            Panel(Text(content), title=f"[cyan]{title}[/cyan]", border_style="dim")
        )
```

src/conductor/cli/run.py
```diff
@@ -376,11 +379,13 @@ def verbose_log_section(title: str, content: str) -> None:
...
 
     # File always gets full untruncated content
```
jrob5756: Minor archaeology, worth fixing while you are in here. Commit `87476b7` deleted the `truncate` parameter and added this comment in the same change, so the "untruncated" contrast has never had anything to contrast against. Nothing in this function truncates the console.

What is true now is a fidelity difference, and it is exactly what the `Text` on the next line buys you.

```suggestion
    # File sink is written even when the console sink is gated off, and records
    # content verbatim: Text parses no markup, so the log is a faithful copy.
```

src/conductor/cli/run.py
```diff
@@ -1705,7 +1710,7 @@ async def run_workflow_async(
         emitter.subscribe(console_subscriber.on_event)
 
         if inputs:
-            verbose_log_section("Workflow Inputs", json.dumps(inputs, indent=2))
+            verbose_log_section("Workflow Inputs", f"```json\n{json.dumps(inputs, indent=2)}\n```")
```
jrob5756: This fence is load-bearing and nothing says so. Drop it and Markdown reflows the payload onto one line, losing `indent=2` entirely.

Two side effects worth knowing. The backticks reach the file sink as well, so the plain-text log now carries markdown syntax that exists only to satisfy the console. And nothing covers it: I reverted this hunk on its own and all 4968 tests still passed.

Moving the console to `Text` removes the problem rather than working around it. While you are on this line, it is also the last JSON display path in the repo still on the default `ensure_ascii`, so Cyrillic and CJK inputs render as `\uXXXX`. Compare `gates/dialog.py:616` and `engine/workflow.py:3113`, which both carry a comment explaining the flag.

```suggestion
            verbose_log_section("Workflow Inputs", json.dumps(inputs, indent=2, ensure_ascii=False))
```

tests/test_cli/test_logging.py
```diff
@@ -480,6 +480,40 @@ def test_verbose_log_section(self) -> None:
...
+            full_mode.reset(token_full)
+            verbose_mode.reset(token_verbose)
+
+        console_panel = console.print.call_args.args[0]
+        file_panel = file_console.print.call_args.args[0]
+        assert isinstance(console_panel, Panel)
+        assert isinstance(console_panel.renderable, Markdown)
+        assert isinstance(file_panel, Panel)
+        assert isinstance(file_panel.renderable, Text)
+        assert file_panel.renderable.plain == content
```
jrob5756: This asserts the renderer class rather than the behaviour, and that has a real cost. I swapped the console to `Panel(Text(content))`, which fixes #371 identically and loses no content, and this test fails on it. It rejects a correct fix.

It also never touches the brackets. Re-run it with `content = "hello"` and it still passes, so the witness string is currently decorative.

Below is a version I validated three ways: passes on this branch, passes with the `Text` alternative, and fails against pre-fix code on `assert "[bold]literal[/bold]" in rendered`. It uses the real-Console-into-StringIO pattern the neighbouring tests already use, and keeps your `.plain == content` assertion, which is the one genuinely behavioural check in the original.

```suggestion
        import re
        from io import StringIO
        from unittest.mock import MagicMock

        from rich.console import Console
        from rich.panel import Panel
        from rich.text import Text

        from conductor.cli import run as run_module

        output = StringIO()
        file_console = MagicMock()
        token_verbose = verbose_mode.set(True)
        token_full = full_mode.set(True)
        try:
            with (
                patch(
                    "conductor.cli.run._verbose_console",
                    Console(file=output, force_terminal=True, no_color=True, width=200),
                ),
                patch("conductor.cli.run._file_console", file_console),
            ):
                # Issue #371 regression witness: keep the brackets. Drop them and every
                # assertion below still passes on any string at all.
                content = "[bold]literal[/bold] [#63111] [/] tail"
                run_module.verbose_log_section("Prompt", content)
        finally:
            full_mode.reset(token_full)
            verbose_mode.reset(token_verbose)

        rendered = re.sub(r"\x1b\[[0-9;]*m", "", output.getvalue())
        assert "[bold]literal[/bold]" in rendered
        assert "[#63111]" in rendered
        assert "tail" in rendered

        file_panel = file_console.print.call_args.args[0]
        assert isinstance(file_panel, Panel)
        assert isinstance(file_panel.renderable, Text)
        assert file_panel.renderable.plain == content
```

jrob5756 (commented):
Nice find here. I reproduced #371 on main (175 KB of link-heavy prompt: 63s, 993 MB) and this branch renders the same input in about 0.3s. The file-log half is a real improvement the issue never even asked for, since `Text(content)` makes the `--log-file` artifact a faithful copy for the first time.

My one substantive worry is `Markdown` on the console. Rich's Markdown renderer rewrites content rather than escaping it, and some of what it discards matters here. A prompt wrapped in XML-ish tags with no blank lines renders as an empty panel. That is not hypothetical: `skills/loader.py` wraps injected skills in `<skills>`/`<skill>` tags and `executor/agent.py` prepends that before this logs it, so on `claude` and `hermes` the console shows a prompt with every injected skill missing. Link URLs vanish too, which is the exact content class from the issue's own reproduction.

`rich.text.Text` closes #371 just as completely, measured about 5x faster and 3x lighter, and it keeps the panel honest. It also makes the ```json fence unnecessary. That was option 1 in the issue, and I think it is the better trade. Details inline.

hertznsk:
Thanks, @jrob5756 — I switched the verbose console path to `Text(content)` so it fixes #371 without rewriting the diagnostic payload. The updated regression test now exercises the rendered behavior and covers Rich markup literals, XML-ish skill wrappers, and Markdown link source; the file sink still records the literal content.

I also removed the workflow-input Markdown fence, added `ensure_ascii=False`, and updated the stale file-sink comment. Markdown formatting itself was not a requirement for me: I chose it as an alternative to Rich markup because I assumed preserving the visual formatting might matter, but preserving exactly what was sent is the better contract for verbose diagnostics.

jrob5756:
Heads up @hertznsk, I rebased this branch and force-pushed, so please `git fetch` and reset before doing any more work on it.

Your v2 was correct. It got overtaken. While it was in review, #387 merged and fixed the same root cause in the same function, landing on `main` about three hours ago. It reached the identical `Panel(Text(content), ...)` console fix you did, arrived at independently, and additionally set `markup=False` on the file-log console, which makes the `Text()` on the file sink redundant. That put the branch in conflict and left most of it duplicating code already on `main`.

I checked what was still unique after #387 and it was one thing: `ensure_ascii=False` on the workflow-inputs `json.dumps`. So the branch is now that change plus a regression test, retitled to match. I dropped the added test in `test_logging.py` because `tests/test_cli/test_markup_safety.py` from #387 already covers both sinks, including a case yours did not: opening tags like `[bold]`, which Rich swallows without raising, so a crash-only test cannot tell a real fix from one that silently drops text.

Worth recording since the issue is still open: #371 is fixed on `main`. I measured the 175 KB reproduction at 388 ms and 2.7 MB, against 63 s and 993 MB before, and the file log keeps content verbatim.

Thanks for the work here, and for the thorough follow-up on the review comments. The diagnosis in #371 was accurate and the benchmarks in it were genuinely useful for verifying #387 too.

jrob5756 (approved):
LGTM. Approved!

PR #403, opened 2026-08-11
fix(cli): render verbose log content literally instead of as Rich markup

### LINKED ISSUE #402: MarkupError in verbose logging kills the workflow when agent output contains a Rich tag like [/bold]
## Summary

Verbose logging renders untrusted text as Rich console markup. An agent that writes a square-bracket tag in its output — `[/bold]` is enough — produces an unbalanced closing tag, Rich raises `MarkupError`, and the **entire workflow dies**.

```
MarkupError: closing tag '[/bold]' at position 1629 doesn't match any open tag
```

This is not a cosmetic logging defect. The exception propagates out of the log call and terminates the run.

## How it happened

A real run, killed at iteration 45 after seven epics had already been committed.

The `epic_reviewer` filed this blocking issue against a Textual screen:

> Registry, workflow, and input values are passed directly to Textual widgets as markup-enabled strings. Valid metadata such as `` `[/bold]` `` raises Rich MarkupError and crashes the screen. Escape all data-derived display values...

That finding was interpolated into the next agent's prompt. Rendering **that prompt** raised the exact error the finding describes. The reviewer correctly diagnosed a markup-escaping bug, and reporting it triggered the same bug one layer up.

Confirmed from the event log — the offending text sits at position 1629 of the rendered prompt, matching the error:

```python
rendered_prompt[1560:1700]
# '...to Textual widgets as markup-enabled strings. Valid metadata such as `[/bold]` raises Rich MarkupError and crashes...'
```

## Root cause

`src/conductor/cli/run.py`, two functions.

`verbose_log_section` passes `content` straight to `Panel`:

```python
if should_console:
    _verbose_console.print(Panel(content, title=f"[cyan]{title}[/cyan]", border_style="dim"))

if _file_console is not None:
    _file_console.print(Panel(content, title=title, border_style="dim"))
```

`verbose_log` has the same flaw, interpolating into a markup string:

```python
_verbose_console.print(f"[{style}]{message}[/{style}]")
```

Rich treats a plain `str` as markup. `content` is called with the rendered prompt (`executor/agent.py:435`), which by construction contains agent output, plan text and tool arguments — none of it under conductor's control.

Minimal reproduction:

```python
from rich.panel import Panel
Console().print(Panel("Valid metadata such as `[/bold]` raises MarkupError."))
# rich.errors.MarkupError: closing tag '[/bold]' at position 24 ...
```

## Why it is worse than a verbosity flag

The file-console branch is guarded only by `_file_console is not None`, and file logging is enabled for **every `--web-bg` run**. So this fires on background runs regardless of `--quiet` — the mode where a long unattended workflow has the most to lose, and where the operator is least likely to be watching.

Any agent discussing Rich, Textual, BBCode, or writing `[skip ci]`, `[WIP]`, or a bracketed placeholder in a code span can trigger it.

## Fix

Wrap in `rich.text.Text`, which renders literally. `Text` accepts a style, so the console keeps its formatting:

```python
body = Text(content)
_verbose_console.print(Panel(body, title=f"[cyan]{title}[/cyan]", border_style="dim"))
```

PR incoming from `fix/verbose-log-section-markup-error`, with three regression tests that fail on `main` and pass with the fix (console path, file path, and `verbose_log`).

## Worth a follow-up

`rich.markup.escape` is used correctly in `cli/doctor.py` and `gates/interrupt.py`, so the pattern is understood — these two call sites simply predate it. A broader audit of `console.print` with interpolated runtime data would be worthwhile; the same class of bug is what the review agent was flagging in the Fleet Manager TUI when it hit this.

## Environment

- conductor: v0.1.27
- Repro'd on a `--web-bg` run; affects any run with file logging or `-V` console verbosity

jrob5756:
Fixes #402.

## The bug

`verbose_log_section` passed its content straight to `Panel`, and `verbose_log` interpolated its message into `f"[{style}]{message}[/{style}]"`. Rich parses a plain `str` as console markup, so any square-bracket tag in that text was interpreted rather than displayed.

That content is untrusted — `verbose_log_section` is called with the rendered prompt (`executor/agent.py:435`), which carries agent output, plan text and tool arguments verbatim. An agent that writes `[/bold]` in a code span produces an unbalanced closing tag, Rich raises `MarkupError`, and the exception propagates out of the log call and **kills the workflow**.

It killed a real run at iteration 45, after seven epics had been committed. The trigger is worth reading twice: a review agent had filed a finding about unescaped markup crashing a Textual screen, quoting `[/bold]` as the example. Rendering the prompt that carried that finding to the next agent raised the exact error the finding described.

The file-console path makes this more than a verbosity-flag bug — file logging is on for every `--web-bg` run regardless of console verbosity, so background runs are exposed even with `--quiet`.

## The fix

Wrap both in `rich.text.Text`, which renders literally. `Text` takes a style directly, so the console output keeps its formatting and the file output is unchanged apart from no longer being parsed.

## Tests

Three regression tests in `tests/test_cli/test_logging.py`, each asserting the literal tag survives to the output:

- `test_verbose_log_section_rejects_markup_interpretation` — console path
- `test_verbose_log_section_file_output_rejects_markup` — file path, reached with console verbosity off (the `--web-bg` case)
- `test_verbose_log_rejects_markup_interpretation` — `verbose_log`

Verified they fail on `main` with `MarkupError` and pass with the fix.

## Verification

- `make check` — ruff check, ruff format, ty: all pass
- `uv run pytest tests/test_cli/` — 529 passed, 3 skipped

## Note on scope

I fixed `verbose_log` alongside `verbose_log_section` even though only the latter caused the crash: it is the same one-line flaw in the adjacent function in the same file, and its callers interpolate plugin warnings and MCP server names that conductor does not control either. Happy to split it out if you would rather keep the change to the reproduced path.

`rich.markup.escape` is already used correctly in `cli/doctor.py` and `gates/interrupt.py`, so these two call sites look like they simply predate that pattern. #402 suggests a broader audit as follow-up.

### DIFF
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -136,16 +136,23 @@ def close_file_logging() -> None:
 def verbose_log(message: str, style: str = "dim") -> None:
     """Log a message if verbose mode is enabled.
 
+    ``message`` is rendered literally rather than as console markup, for the
+    same reason as :func:`verbose_log_section`: callers interpolate names,
+    paths and warning text that conductor does not control, and a stray
+    ``[/bold]`` in any of them would otherwise raise ``MarkupError``.
+
     Args:
-        message: The message to log.
-        style: Rich style for the message.
+        message: The message to log. Rendered literally, never as markup.
+        style: Rich style applied to the message on the console.
     """
+    from rich.text import Text
+
     from conductor.cli.app import is_verbose
 
     if is_verbose():
-        _verbose_console.print(f"[{style}]{message}[/{style}]")
+        _verbose_console.print(Text(message, style=style))
     if _file_console is not None:
-        _file_console.print(message)
+        _file_console.print(Text(message))
 
 
 def _describe_provider(provider: ProviderSettings) -> str:
@@ -317,10 +324,18 @@ def verbose_log_section(title: str, content: str) -> None:
     They are shown in FULL mode (default) but skipped in MINIMAL mode (--quiet).
     File logging always receives full content regardless of console verbosity.
 
+    ``content`` is untrusted — rendered prompts carry agent output, tool
+    arguments and plan text — so it is wrapped in :class:`~rich.text.Text` and
+    rendered literally. Passing the raw ``str`` would make Rich parse it as
+    console markup, and a bare ``[/bold]`` in an agent's own words is enough to
+    raise ``MarkupError`` and take the workflow down with it.
+
     Args:
         title: Section title.
-        content: Section content.
+        content: Section content. Rendered literally, never as markup.
     """
+    from rich.text import Text
+
     from conductor.cli.app import is_full, is_verbose
 
     # Sections are detail-level: show on console only in FULL mode
@@ -329,12 +344,14 @@ def verbose_log_section(title: str, content: str) -> None:
     if not should_console and not should_file:
         return
 
+    body = Text(content)
+
     if should_console:
-        _verbose_console.print(Panel(content, title=f"[cyan]{title}[/cyan]", border_style="dim"))
+        _verbose_console.print(Panel(body, title=f"[cyan]{title}[/cyan]", border_style="dim"))
 
     # File always gets full untruncated content
     if _file_console is not None:
-        _file_console.print(Panel(content, title=title, border_style="dim"))
+        _file_console.print(Panel(body, title=title, border_style="dim"))
 
 
 def verbose_log_timing(operation: str, elapsed: float) -> None:
--- a/tests/test_cli/test_logging.py
+++ b/tests/test_cli/test_logging.py
@@ -504,6 +504,91 @@ class TestVerboseLogging:
             full_mode.reset(token_full)
             verbose_mode.reset(token_verbose)
 
+    def test_verbose_log_section_rejects_markup_interpretation(self) -> None:
+        """Content that looks like Rich markup must not be parsed as markup.
+
+        Rendered prompts carry agent output verbatim, so a bare ``[/bold]`` in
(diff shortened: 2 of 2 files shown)


jrob5756:
Closing this as superseded by #387, which landed in a58c88f while this branch was open.

#387 fixes the same defect in both functions using the same `rich.text.Text` mechanism, and goes further:

- `markup=False` on `_file_console` in `init_file_logging`, which protects the whole file sink structurally rather than at each call site
- the two `error_msg` sinks in `verbose_log_parallel_agent_failed` and `verbose_log_for_each_item_failed`, where a provider exception was replacing the real error with a `MarkupError`
- `_maybe_print_experimental_banner`, resolved once via `Text.from_markup` so the panel renders correctly on both sinks
- `tests/test_cli/test_markup_safety.py`, which covers opening tags and closing tags with negative controls proving the samples are genuine triggers

Worth recording why this was not obvious: this branch was cut before #387 merged, so the PR diff GitHub shows (and `git diff main...branch`) compares against the merge base and looks like a clean fix. Diffing against current `main` instead shows this branch would **revert** all four of the changes above, including removing `markup=False` and restoring the two crashing error sinks. Rebasing would leave nothing behind, so there is nothing to salvage here.

One finding from reviewing this does survive on `main` and is not covered by #387: the panel `title` is still interpolated into markup at `src/conductor/cli/run.py:357`, and it is not conductor-controlled the way the comment above it claims. Filed separately.

jrob5756:
The residual `title` finding referenced above is now #406, filed as the wider CLI markup issue rather than a `run.py`-only fix, since `conductor validate` turned out to crash outright on the same input and `conductor status` corrupts its listing.

PR #407, opened 2026-08-11
fix(executor): make markdown linkification linear

### LINKED ISSUE #395: perf(executor): linkify_markdown is O(n²) on backtick/bracket runs and stalls the event loop
## Summary

`linkify_markdown` (`src/conductor/executor/linkify.py`) has **quadratic (O(n²)) worst-case runtime** on certain markdown inputs. Because it is a synchronous CPU call made directly on the async event loop, a large prompt stalls the *entire* Conductor process — the web dashboard stops responding, concurrent agents cannot progress, and `limits.timeout_seconds` cannot fire (the timeout enforcer shares the blocked loop).

This is the same class of problem as #371 (Rich markup parsing blowing up on bracket-heavy changelogs), but in a different code path — and it triggers regardless of `--quiet`.

## Measurements

Measured against `main` (v0.1.27), calling `linkify_markdown` directly:

| Payload | Size | Time |
|---|---|---|
| `` ` `` × n (single line) | 40 000 chars | **4.2 s** |
| `` ` `` × n (single line) | 80 000 chars | **17.0 s** |
| `` ` `` × n (single line) | 160 000 chars | **66.0 s** |
| `[` × n + `]` × n | 40 000 chars | 0.29 s |
| `[` × n + `]` × n | 80 000 chars | 1.1 s |
| normal prose | 40 000 chars | 0.004 s |

Cost quadruples each time input doubles — quadratic confirmed. Normal prose of the same size is ~4 ms, so this is purely pathological-input driven.

## Root causes

Three independent issues, all in `linkify.py`:

1. **`_FENCED_CODE_RE`** (line 55) — `^(`{3,}|~{3,}).*?^\1` with `DOTALL`. On a long single-line run of backticks, the greedy `{3,}` matches the whole run, `.*?` scans to EOF looking for a backreference that never matches, then `{3,}` backtracks one character and the scan repeats. n backtrack positions × O(n) scan each. **This is the most expensive of the three by ~12×.**

2. **`_EXISTING_LINK_RE`** (line 61) — `\[[^\]]*\]\([^)]*\)|\[[^\]]*\]\[[^\]]*\]`. On a wall of `[`, the engine retries `\[[^\]]*\]` at every `[` position, scanning forward O(n) each time.

3. **`_try_linkify_path`** (~lines 205, 210) — strips leading/trailing punctuation one character at a time via `stripped = stripped[1:]` / `stripped[:-1]`. Each slice copies the remaining string, so a token of repeated `"` or `.` characters costs O(n²) in copying. Much smaller constant than the regexes (~17 ms at 40 KB) but the same shape.

Note that (1) and (2) run inside `_linkify_with_protection` via `pattern.finditer(text)` on **every** gate prompt and dialog render.

## Affected call sites

All synchronous, all on the event loop:

- `engine/workflow.py:2506` — questions node intro
- `engine/workflow.py:4060` — `gate_presented` prompt (dashboard path)
- `gates/human.py:299` — terminal gate prompt
- `gates/dialog.py:623,637` — dialog output and opening question

## How large inputs get here

Gate prompts routinely interpolate raw agent output. The shipped `examples/test.yaml:264-274` is exactly this shape:

```yaml
  type: human_gate
  prompt: |
    ## Research Synthesis Review
    **Synthesis:**
    {{ synthesizer.output.synthesis }}
```

So any agent, MCP tool result, or input data that produces a long run of backticks or brackets — a minified file, a base64 blob, a malformed code fence, a generated table — reaches `linkify_markdown` verbatim. Brackets and backticks are not escaped by `json.dumps`, so they survive serialization of agent output intact.

## Suggested fix

1. Anchor `_FENCED_CODE_RE` so a fence opener must be followed by a newline (`^(`{3,}|~{3,})[^\n]*\n.*?^\1`), which removes the backtracking-into-the-fence-run behavior.
2. Make the bracket scan linear — either bound `[^\]]*` with a sane repetition cap, or replace `_EXISTING_LINK_RE` with a single-pass scanner.
3. Replace the character-at-a-time strip loops in `_try_linkify_path` with `str.lstrip()` / `str.rstrip()`.
4. Add a defensive input-size cap in `linkify_markdown` — above some threshold (a few hundred KB), return the text unmodified rather than attempting to linkify. Linkification is a display nicety; skipping it on huge inputs is strictly better than stalling the process.
5. Add a regression test asserting near-linear scaling on the pathological shapes above.

Fix (4) alone bounds the damage and is a one-line change; (1)–(3) remove the quadratic behavior properly.

macayu17:
Closes #395.

  ## What

  `linkify_markdown` could take seconds to process gate and dialog text containing long runs of backticks, brackets, or path punctuation.

  Because linkification runs synchronously, a pathological prompt blocked the event loop along with dashboard updates, concurrent agents, and timeout enforcement.

  ## Why

  Three paths repeatedly processed the remaining input:

  - The fenced-code regex could backtrack through a long opening fence one character at a time.
  - The existing-link regex retried unmatched `[` characters from every position.
  - Path punctuation was removed with repeated string slices, copying the remaining token on each iteration.

  On the original implementation, doubling a backtick run from 10,000 to 20,000 characters increased local runtime from 0.49s to 1.97s.

  ## How

  - Make fenced-code openers possessive and require the opener line to end with a newline.
  - Replace the existing-link regex with a forward-only scanner for `[text](url)` and `[text][ref]`.
  - Replace character-at-a-time path trimming with `str.lstrip()` and `str.rstrip()`.
  - Add regression coverage for reference links and near-linear scaling across the reported pathological inputs.

  The change stays inside the shared linkification helper, so all gate and dialog callers use the same fix.

  ## Validation

  - `pytest tests/test_executor/test_linkify.py` — 33 passed
  - `ruff check src tests` — passed
  - `ruff format --check src tests` — passed
  - `ty check src/conductor/executor/linkify.py` — passed
  - 40,000/80,000-character backtick runs completed in 0.0022s/0.0042s locally
  - 40,000/80,000-character bracket runs completed in 0.0018s/0.0045s locally

The full local Windows suite is not reported as passing because platform-specific tests outside this module failed. The focused linkification tests and all checks listed above pass.

### DIFF
--- a/src/conductor/executor/linkify.py
+++ b/src/conductor/executor/linkify.py
@@ -52,13 +52,10 @@ _LINKABLE_SUFFIXES = tuple(LINKABLE_EXTENSIONS)
 # ---------------------------------------------------------------------------
 
 # Fenced code block (``` or ~~~, with optional language tag)
-_FENCED_CODE_RE = re.compile(r"^(`{3,}|~{3,}).*?^\1", re.MULTILINE | re.DOTALL)
+_FENCED_CODE_RE = re.compile(r"^(`{3,}+|~{3,}+)[^\n]*\n.*?^\1", re.MULTILINE | re.DOTALL)
 
 # Inline code span (`...`)
 _INLINE_CODE_RE = re.compile(r"`[^`\n]+`")
-
-# Existing markdown links: [text](url) or [text][ref]
-_EXISTING_LINK_RE = re.compile(r"\[[^\]]*\]\([^)]*\)|\[[^\]]*\]\[[^\]]*\]")
 
 # Bare URL: http(s)://... terminated at whitespace or common punctuation
 _URL_RE = re.compile(
@@ -121,9 +118,10 @@ def _linkify_with_protection(text: str, base_dir: Path | None) -> str:
     """
     protected: list[tuple[int, int]] = []
 
-    for pattern in (_FENCED_CODE_RE, _INLINE_CODE_RE, _EXISTING_LINK_RE):
+    for pattern in (_FENCED_CODE_RE, _INLINE_CODE_RE):
         for m in pattern.finditer(text):
             protected.append((m.start(), m.end()))
+    protected.extend(_existing_link_spans(text))
 
     # Sort and merge overlapping spans
     protected.sort()
@@ -149,6 +147,39 @@ def _linkify_with_protection(text: str, base_dir: Path | None) -> str:
         result.append(_linkify_segment(text[prev_end:], base_dir))
 
     return "".join(result)
+
+
+def _existing_link_spans(text: str) -> list[tuple[int, int]]:
+    """Find existing ``[text](url)`` and ``[text][ref]`` links."""
+    spans: list[tuple[int, int]] = []
+    search_start = 0
+    unavailable_closers: set[str] = set()
+
+    while (start := text.find("[", search_start)) != -1:
+        label_end = text.find("]", start + 1)
+        if label_end == -1:
+            break
+
+        target_start = label_end + 1
+        if target_start == len(text) or text[target_start] not in "([":
+            search_start = target_start
+            continue
+
+        closer = ")" if text[target_start] == "(" else "]"
+        if closer in unavailable_closers:
+            search_start = target_start + 1
+            continue
+
+        target_end = text.find(closer, target_start + 1)
+        if target_end == -1:
+            unavailable_closers.add(closer)
+            search_start = target_start + 1
+            continue
+
+        spans.append((start, target_end + 1))
+        search_start = target_end + 1
+
+    return spans
 
 
 def _linkify_segment(segment: str, base_dir: Path | None) -> str:
@@ -200,19 +231,11 @@ def _try_linkify_path(token: str, base_dir: Path | None) -> str | None:
     Returns the markdown link string, or None if the token is not a file path.
     """
     # Strip leading/trailing punctuation that isn't part of the path
-    prefix = ""
-    suffix = ""
-    stripped = token
-
-    # Strip common leading chars
-    while stripped and stripped[0] in "([\"'":
-        prefix += stripped[0]
-        stripped = stripped[1:]
-
-    # Strip common trailing chars
-    while stripped and stripped[-1] in ")]\"'.,;:!?":
-        suffix = stripped[-1] + suffix
-        stripped = stripped[:-1]
+    stripped = token.lstrip("([\"'")
+    prefix = token[: len(token) - len(stripped)]
+    path = stripped.rstrip(")]\"'.,;:!?")
+    suffix = stripped[len(path) :]
+    stripped = path
 
     if not stripped:
         return None
--- a/tests/test_executor/test_linkify.py
+++ b/tests/test_executor/test_linkify.py
@@ -2,7 +2,11 @@
 
 from __future__ import annotations
 
+import time
+from collections.abc import Callable
 from pathlib import Path
+
+import pytest
 
 from conductor.executor.linkify import linkify_markdown
 
@@ -61,6 +65,10 @@ class TestUrlLinking:
 
     def test_preserves_existing_markdown_link(self) -> None:
         text = "See [docs](https://example.com/docs) for more"
+        assert linkify_markdown(text) == text
+
+    def test_preserves_existing_reference_link(self) -> None:
+        text = "See [docs][reference] for more"
         assert linkify_markdown(text) == text
 
     def test_url_in_inline_code_untouched(self) -> None:
@@ -211,3 +219,24 @@ class TestCombined:
         """Paths that escape base_dir should not be linked."""
         result = linkify_markdown("See ../../../etc/passwd.txt", base_dir=tmp_path)
         assert "[../../../etc/passwd.txt]" not in result
+
+
+@pytest.mark.performance
+@pytest.mark.parametrize(
+    "make_text",
+    [
+        pytest.param(lambda size: "`" * size, id="backticks"),
+        pytest.param(lambda size: "[" * (size // 2) + "]" * (size // 2), id="brackets"),
+        pytest.param(lambda size: '"' * size, id="punctuation"),
+    ],
+)
+def test_pathological_inputs_scale_near_linearly(make_text: Callable[[int], str]) -> None:
+    """Doubling pathological input must not quadruple processing time."""
+    elapsed: list[float] = []
+    for size in (20_000, 40_000):
+        text = make_text(size)
+        start = time.perf_counter()
+        assert linkify_markdown(text) == text
+        elapsed.append(time.perf_counter() - start)
+
+    assert elapsed[1] < elapsed[0] * 3 + 0.01


src/conductor/executor/linkify.py
```diff
@@ -52,14 +52,11 @@
 # ---------------------------------------------------------------------------
 
 # Fenced code block (``` or ~~~, with optional language tag)
-_FENCED_CODE_RE = re.compile(r"^(`{3,}|~{3,}).*?^\1", re.MULTILINE | re.DOTALL)
+_FENCED_CODE_RE = re.compile(r"^(`{3,}+|~{3,}+)[^\n]*\n.*?^\1", re.MULTILINE | re.DOTALL)
```
jrob5756, line 55: Making the opener possessive removes a behaviour along with the backtracking: a 4-backtick opener can no longer close on a 3-backtick fence, so that block stops being protected and its contents get rewritten.

```
input:  "````\nSee src/x.py\n```"
main:   "````\nSee src/x.py\n```"
this:   "````\nSee [src/x.py](src/x.py)\n```"
```

Across 60k fence-shaped inputs, 87,237 characters that `main` protected are now unprotected. CommonMark says a short closer does not close a longer fence, so the renderer still shows the region as code and the injected link syntax appears literally inside it. Models wrap nested code blocks in longer fences routinely, and that text reaches `gates/dialog.py:623` and `gates/human.py:299`.

Adding CommonMark's unterminated-fence fallback keeps the possessive quantifier and restores the protection. I measured zero characters under-protected against `main` on the same corpus, still linear at 0.10/0.18/0.37ms for 20k/40k/80k backticks, ruff-clean at 96 characters, and `tests/test_executor` plus `tests/test_gates` pass (91 tests).

```suggestion
_FENCED_CODE_RE = re.compile(r"^(`{3,}+|~{3,}+)[^\n]*\n(?:.*?^\1|.*)", re.MULTILINE | re.DOTALL)
```

The comment above this line is worth updating too. The `+` in `{3,}+` is the entire fix (161ms to 0.03ms on 8k backticks, while `[^\n]*\n` alone stays quadratic), and as written it reads like a typo someone will delete.

src/conductor/executor/linkify.py
```diff
@@ -151,6 +149,39 @@ def _linkify_with_protection(text: str, base_dir: Path | None) -> str:
     return "".join(result)
 
 
+def _existing_link_spans(text: str) -> list[tuple[int, int]]:
+    """Find existing ``[text](url)`` and ``[text][ref]`` links."""
+    spans: list[tuple[int, int]] = []
+    search_start = 0
+    unavailable_closers: set[str] = set()
```
jrob5756, line 156: This memo is what keeps `"[a](" * n` linear (6.7ms versus 16.8ms at 80k without it), and its correctness depends on `search_start` never moving backwards. That invariant is not written down anywhere, so a later change that backtracks would quietly stop protecting existing links, with no error and no test to catch it.

The `"]"` half is also never read back: once it is recorded, the next `text.find("]", start + 1)` returns -1 and the loop breaks before reaching the check. Confirmed over an exhaustive enumeration of `[]()x` up to length 10.

```suggestion
    # Closers proven absent from the rest of the text. Sound only because
    # search_start never moves backwards: once find(closer, i) misses, every
    # later start is > i and misses too. Without it, text full of unclosed
    # "[a](" candidates rescans the tail per candidate, which is quadratic.
    # Only ")" is ever read back; with no "]" left, the label scan breaks first.
    unavailable_closers: set[str] = set()
```

src/conductor/executor/linkify.py
```diff
@@ -200,19 +231,11 @@ def _try_linkify_path(token: str, base_dir: Path | None) -> str | None:
...
-
-    # Strip common trailing chars
-    while stripped and stripped[-1] in ")]\"'.,;:!?":
-        suffix = stripped[-1] + suffix
-        stripped = stripped[:-1]
+    stripped = token.lstrip("([\"'")
+    prefix = token[: len(token) - len(stripped)]
+    path = stripped.rstrip(")]\"'.,;:!?")
+    suffix = stripped[len(path) :]
+    stripped = path
```
jrob5756, line 238: `stripped` holds two different values four lines apart, and `path` is dead after the last line. Dropping line 238 during a later cleanup leaves `stripped` un-rstripped and everything still runs, so this is easy to break silently.

The comment can carry the part that is not obvious instead: that the trimmed characters are reattached rather than discarded.

```suggestion
    # Split off surrounding punctuation; prefix and suffix are reattached below
    # so "(see foo/bar.md)." round-trips. The trailing set is wider because
    # sentence punctuation can only follow a path, never precede it.
    unprefixed = token.lstrip("([\"'")
    prefix = token[: len(token) - len(unprefixed)]
    stripped = unprefixed.rstrip(")]\"'.,;:!?")
    suffix = unprefixed[len(stripped) :]
```

tests/test_executor/test_linkify.py
```diff
@@ -211,3 +219,24 @@ def test_path_traversal_blocked(self, tmp_path: Path) -> None:
         """Paths that escape base_dir should not be linked."""
         result = linkify_markdown("See ../../../etc/passwd.txt", base_dir=tmp_path)
         assert "[../../../etc/passwd.txt]" not in result
+
+
+@pytest.mark.performance
```
jrob5756, line 224: CI deselects this marker: both `.github/workflows/ci.yml:120` and `release.yml:108` run `-m "not real_api and not performance"`, and I confirmed 3 tests deselected locally. So the only regression guard for this fix never executes on a pull request.

Excluding the marker makes sense for ratio-based microbenchmarks, so the test's shape is the thing to change rather than the marker. The separation is wide enough that an absolute ceiling will not flake: 40k backticks is 1.5ms on this branch against 4.3s on `main`, and even the cheapest case at 320k is around 10ms against 1s.

```suggestion
# Deliberately unmarked: CI runs `-m "not performance"`, so a marked test cannot
# guard this fix. The ceiling below is absolute and sits far above the fixed-code
# cost, so it can only fire on a return to quadratic behaviour (#395).
```

tests/test_executor/test_linkify.py
```diff
@@ -211,3 +219,24 @@ def test_path_traversal_blocked(self, tmp_path: Path) -> None:
...
+@pytest.mark.parametrize(
+    "make_text",
+    [
+        pytest.param(lambda size: "`" * size, id="backticks"),
+        pytest.param(lambda size: "[" * (size // 2) + "]" * (size // 2), id="brackets"),
+        pytest.param(lambda size: '"' * size, id="punctuation"),
```
jrob5756, line 230: This parameter passes against `origin/main`, so it cannot catch the regression it exists for. Measured on the unfixed code: 20k takes about 5ms and 40k about 17ms, and `5 * 3 + 10ms` is 25ms, so it clears comfortably. The `+ 0.01` epsilon is larger than this case's entire pre-fix cost.

That leaves the `lstrip`/`rstrip` change with no coverage at all: reverting `_try_linkify_path` to the character loops keeps the suite green. Raising this case to around 320k separates the two cleanly (about 10ms fixed against about 1s broken), which needs `size` parametrized alongside `make_text` rather than fixed in the loop body.

A fourth case for `_wrap_url` would be worth adding at the same time, once that loop is fixed:

```python
pytest.param(lambda size: "https://example.com/x" + "." * size, id="url-punctuation"),
```

tests/test_executor/test_linkify.py
```diff
@@ -211,3 +219,24 @@ def test_path_traversal_blocked(self, tmp_path: Path) -> None:
...
+def test_pathological_inputs_scale_near_linearly(make_text: Callable[[int], str]) -> None:
+    """Doubling pathological input must not quadruple processing time."""
+    elapsed: list[float] = []
+    for size in (20_000, 40_000):
+        text = make_text(size)
+        start = time.perf_counter()
+        assert linkify_markdown(text) == text
+        elapsed.append(time.perf_counter() - start)
+
+    assert elapsed[1] < elapsed[0] * 3 + 0.01
```
jrob5756, line 242: The 10ms epsilon is doing most of the work in both directions. On the fixed code I saw ratios up to 3.19, above the threshold of 3, so the epsilon is what keeps it green; on the unfixed code it is what lets `punctuation` through. The effective assertion is already an accidental 12ms ceiling, just not a chosen one.

An explicit per-case ceiling would say what it means and match `tests/test_performance.py`, which uses absolute bounds throughout. Either way the failure message should carry the timings, since a bare ratio prints nothing to diagnose from.

```suggestion
    assert elapsed[1] < elapsed[0] * 3 + 0.01, (
        f"doubling input grew runtime from {elapsed[0] * 1e3:.1f}ms to {elapsed[1] * 1e3:.1f}ms"
    )
```

The docstring above says "must not quadruple" while the assertion allows 3x; worth aligning whichever number survives.

jrob5756 (commented):
The performance work here is solid, and two of the three rewrites are provably behaviour-preserving: I fuzzed `_existing_link_spans` against the regex it replaces over 200k inputs, and the `lstrip`/`rstrip` trimming over another 200k, with zero divergences in both. The monotonic `search_start` argument behind the memo holds up. Measured win on 20k backticks: 1400ms down to 0.74ms.

Three things I would want resolved before merge.

The possessive quantifier on the fence regex changes behaviour as well as complexity, and it silently stops protecting a code block whose opening fence is longer than its closing one. That reaches real agent output through the gate and dialog paths. Details and a verified drop-in are inline.

The regression test is marked `performance`, which CI deselects, so nothing guards the fix. And the `punctuation` parameter passes against `origin/main` as written, which leaves `_try_linkify_path` with no working coverage at all.

Two things sit outside the diff, so I could not anchor them:

`_wrap_url` (`linkify.py:194-204`) still has the exact character-at-a-time pattern this PR removes from `_try_linkify_path`. `_URL_RE` admits `.`, so `"https://a.example/x" + "." * n` drives the loop: I measured 16.5ms, 50.4ms, 189ms and 946ms at 25k/50k/100k/200k, roughly quadrupling per doubling. Applying this PR's own assertion to that input passes, so the new test does not see it. The `)` branch is unreachable, since `_URL_RE`'s character class excludes `)` and a Wikipedia URL arrives already truncated at `Foo_(bar`, so the whole loop collapses to `url = matched.rstrip(".,;:!?")`. Worth doing here, given the title claims linearity and closes #395.

There is also no `CHANGELOG.md` entry. This changes gate and dialog rendering, and recent user-facing fixes such as #393 and #387 added one under Unreleased.

PR #409, opened 2026-08-11
feat(claude-agent-sdk): continue one Claude session across executions via session_key

lukeellison:
Sorry for not opening an issue for this - I just needed it soon so happy to work off my own branch for a while and then leave this PR for your review in case you found it useful!

## Motivation

Every agent execution on the `claude-agent-sdk` provider starts a new `claude`
session. An agent that runs a second time reads the same files again and
repeats work it already did.

One workflow shape shows the cost. An agent investigates, a script step checks
the result, and a failed check routes back to the agent. Each pass starts
cold. Today there is no way to say "continue the session you already have".

## Approach

This PR adds an optional per-agent field, `session_key`. Executions that share
a key continue one Claude session. This covers a loop-back to the same agent,
and a hand-off to a later agent that declares the same key.

The key is a static label, not a value that a workflow step produces. The
provider maps the key to the real session id internally. No session id travels
through the workflow context, so the engine needs no new plumbing.

The provider keys sessions by the label and the working directory together,
because the CLI stores transcripts per directory. Conductor writes the map to
checkpoints, so continuity survives `conductor resume`.

The provider confirms that the transcript is on disk before it resumes.
`--resume` on a transcript the CLI cannot find stops the CLI before the agent
runs. Without this guard, a pruned transcript becomes a hard failure instead
of a fresh session.

A new `session_continuity` capability gates the field. A `session_key` on a
provider that cannot honor it is a `conductor validate` error. The validator
also rejects a key shared by executions that run at the same time, because
two `claude` processes cannot append to one transcript.

## Alternatives considered

**A `session_scope` enum** (`workflow`, `agent`, `item`, `none`), modelled on
the `aca` provider's `identifier_scope`. It covers more cases, but it adds a
workflow-level default and four values that no use case asked for. One key per
agent is smaller and solves the same problem.

**Exposing the real session id in the workflow context**, with a templated
`resume_session:` field. This is more flexible, because you can then resume a
session from an earlier run. It also needs a new `AgentOutput` field, context
plumbing, and engine changes. It composes on top of this PR, so nothing here
blocks it.

https://claude.ai/code/session_01RVihSFdo3stWanwWxDtnfj


CHANGELOG.md
```diff
@@ -321,6 +321,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   reachable and the declaration was simply inaccurate. Now bounded by
   `runtime.skill_injection` like `claude`.
 
+- **Session continuity for the `claude-agent-sdk` provider via a per-agent
```
jrob5756: Both new entries landed inside the `## [0.1.27] - 2026-08-04` section, which begins at line 232. `[Unreleased]` runs from line 8 to 231, and v0.1.27 is already a pushed tag with `pyproject.toml` still reading `version = "0.1.27"`.

As it stands, the 0.1.27 notes tell readers that `session_key` shipped on August 4th, and the next release will not pick either bullet up. Moving both under `[Unreleased]` then `### Added` fixes both halves.

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -46,6 +47,17 @@
...
     AgentDefinition: Any = None
 
+try:
+    # Separate try from the required symbols above: an SDK missing only these
+    # must degrade the session_key guard, not disable the whole provider.
+    from claude_agent_sdk import (  # ty: ignore[unresolved-import]
+        get_session_info,
+        project_key_for_directory,
+    )
+except ImportError:  # pragma: no cover - SDK below the session-lookup floor
```
jrob5756: I checked every cached SDK build at and above the declared pin: 0.2.82 (the exact floor), 0.2.87, 0.2.128, 0.2.132, 0.2.134. All five export both symbols from `__init__.py`, so no supported version sits below a "session-lookup floor".

The reachable failure is a different one, and it is worse. `project_key_for_directory` is re-exported from `_internal.session_store`, so an upstream move sets both names to None. From then on `_resolve_resume_session` returns None on every call and logs it at DEBUG, which Conductor never prints because it installs no logging handlers. `session_continuity=True` is a static class attribute, so `conductor validate` keeps accepting `session_key` and the experimental banner keeps advertising support while every keyed agent starts cold.

Two suggestions beyond the wording fix below. Derive the capability from actual availability rather than declaring it unconditionally, and warn once at construction when the symbols are missing. Also worth noting that the comment on lines 51-52 promises degradation when "an SDK missing only these" symbols is present, but a single import statement binds both or neither, so that mixed state cannot occur.

```suggestion
except ImportError:  # pragma: no cover - both symbols exist at our >=0.2.82 floor
```

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -985,6 +1036,139 @@ async def validate_connection(self) -> bool:
...
+        restored: dict[tuple[str, str], str] = {}
+        for raw_key, sid in ids.items():
+            if not raw_key.startswith(_SESSION_KEY_NAMESPACE):
+                continue
+            try:
+                key, cwd = json.loads(raw_key.removeprefix(_SESSION_KEY_NAMESPACE))
+            except (ValueError, TypeError):
+                logger.debug("Ignoring unreadable session map entry %r", raw_key)
+                continue
+            restored[(key, cwd)] = sid
```
jrob5756, line 1330: `restored[(key, cwd)] = sid` sits outside the `try`, so the docstring's promise that malformed entries are skipped rather than raised does not hold. Running these through the current code:

- `claude-agent-sdk:[["x"],["y"]]` raises an uncaught `TypeError` out of the method
- `claude-agent-sdk:[1, 2]` stores an int-keyed entry
- `claude-agent-sdk:"ab"` unpacks the two characters into a key
- `claude-agent-sdk:{"a":1,"b":2}` unpacks the dict keys

The last three violate the declared `dict[tuple[str, str], str]` and then re-export cleanly through `get_session_ids`, so one corrupt entry persists into every later checkpoint.

```suggestion
            try:
                parsed = json.loads(raw_key.removeprefix(_SESSION_KEY_NAMESPACE))
            except ValueError:
                parsed = None
            if not (
                isinstance(parsed, list)
                and len(parsed) == 2
                and all(isinstance(part, str) for part in parsed)
            ):
                logger.debug("Ignoring unreadable session map entry %r", raw_key)
                continue
            restored[(parsed[0], parsed[1])] = sid
```

src/conductor/config/validator.py
```diff
@@ -2531,6 +2541,54 @@ def _check_agent_capabilities(
...
+
+    for fe in config.for_each:
+        if fe.max_concurrent <= 1 or fe.agent.session_key is None:
+            continue
+        # A per-item working_dir gives each iteration its own session, so only
+        # a directory shared by every iteration is unsafe.
+        working_dir = _effective_working_dir(fe.agent) or ""
+        loop_vars = (f"{{{{ {fe.as_}", f"{{{{{fe.as_}", "_index", "_key")
+        if any(var in working_dir for var in loop_vars):
+            continue
```
jrob5756, line 2612: `_index` and `_key` match as bare substrings anywhere in the path, and the `{{ item` forms cover only two exact spacings. Run against this branch with `max_concurrent: 5` and a shared key:

| `working_dir` | result | should be |
|---|---|---|
| `/var/lib/search_index` | allowed | error |
| `/srv/api_key` | allowed | error |
| `/srv/{{ workflow.input.api_key }}` | allowed | error |
| `/tmp/{{- item }}` | error | allowed |

The false negatives switch the guard off for a directory every iteration shares, which is the one case it exists for. The false positive blocks a legal fan-out.

This file already parses templates properly at line 962, and `jinja2`, `meta` and `_JINJA_ENV` are all in scope here.

```suggestion
        working_dir = _effective_working_dir(fe.agent) or ""
        try:
            referenced = meta.find_undeclared_variables(_JINJA_ENV.parse(working_dir))
        except (jinja2.TemplateSyntaxError, jinja2.TemplateAssertionError):
            referenced = set()
        if referenced & {fe.as_, "_index", "_key"}:
            continue
```

src/conductor/config/validator.py
```diff
@@ -2531,6 +2541,54 @@ def _check_agent_capabilities(
                 f"This provider is not safe to run in parallel."
             )
 
+    # ----- Concurrent executions must not share a session -----
+    # Two executions resuming one session leaves two CLI processes appending to
+    # one transcript. Sessions are scoped to (session_key, working directory),
+    # so different directories are already distinct sessions and are not
+    # flagged — that is what makes multi-worktree fan-out legal.
+    runtime_working_dir = config.workflow.runtime.working_dir
```
jrob5756: `runtime_working_dir` is already bound at line 1794 in this same function, under a comment explaining that these are bound at the top ahead of the nested helpers, and it is read at 2443 and 2456. Re-binding the same expression 750 lines later means a later change to one is invisible from the other.

```suggestion
```

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -528,14 +548,17 @@ class ClaudeAgentSdkProvider(AgentProvider):
...
+        # ``session_continuity`` below is the granular, honest claim.
         checkpoint_resume=False,
         # Token counts come from ``ResultMessage.usage`` (cumulative
         # session total — see A4 fix).
         usage_tracking=True,
-        # No global mutable state shared across calls — the SDK spawns
-        # an independent subprocess per query() invocation.
+        # Each call spawns an independent subprocess. The ``session_key`` map
+        # is shared mutable state but is only read by agents that opted in;
+        # concurrent executions sharing one key are rejected by the validator.
```
jrob5756: The rejection cited here runs only under `conductor validate`. `validate_workflow_config` is reached from `cli/validate.py` alone, and `conductor run` calls `load_config` without it, so nothing re-checks concurrent key sharing on the path most runs take.

That matters more than usual because this comment is the entire justification for a safety-relevant flag. The wording it replaced (no global mutable state shared across calls) was at least self-contained.

```suggestion
        # Each call spawns an independent subprocess, and the ``session_key``
        # map is a plain dict mutated only from the event loop. Two executions
        # resuming one key concurrently is the unsafe case; ``conductor
        # validate`` rejects it statically, but ``conductor run`` does not.
```

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -985,6 +1036,139 @@ async def validate_connection(self) -> bool:
...
+
+        Checks the exact on-disk path the CLI uses,
+        ``<config>/projects/<project key for cwd>/<id>.jsonl``, rather than
+        trusting ``get_session_info``, which is wrong in both directions: it
+        derives a *summary* and returns ``None`` when it cannot, so a resumable
+        session can look absent; and it falls back to scanning sibling git
+        worktrees, so it reports sessions the CLI would then refuse to resume.
+        It is still consulted as a fallback — it tolerates a hash mismatch for
+        very long paths the exact check cannot model — but only when it agrees
+        the session belongs to this directory.
```
jrob5756: The worktree half of this reads backwards to me. Disassembling the bundled CLI, `--resume` resolves in three stages: the cwd's project dir, then a sibling git worktree fallback (instrumented as `tengu_resume_worktree_fallback`, iterating `git worktree list --porcelain`), then a global scan of `~/.claude/projects` for a unique match (`tengu_transcript_id_scan_fallback`).

So the CLI performs the same worktree fallback this cites as the reason to distrust `get_session_info`, and then goes wider still. A session that `get_session_info` finds only through that fallback would very likely resume fine rather than be refused.

The strict `realpath` gate below is defensible on scoping grounds, since it keeps the resolution inside the `(session_key, cwd)` contract the docs promise. The stated reason is what I would change, because as written it will steer the next maintainer away from loosening a gate that is currently stricter than the CLI.

```suggestion
        trusting ``get_session_info``, which derives a *summary* and returns
        ``None`` when it cannot, so a resumable session can look absent. It
        also resolves through sibling git worktrees and a global project scan,
        which is wider than the ``(session_key, cwd)`` scoping we promise
        authors, so its answer is accepted only when the ``cwd`` it recorded
        matches ours. It is still consulted as a fallback, since it tolerates
        a hash mismatch for very long paths the exact check cannot model.
```

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -695,9 +727,18 @@ async def execute(
...
+        resume_session_id = (
+            await self._resolve_resume_session(session_key, resolved_cwd) if session_key else None
+        )
+
         options = ClaudeAgentOptions(
             model=model,
             system_prompt=agent.system_prompt,
+            resume=resume_session_id,
+            # Explicit though it matches the SDK default: a fork would issue a
+            # new session id, stranding _session_ids on a dead one.
```
jrob5756: The default is right, but the reason does not hold. The capture site at lines 833-841 overwrites `_session_ids[(session_key, resolved_cwd)]` on every `AssistantMessage` and `ResultMessage`. Under a fork those frames carry the forked id, so the map would follow the new session rather than strand on the old one. A fork also leaves the original transcript in place, so "dead" is not accurate either.

The actual cost of forking is a fresh id and a fresh transcript on every execution, which is worth stating because someone may later want `fork_session=True` to work around the CLI's refusal to resume a session already running as a background agent.

```suggestion
            # Explicit though it matches the SDK default: forking would mint a
            # new session id and a new transcript on every execution, so the
            # key would chase a moving id and leave one-turn transcripts behind.
```

src/conductor/engine/workflow.py
```diff
@@ -2321,12 +2321,20 @@ def _write_checkpoint(
...
                     if hasattr(p, "get_session_ids"):
-                        copilot_session_ids = p.get_session_ids()  # type: ignore[union-attr]
+                        merged_ids.update(p.get_session_ids())  # type: ignore[union-attr]
                         if hasattr(p, "get_session_cwds"):
-                            copilot_session_cwds = p.get_session_cwds()  # type: ignore[union-attr]
-                        break
+                            merged_cwds.update(p.get_session_cwds())  # type: ignore[union-attr]
+                if merged_ids:
+                    copilot_session_ids = merged_ids
+                    copilot_session_cwds = merged_cwds
```
jrob5756: Dropping the `break` widened what the surrounding `except` can lose. Before, a provider that raised could only cost you its own map. Now a raise from any provider sends the whole loop to the handler, which sets `copilot_session_ids = None`, so Copilot's map goes with it and the long-standing non-experimental resume path loses its sessions. `merged_ids` is a local, so whatever was already collected is discarded too.

Catching per provider keeps the rest and lets the warning name the one that failed. The `if merged_ids:` guard can go at the same time: the single-provider branch above assigns unconditionally, and `checkpoint.py` already normalises an empty map with `or {}`.

```suggestion
                        try:
                            merged_ids.update(p.get_session_ids())  # type: ignore[union-attr]
                            if hasattr(p, "get_session_cwds"):
                                merged_cwds.update(p.get_session_cwds())  # type: ignore[union-attr]
                        except Exception as exc:
                            logger.warning(
                                "Provider %s failed to report session IDs for the "
                                "checkpoint (%s: %s); its sessions will not be "
                                "resumable, other providers' maps are kept.",
                                type(p).__name__,
                                type(exc).__name__,
                                exc,
                            )
                copilot_session_ids = merged_ids
                copilot_session_cwds = merged_cwds
```

src/conductor/config/schema.py
```diff
@@ -1796,6 +1818,21 @@ def validate_timeout(cls, v: int | None) -> int | None:
             raise ValueError("timeout must be a positive integer")
         return v
 
+    @field_validator("session_key")
+    @classmethod
+    def validate_session_key_is_literal(cls, v: str | None) -> str | None:
+        """Reject a Jinja2 template: ``session_key`` is never rendered, so
+        ``"item-{{ _key }}"`` would be one literal key shared by every
+        iteration rather than the per-item key the author intended.
+        """
```
jrob5756: Google style asks for a one-line summary. I scanned the three Python modules this PR touches (106 docstrings between them) and this is the only one opening with a multi-line summary; the adjacent `validate_timeout` is the usual shape. The rationale is worth keeping, just as the body.

```suggestion
        """Reject a Jinja2 template in ``session_key``.

        The field is never rendered, so ``"item-{{ _key }}"`` would become one
        literal key shared by every iteration rather than the per-item key the
        author intended.
        """
```

jrob5756 (commented):
Reviewed this against the cached SDK builds (0.2.82 through 0.2.134) and the bundled `claude` binary. The design reasoning here is careful and the docstrings are more honest than most, so most of what follows is about enforcement reaching `conductor run`, plus a few claims that did not survive checking.

Three items have no single line to hang them on.

**Both new guards only run under `conductor validate`.** `validate_workflow_config` has one caller, `cli/validate.py`. `conductor run` and `resume_workflow_async` call `load_config` and nothing else. So a `session_key` against copilot, claude, hermes or aca is accepted at run time and quietly does nothing, and two concurrent executions sharing a key are never stopped. The repo already treats this as a known trap: `_reject_unsupported_skills`, `_reject_unsupported_plugins` and `_reject_discovery_without_native_skills` all exist in `executor/agent.py` for this exact reason. An in-flight set of `(session_key, cwd)` in the provider would cover the concurrency half and the case below at the same time.

**Sub-workflows slip past the concurrency guard.** `_execute_subworkflow` passes `registry=self._registry`, so child engines share one provider instance and one `_session_ids` map. A `for_each` with `max_concurrent: 4` over a `type: workflow` agent whose inner agent is keyed produced three overlapping resumes of one live session, and `conductor validate` reported nothing. A `workflow` agent cannot carry a `session_key` itself, so the parent check never sees one, and the child is validated on its own.

**Usage accounting across a resumed session is untested in both directions.** The comment at line 556 states that `ResultMessage.usage` is a cumulative session total and points at `ApiUsage.apiUsage`, but that docstring belongs to the context-window TypedDict rather than to `ResultMessage.usage`. Before this PR the distinction could not matter, because every execution opened a new session. `session_key` makes "a ResultMessage for a session that already reported usage" reachable for the first time, and `UsageTracker.record` appends one row per execution. If the comment is right, a loop-back inflates reported cost and `budget_mode: enforce` can abort a run over spend that never happened. Worth a two-pass run against the live CLI and a test pinning whichever answer is correct.

Smaller things: `fork_session=False` is never asserted (the recorder mock simulates it, so flipping the source leaves every test green), the new example appears in no validation test, and `_write_checkpoint`'s docstring still says "for Copilot session resume".

`TestTranscriptProbe` deserves a mention. Using real config dirs and real transcript files, and asserting that `get_session_info` returns None before showing the exact-path check overrides it, pins a real upstream quirk that a mocked test would have missed entirely.

CHANGELOG.md
lukeellison: Fixed 👍 Both entries are now under `[Unreleased]`. They landed in the released section when I merged `main` into the branch.

src/conductor/providers/claude_agent_sdk.py
lukeellison: Fixed 👍 Parsing now happens inside the `try`, and an entry is stored only when it is a two-item list of strings. I checked all four of your cases against the built code, and each one is skipped now.

src/conductor/config/validator.py
lukeellison: Fixed 👍 It now parses the template with `meta.find_undeclared_variables`, as line 962 does. All four rows of your table are pinned as a parametrized test.

src/conductor/config/validator.py
lukeellison: Fixed 👍 Removed. The nested helper closes over the binding at 1794.

src/conductor/providers/claude_agent_sdk.py
lukeellison: You are right, thank you. I tested it: I made a session at the repo root and resumed it from `src/`, and the model recalled the earlier turn. Our reason was backwards, so I took your wording. The gate stays because it holds the `(session_key, cwd)` scope we promise authors.

src/conductor/providers/claude_agent_sdk.py
lukeellison: Fixed 👍 Took your wording. You are right that the map follows the forked id rather than strands, because capture overwrites on every frame.

src/conductor/engine/workflow.py
lukeellison: Fixed 👍 Each provider now has its own `try`, the warning names the provider that failed, and the `if merged_ids:` guard is gone.

src/conductor/config/schema.py
lukeellison: Fixed 👍 Took your wording.

src/conductor/providers/claude_agent_sdk.py
lukeellison: You are right about the import: it binds both names or neither, so I fixed that comment and took your `# pragma: no cover` line. I also fixed the same wrong claim inside `_resolve_resume_session`.

On deriving the capability, I went with a one-time `logger.warning` at construction instead, and I would value your view on whether that is enough. Three things made me hesitate. Every capability in the codebase is a literal that describes the provider's contract rather than the current machine — `working_dir` stays `True` when the `claude` binary is missing, and the descriptor still reads `True` when the SDK is not installed at all. `get_capabilities` reads the class without instantiating, which is what keeps `conductor validate` reproducible, so deriving from import success would let the same YAML validate on CI and fail on a laptop. And your own check across 0.2.82 to 0.2.134 bounds the risk to a future upstream move, where the real harm is silence rather than the declaration. Happy to make it dynamic if you would rather have that.

src/conductor/providers/claude_agent_sdk.py
lukeellison: Gone with a slightly different wording, because `conductor run` now does refuse this. The provider holds an in-flight set of `(session_key, cwd)` and raises on a second execution for a pair already running, so the comment says that instead. Are you fine with that?

lukeellison:
Thank you — this was a very useful review. Pushed as `62a811b`. Replies to the three items without a line anchor:

**Enforcement reaching `conductor run`.** Fixed 👍 You are right, and the precedent you pointed at was the answer. `_reject_unsupported_session_key` now sits in `executor/agent.py` next to `_reject_unsupported_skills`, as the first statement of `execute`.

**Sub-workflows.** Fixed 👍 The provider now holds an in-flight set of `(session_key, cwd)` and refuses a second execution for a pair already running, which covers this and the concurrency half together, as you suggested. There is a test that runs a real parent workflow with `for_each max_concurrent: 2` over a `type: workflow` step whose inner agent is keyed. `execute` became a thin wrapper around `_execute_session`, because `_build_output` can raise after the existing `finally` and the interrupt path returns mid-loop, so a claim released only there would leak the slot.

**Usage across a resumed session.** You are right that the comment was wrong, and it pointed at the wrong docstring. I ran the two passes against the live CLI: the resumed execution reports `input 2 / output 4 / num_turns 1`, so `ResultMessage.usage` is per-execution and there is no inflation. I corrected the comment, corrected the same claim on the `usage_tracking` capability line, and added a test that pins it.

Smaller ones: `fork_session=False` is now asserted on the options object rather than simulated by the mock, and `_write_checkpoint`'s docstring no longer says Copilot — that one is in `engine/workflow.py` rather than `engine/checkpoint.py`.

On the example, I think this one is already covered: `TestExamplesRegression` in `tests/test_config/test_validator.py` globs `examples/*.yaml`, so the new file has been validated since it landed. No test names it though, so I added one that also asserts the two agents share a key and the loop-back route exists, since the sweep would not notice the demonstration being edited away. Does that match what you meant?

One thing I left alone and want to flag: `_resolve_resume_session` bails when both lookup symbols are `None`, while the new constructor warning fires when either is. Under "both or neither" they agree, so behaviour is the same either way, but the asymmetry is visible now. Happy to tighten it here or in a follow-up.

jrob5756 (approved):
LGTM. Approved!

PR #418, opened 2026-08-12
fix(providers): restore live provider pricing by bumping the Copilot SDK

jrob5756:
Refs #386. Fixes the mechanism; the issue also asks about static-table entries, which this deliberately does not touch (see below).

## What was actually wrong

`CopilotProvider.get_model_pricing` reads `billing.token_prices`. The pinned `github-copilot-sdk` 1.0.1 ships **two** `ModelBilling` classes:

| file | declares | `from_dict` |
|---|---|---|
| `copilot/generated/rpc.py` | `multiplier`, `token_prices` | parses `tokenPrices` |
| `copilot/client.py` | `multiplier` only | reads `multiplier`, drops `tokenPrices` |

`client.list_models()` returns the second one, which is what the hook reads. So the field never left the API and was still modelled in the generated types; the client dataclass just threw it away.

The effect is that the hook returned `None` for every model. The chain #265 built — workflow override, then hook, then static table, then unpriced — ran permanently on its fallback, and any model missing from `DEFAULT_PRICING` reported no cost at all. #388 made that visible; this makes it stop happening.

1.0.9 parses the field again. Verified against both wheels rather than taken on trust, and end to end: a model parsed by the real `ModelInfo.from_dict` yields `$15.00` / `$75.00` per million tokens for `claude-opus-5`, which is 1.5 credits per 1k tokens at 100 credits to the dollar.

## Why the floor moves, not just the lock

`>=1.0.0` is satisfied by 1.0.1, so an existing environment would keep dead pricing while reporting a perfectly healthy dependency. The feature genuinely requires 1.0.9, so the floor should say so.

## Is the bump safe

I diffed every SDK method Conductor calls between the two versions by AST rather than by eye:

- `create_session` and `resume_session` gained 16 optional kwargs and lost nothing
- `list_models`, `start`, `stop`, `abort`, `destroy`, `disconnect`, `on`, `rpc`, `send`, `RuntimeConnection.for_uri` are unchanged
- `approve_all`'s `invocation` narrowed from `dict[str, str]` to a `PermissionInvocation` TypedDict

That last one is the `ty` error @franklixuefei reported on #388. `PermissionInvocation` is a TypedDict, so it was always dict-compatible at runtime; the annotation now matches.

One thing worth flagging that the type error hid: 1.0.9 also gave `approve_all` behaviour it did not have before.

```python
if invocation.get("managed_settings_enabled", False):
    raise RuntimeError("approve_all cannot be used when managed settings are enabled")
```

That flag is populated from `enable_managed_settings`, an opt-in kwarg on `create_session` that Conductor never passes, so the raise is unreachable. I recorded it in the handler docstring rather than leave the next person to trace it. If we ever do opt in, the blanket approve-all handler becomes a hard failure and will need to handle it.

## The static table

`claude-opus-5` and `gpt-5.6-sol` are still absent, on purpose. With the hook alive they price from the SDK, so #386's practical complaint is answered without anyone inventing per-token rates — which is the trade @franklixuefei declined to make on #388, correctly. Guessed rates are worse than no rates.

## Tests

The existing hook tests build their models from `SimpleNamespace`, so they assert what Conductor does *with* a billing object, never whether the SDK still hands it one. That is why 21 models silently went unpriced against a fully green suite.

Two new tests build the model through the SDK's own `ModelInfo.from_dict`. I confirmed they fail on 1.0.1 and pass on 1.0.9, so the next SDK release that drops the field fails the build instead of quietly reverting every run to static pricing.

## Validation

6091 passed, 47 skipped. `ruff check`, `ruff format --check` and `ty check src` all clean.


jrob5756:
Merged `main` in and resolved two conflicts. One of them was load-bearing, so flagging it rather than burying it in a merge commit.

#390 hardened the "Remove bundled Copilot CLI binary" step to fail when `find` matches nothing:

```bash
removed=$(find .venv -path '*/copilot/bin/copilot*' -print -delete)
if [ -z "$removed" ]; then
  echo "::error::No bundled Copilot CLI matched ..."
  exit 1
fi
```

The reasoning is right, and it is the same defect I found from the other end: a silent no-op reinstates the auth hang the step exists to prevent. But the assertion stops being true with this PR. SDK 1.0.1 ships per-platform wheels carrying `copilot/bin/`; 1.0.9 is a single pure-Python wheel that fetches the CLI on first use, so there is nothing to delete. Confirmed against both venvs.

So taking that side unchanged would have turned `main` red on the merge commit, for a condition that is correct rather than broken.

Resolution keeps everything #390 added — `shell: bash`, `set -euo pipefail`, and the pattern matching both `lib/pythonX.Y/` and Windows `Lib/` with the `.exe` suffix — and drops only the hard failure, replacing it with a message that says an empty match is expected. What now guarantees the original intent is `COPILOT_SKIP_CLI_DOWNLOAD=1` in the test step's env, which blocks the on-demand fetch that replaced the bundled binary. The step is now only there for a venv restored from an older lock.

Exercised the resolved script against all three layouts before pushing: old POSIX (`copilot/bin/copilot`), old Windows (`Lib/site-packages/copilot/bin/copilot.exe`), and 1.0.9 with no binary. All exit 0, and the first two still delete.

The env var sits on the test step, so it covers the Windows matrix entry #390 added as well as the Linux ones.

CHANGELOG conflict was the ordinary both-sides-appended kind; kept both, no wording changed.

6129 passed locally after the merge, ruff and ty clean.

PR #420, opened 2026-08-12
fix(mcp): support MCP 2.0 tool schemas

### LINKED ISSUE #419: MCP tool discovery crashes with MCP 2.0 Tool models
## Description

Conductor crashes while discovering tools when installed with `mcp==2.0.0`. The dependency constraint currently allows this version (`mcp>=1.28.1`), but `MCPManager` reads the MCP model attribute as `tool.inputSchema`.

MCP 2.0 renamed the Python attribute to `input_schema` while preserving `inputSchema` as its serialization alias. As a result, connecting any stdio MCP server can fail before the workflow starts.

## Error

```text
AttributeError: 'Tool' object has no attribute 'inputSchema'. Did you mean: 'input_schema'?
```

## Reproduction

1. Install Conductor in an environment that resolves `mcp==2.0.0`.
2. Configure a workflow with a stdio MCP server that exposes at least one tool.
3. Run the workflow.
4. Tool discovery fails in `src/conductor/mcp/manager.py` while reading `tool.inputSchema`.

A minimal model-level reproduction is:

```python
from mcp.types import Tool

tool = Tool(name="search", inputSchema={"type": "object"})
print(tool.inputSchema)
```

## Expected behavior

Conductor should discover tool schemas with both MCP 1.x and MCP 2.x model field naming.

## Proposed fix

Read the schema through Pydantic serialization with aliases (`tool.model_dump(by_alias=True)["inputSchema"]`), which produces the same key for both supported model shapes, and add a regression test modeled after MCP 2.0.

hertznsk:
## Summary

- read MCP tool schemas from the Pydantic model serialization alias instead of the removed camel-case Python attribute
- preserve compatibility with both MCP 1.x and MCP 2.x
- add regression coverage using the MCP 2.x snake-case field shape

## Testing

- `uv run pytest -q tests/test_mcp/test_manager.py tests/test_providers/test_pydantic_ai_mcp.py`
- `make check`
- manual MCP 2.0 tool-discovery smoke test

Fixes #419

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/microsoft/conductor/compare/v0.1.33...HEAD)
 
+### Fixed
+
+- **MCP tool discovery and structured tool results no longer break with MCP
+  2.0** (#419). MCP 2.0 renamed the Python field on `mcp.types.Tool` from
+  `inputSchema` to `input_schema` and on `mcp.types.CallToolResult` from
+  `structuredContent` to `structured_content`, retaining the camelCase name as
+  the serialization alias in both cases. The second rename failed quietly: a
+  tool returning only structured content raised `AttributeError`, which was
+  wrapped into a `RuntimeError` the model read as an ordinary tool failure.
+  Conductor now reads both fields through a shared helper that tries the 2.x
+  name and falls back to the 1.x one, preserving compatibility with both MCP
+  1.x and 2.x.
+
 ## [0.1.33](https://github.com/microsoft/conductor/compare/v0.1.32...v0.1.33) - 2026-08-18
 
 ### Added
--- a/src/conductor/mcp/manager.py
+++ b/src/conductor/mcp/manager.py
@@ -45,6 +45,17 @@ except ImportError:
 MCP_SDK_AVAILABLE = ClientSession is not None
 
 
+_MISSING: Any = object()
+
+
+def _mcp_field(model: Any, current_name: str, legacy_name: str) -> Any:
+    """Read a model field renamed between MCP 1.x and 2.x."""
+    value = getattr(model, current_name, _MISSING)
+    if value is _MISSING:
+        value = getattr(model, legacy_name)
+    return value
+
+
 # Marker constants. The generic hint is embedded by the manager and replaced
 # with the fs hint by Claude's agentic loop when filesystem-like tools are
 # available. No placeholder mechanism is used; callers replace the exact
@@ -204,7 +215,9 @@ class MCPManager:
                             {
                                 "name": prefixed_name,
                                 "description": tool.description or "",
-                                "input_schema": tool.model_dump(by_alias=True)["inputSchema"],
+                                # model_dump(by_alias=True) works for Tool, but the
+                                # helper keeps both rename sites in one idiom.
+                                "input_schema": _mcp_field(tool, "input_schema", "inputSchema"),
                                 "server": name,
                                 "original_name": tool.name,
                             }
@@ -327,8 +340,9 @@ class MCPManager:
             response_text = "\n".join(text_parts) if text_parts else ""
 
             # If no text content, try structured content
-            if not response_text and result.structuredContent:
-                response_text = str(result.structuredContent)
+            structured = _mcp_field(result, "structured_content", "structuredContent")
+            if not response_text and structured:
+                response_text = str(structured)
 
             try:
                 response_text = self._maybe_truncate_response(
--- a/src/conductor/mcp/manager.py
+++ b/src/conductor/mcp/manager.py
@@ -204,7 +204,7 @@ class MCPManager:
                             {
                                 "name": prefixed_name,
                                 "description": tool.description or "",
-                                "input_schema": tool.inputSchema,
+                                "input_schema": tool.model_dump(by_alias=True)["inputSchema"],
                                 "server": name,
                                 "original_name": tool.name,
                             }
--- a/tests/test_mcp/test_manager.py
+++ b/tests/test_mcp/test_manager.py
@@ -17,6 +17,53 @@ from unittest.mock import AsyncMock, MagicMock, patch
 
 import anyio
 import pytest
+from pydantic import BaseModel, Field
+
+
+class MCP2Tool(BaseModel):
+    """MCP 2.x-shaped stand-in for Tool with snake-case field names."""
+
+    name: str
+    description: str | None = None
+    input_schema: dict[str, Any] = Field(alias="inputSchema")
+
+
+class MCP2CallToolResult(BaseModel):
+    """MCP 2.x-shaped stand-in for CallToolResult with snake-case fields."""
+
+    content: list[Any] = []
+    structured_content: Any | None = Field(alias="structuredContent", default=None)
+    isError: bool = False
+
+
+def _make_mcp1_tool() -> Any:
+    """Build a real mcp.types.Tool using its 1.x field name."""
+    from mcp.types import Tool
+
+    return Tool.model_validate(
+        {
+            "name": "search",
+            "description": "Search the web",
+            "inputSchema": {
+                "type": "object",
+                "properties": {"query": {"type": "string"}},
+            },
+        }
+    )
+
+
+def _make_mcp2_tool() -> MCP2Tool:
+    """Build an MCP 2.x-shaped Tool stand-in."""
+    return MCP2Tool.model_validate(
+        {
+            "name": "search",
+            "description": "Search the web",
+            "inputSchema": {
+                "type": "object",
+                "properties": {"query": {"type": "string"}},
+            },
+        }
+    )
 
 
 class _TaskAffineMCP:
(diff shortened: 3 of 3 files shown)


src/conductor/mcp/manager.py
```diff
@@ -204,7 +204,7 @@ async def own_connection_lifecycle() -> None:
                             {
                                 "name": prefixed_name,
                                 "description": tool.description or "",
-                                "input_schema": tool.inputSchema,
+                                "input_schema": tool.model_dump(by_alias=True)["inputSchema"],
```
jrob5756: Confirmed this returns the schema on both 1.28.1 and 2.0.0.

The same rename hits `result.structuredContent` at lines 330-331, which MCP 2.0 renamed to `structured_content`. That path fails quietly: the AttributeError is caught at line 353, re-raised as a RuntimeError, and `mcp_toolset.py` turns it into a tool result the model reads as an ordinary tool failure. `max_retries=0` there, so there is no retry, and the dashboard records it as `agent_tool_complete` rather than an error. Only structured-output tools reach it, since `not response_text` short-circuits whenever there is text content, which is why it survives a smoke test.

One helper covering both sites would be easier to keep in sync than two separate idioms:

```python
_MISSING = object()

def _mcp_field(model: Any, current_name: str, legacy_name: str) -> Any:
    """Read a model field renamed between MCP 1.x and 2.x."""
    value = getattr(model, current_name, _MISSING)
    if value is _MISSING:
        value = getattr(model, legacy_name)
    return value
```

Two things I ran into while testing that are worth knowing. `model_dump` is fine here on `Tool` (about 1.8 microseconds, once per tool at connect) but not at the `call_tool` site, where `CallToolResult` carries the content payload up to `max_chars`, so dumping it per call copies the whole thing. And the shorter `getattr(m, new, None) or getattr(m, legacy)` is subtly wrong for this field, because `structured_content` is legitimately None on most results and would fall through to the 1.x name on the common path.

tests/test_mcp/test_manager.py
```diff
@@ -275,12 +275,24 @@ def manager(self) -> Any:
...
-        """Test connect_server with fully mocked MCP client."""
-        # Create mock tool
-        mock_tool = MagicMock()
-        mock_tool.name = "search"
-        mock_tool.description = "Search the web"
-        mock_tool.inputSchema = {"type": "object", "properties": {"query": {"type": "string"}}}
+        """Requirement: tool discovery accepts MCP 2.x snake-case model fields."""
+        from pydantic import BaseModel, Field
+
+        class MCP2Tool(BaseModel):
```
jrob5756: The 2.x fixture replaced the 1.x one rather than joining it, so nothing now covers the version the lockfile actually pins. I checked by setting line 207 to the 2.x-only `tool.input_schema`: `tests/test_mcp` still reports 53 passed, even though that breaks every user on mcp 1.28.1.

`mcp` is a hard dependency, so `mcp.types.Tool` imports freely here, and `Tool.model_validate({"name": ..., "inputSchema": ...})` works on both majors. Parametrizing over the real type plus this stand-in would pin both directions, and it keeps working when the floor eventually moves.

Worth noting the stand-in is faithful today. I checked `alias`, `validation_alias` and `serialization_alias` on real mcp 2.0.0 and all three are `inputSchema`. The gap is that nothing would tell you if that ever stopped being true.

tests/test_mcp/test_manager.py
```diff
@@ -275,12 +275,24 @@ def manager(self) -> Any:
...
 
     async def test_connect_server_mocked(self, manager: Any) -> None:
-        """Test connect_server with fully mocked MCP client."""
-        # Create mock tool
-        mock_tool = MagicMock()
-        mock_tool.name = "search"
-        mock_tool.description = "Search the web"
-        mock_tool.inputSchema = {"type": "object", "properties": {"query": {"type": "string"}}}
+        """Requirement: tool discovery accepts MCP 2.x snake-case model fields."""
+        from pydantic import BaseModel, Field
```
jrob5756: Minor: the other function-local imports in this file exist because the module has to be imported under the `MCP_SDK_AVAILABLE` patch. pydantic has no such constraint, so this one reads as an exception to a pattern that otherwise carries a signal. Moving it and the stand-in to module scope would also let a `call_tool` test reuse the model.

CHANGELOG.md
```diff
@@ -183,6 +183,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **MCP tool discovery no longer crashes with MCP 2.0** (#419). MCP 2.0
+  renamed the Python field on `mcp.types.Tool` from `inputSchema` to
+  `input_schema` while retaining `inputSchema` as its serialization alias.
+  Conductor now reads the aliased model representation, preserving compatibility
+  with both MCP 1.x and 2.x.
```
jrob5756: Discovery is compatible after this change. Tool execution is not, because of the `structuredContent` read at `mcp/manager.py:330-331`. Someone who hits that will read this entry and rule out MCP 2.0 as the cause, which is the debugging session the fix is meant to prevent.

If the second site lands in this PR, this sentence becomes true as written and you can ignore the suggestion. Otherwise it needs narrowing:

```suggestion
  Conductor now reads the aliased model representation, so tool discovery works
  under both MCP 1.x and 2.x. Tool calls whose results carry only structured
  content are still affected on 2.x.
```

jrob5756 (commented):
Thanks for tracking this down. I verified the fix against real installs of mcp 1.28.1 and 2.0.0, and reading the schema through the serialization alias works on both, with no pydantic serializer warnings. The diagnosis is right and the mechanism is the one I'd have picked.

My concern is scope rather than correctness. MCP 2.0 applied the same rename a second time in this file: `result.structuredContent` at lines 330-331 became `structured_content`. That one sits on the tool call path and it fails quietly. The AttributeError gets wrapped into a RuntimeError, handed to the model as a tool result string, and reported to the dashboard as `agent_tool_complete`. A run can finish "successfully" with the model having worked around a tool that never ran.

Second thing worth a look: the test swapped the 1.x fixture for a 2.x one rather than adding to it. If you change line 207 to the 2.x-only `tool.input_schema`, `tests/test_mcp` still reports 53 passed, while every user on the pinned floor breaks.

Details inline.

src/conductor/mcp/manager.py
hertznsk: Both points landed. The `structuredContent` read at lines 330-331 now goes through a shared `_mcp_field(model, current_name, legacy_name)` helper, exactly as you sketched — sentinel-based, with the legacy `getattr` raising AttributeError when neither name exists (same behavior as the pre-fix code on the call_tool site, where the existing wrap at line 353 still applies). The line-207 read moved to the same helper. I verified against real installs of mcp 1.28.1 and 2.0.0: `Tool.inputSchema`/`input_schema` and `CallToolResult.structuredContent`/`structured_content` are the two rename pairs, and on 2.0.0 all three aliases (`alias`, `validation_alias`, `serialization_alias`) hold the camelCase name.

On your two testing notes: `model_dump` is kept out of the call_tool site for the reason you give (the payload copy up to `max_chars` per call) — and the discovery site no longer uses it either, since the helper covers both. The `or`-chain trap is exactly why the sentinel is there: `structured_content` being legitimately None on most 2.x results would have fallen through to the legacy name on the common path.

tests/test_mcp/test_manager.py
hertznsk: Fixed. `test_connect_server_mocked` is now parametrized over both shapes: a real `mcp.types.Tool` (the 1.28.1 shape the lockfile pins, built via `Tool.model_validate({"name": ..., "inputSchema": ...})`) and the MCP-2.x stand-in, with ids `mcp1-real`/`mcp2-standin`. A new `test_call_tool_structured_content_both_shapes` does the same for the call_tool structured-content branch against a real `CallToolResult` and a 2.x-shaped stand-in. I mutation-checked all four one-name-only reads (each site × each single name) and confirmed each fails exactly the opposite parametrization — so a 2.x-only line 207 now fails `mcp1-real` instead of passing 53 green.

On the drift risk you flagged: fair — nothing pins the stand-in to a real 2.x install today. For now the parametrization over the real 1.x type pins one direction against the actual SDK, and the 2.x direction rests on the alias check I ran manually. A guard asserting the stand-in's aliases against a real 2.x install would need 2.x in the tree, which the floor doesn't allow yet — leaving that for when the floor moves.

tests/test_mcp/test_manager.py
hertznsk: Done. The pydantic import and both stand-ins (`MCP2Tool`, `MCP2CallToolResult`) are at module scope now, and the new call_tool test reuses the models. You're right that the other function-local imports carry a signal (the `MCP_SDK_AVAILABLE` patch) that pydantic doesn't — the local import read as an exception to that pattern.

CHANGELOG.md
hertznsk: The second site landed in this PR, so the entry now covers the tool-call path as well as discovery rather than the narrowed wording. It names both renames (`inputSchema`/`input_schema` and `structuredContent`/`structured_content`) and the quiet-failure mode you described, so someone hitting either can no longer rule out MCP 2.0 as the cause.

jrob5756 (approved):
LGTM. Approved!

Issue #401, opened 2026-08-10
conductor doctor table output dies part-written on a cp1252 console (hardcoded glyphs)


franklixuefei:
## What

`conductor doctor`'s default table output dies on a `cp1252` console, part-written.

`doctor.py` hardcodes three glyphs that cp1252 cannot encode:

```python
_CHECK = "[green]✓[/green]"          # U+2713
_CROSS = "[red]✗[/red]"              # U+2717
_OPTIONAL_MARK = "○"                 # U+25CB
```

Verified — all three raise `UnicodeEncodeError` under `cp1252`. (`_DASH`, U+2014, is encodable at `0x97` and is fine.)

## Why it is worse than one bad line

`_render_env` and `_render_providers` are separate `console.print()` calls, so stdout is left truncated mid-report rather than failing cleanly. And it lands on precisely the person least able to absorb it: someone who ran `doctor` *because* they hit #342.

## Scope

#381 fixes the JSON sinks (`--json` is safe via `ensure_ascii=True`) and deliberately leaves the table path alone, because every glyph consumer would need the console threaded through it — roughly 18 call sites across `_render_env`, `_render_providers` and six `_*_cell` helpers, none of which currently take a console.

## Suggested shape

Select glyphs once from the console's stream encoding and pass them down, falling back to ASCII (`OK` / `X` / `o`) when the stream cannot represent them:

```python
def _encodable(text: str, console: Console) -> bool:
    encoding = getattr(getattr(console, "file", None), "encoding", None)
    if not encoding:
        return True
    try:
        text.encode(encoding)
    except (UnicodeEncodeError, LookupError):
        return False
    return True
```

Rich does not check this itself — it hands the string to the underlying file, so the error surfaces from the write, after part of the report is already out.

## Repro

```bash
PYTHONIOENCODING=cp1252 conductor doctor
# exit 1, stdout ends after the Environment table
```

Issue #319, opened 2026-07-20
conductor doctor: offline view misleadingly shows copilot as unconfigured (all credential env vars ✗)


jrob5756:
## Summary

The default (offline) `conductor doctor` output makes the **copilot** provider look unconfigured/not-ready, even when it is fully authenticated and working. It renders all of copilot's credential environment variables as red ✗, with no indication that those env vars are *optional*. New users reasonably read this as "copilot is broken."

In reality, copilot authenticates via the **GitHub/Copilot CLI login on disk**, so those env vars are best-effort hints — not requirements. This is even stated in a code comment, but never surfaced in the output.

## Reproduction

With no `GITHUB_TOKEN` / `GH_TOKEN` / `COPILOT_PROVIDER_*` env vars set (but a valid Copilot CLI login on disk):

```console
$ conductor doctor
...
                                   Providers
┏━━━━━━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┓
┃ Provider         ┃ Insta… ┃ Tier  ┃ Credentials                     ┃ Notes  ┃
┡━━━━━━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━┩
│ copilot          │   ✓    │ stab… │ ✗ GITHUB_TOKEN                  │ —      │
│                  │        │       │ ✗ GH_TOKEN                      │        │
│                  │        │       │ ✗ COPILOT_PROVIDER_API_KEY      │        │
│                  │        │       │ ✗ COPILOT_PROVIDER_BEARER_TOKEN │        │
```

Every credential row is ✗ and the `Notes` column is empty — nothing tells the user this is expected/fine.

## Actual vs. expected

**Actual:** All copilot credential env vars show ✗, implying the provider is not ready.

**Expected:** The offline view should make clear copilot can authenticate via the GitHub/Copilot CLI login and that the env vars are optional overrides — so an all-✗ credentials cell for copilot does **not** signal a problem.

The provider is in fact ready — the connection probe confirms it:

```console
$ conductor doctor providers --check --provider copilot
│ copilot  │ ✓ │ stable │ ✗ GITHUB_TOKEN ... │ ✓ connected │ — │
```

## Root cause

[...]

Issue #455, opened 2026-08-17
bug(claude): validate_connection fails on Azure AI Foundry -- models.list() not supported


nskun:
# Summary:
ClaudeProvider.validate_connection() unconditionally calls client.models.list() to verify connectivity.
Azure AI Foundry's Anthropic endpoint does not support this API, causing validate_connection() to always return False.
This blocks workflow execution at startup.

# Environment
Conductor v0.1.31
Anthropic Python SDK(latest)
Endpoint: Azure AI Foundry(https://<resource>.services.ai.azure.com/anthropic)
Authentication: API key via ANTHROPIC_API_KEY

# Steps to Reproduce
1. Configure Conductor to use Azure AI Foundry's Anthropic endpoint:
```
export ANTHROPIC_BASE_URL=https://<resource>.services.ai.azure.com/anthropic
export ANTHROPIC_API_KEY=<key>
```

2.Run any workflow:
```
conductor run workflow.yaml
```

3.Observe: startup fails with connection validation error.

# Root Cause
In claude.py L379, validate_connection() calls
```
async def validate_connection(self) -> bool:
    # ...
    try:
        await self._client.models.list() # <- fails on Azure AI Foundry
        await self._log_available_models()
        return True
    except Exception as e:
        logger.error(f"Connection validation failed: {e}")
        return False # <- always reached
```

Azure AI Foundry's Anthropic endpoint does not implement the /v1/models API.
The models.list() call raises an exception(404 or similar), and validate_connection() returns False.

The actual inference endpoint(/v1/messages) works fine -- only the model listing is unsupported.

jrob5756:
Confirmed, and it's broader than Foundry: any Anthropic-compatible endpoint that doesn't implement `/v1/models` hits this.

I reproduced it against a stub gateway that 404s `GET /v1/models` and serves `POST /v1/messages` normally:

```
validate_connection() -> False
messages.create() -> 'pong'   (inference works)
```

`conductor run` then exits 1 with `ProviderError: Failed to connect to claude provider`, suggesting you check your credentials and network connection, neither of which was the problem.

Your root cause analysis is correct. Two things make this a defect rather than a misconfiguration.

First, `validate_connection()` is a hard startup gate. `registry.py` calls `create_provider(..., validate=True)` and `factory.py` turns a `False` into a fatal `ProviderError`.

[...]

PR #469, opened 2026-08-20
fix(cli): resolve doctor's glyph set per console encoding (#401)

AmirF194:
## Root cause

`conductor doctor`'s table output hardcodes `Text.from_markup("[green]✓[/green]")` / `[red]✗[/red]` / `[dim]—[/dim]` and the literal glyph `○` (`src/conductor/cli/doctor.py:38-51`), consumed across `_render_providers`, `_render_registries`, and six `_*_cell` helpers, none of which take a console/encoding parameter. Rich hands a rendered line straight to the underlying file's `write()` without checking whether the target encoding can represent it, so a run on a legacy Windows console (`cp1252`, which cannot encode `✓`/`✗`/`○`) raises `UnicodeEncodeError` mid-table, after the Environment section has already printed. The report is truncated and the process exits non-zero for a run that actually succeeded. `--json` already guards this with `ensure_ascii=True` (#381); the table path was deliberately left out of scope there and filed as this issue.

## Fix

Resolve a `_Glyphs` tuple once per `run_doctor()` call from the output console's actual stream encoding (`_resolve_glyphs`, `_encodable`), falling back to `OK`/`X`/`o`/`-` when the Unicode set isn't encodable, and thread it as an explicit parameter through every cell helper instead of reaching for a module-level constant. A console with no `.encoding` attribute (e.g. an in-memory buffer) is treated as capable, matching the existing "nothing to protect against" default. The `--json` path and the credential-rendering logic from #319 are unchanged.

## Verification

- New regression test (`TestDoctorEncodingFallback`) binds the CLI's console to a `cp1252`-encoded stream and asserts the run completes and renders `OK`/`X` instead of raising; a UTF-8 counterpart asserts the Unicode glyphs are still used there. Confirmed failing on `main` (`UnicodeEncodeError` at the same character the issue describes) and passing on this branch, in Docker.
- Full `tests/test_cli/test_doctor.py` (44 tests) green on Python 3.12 and 3.13 in Docker; `ruff check`, `ruff format --check`, and `ty check src` all clean (the 4 pre-existing `ty` warnings in `providers/claude_agent_sdk.py` are unrelated to this diff, confirmed by running `ty check` against unmodified `main`).
- Full suite (`pytest --cov`, 7671 passed) is green except 6 failures I traced to running as root in a container (permission/unreadable-file tests that assume a non-root user, e.g. `test_init_file_logging_permission_denied`) and one order-dependent flake (`test_renders_the_wordmark`, passes in isolation): reproduced identically against unmodified `main`, so none are caused by this change.
- Did not additionally run a full `conductor doctor` CLI invocation under `PYTHONIOENCODING=cp1252`; the regression test binds the module's own console object directly to a `cp1252`-encoded `TextIOWrapper`, which exercises the same write path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   as fatal.
 - `runtime.default_reasoning_effort` was silently dropped at run time for every
   provider and is now forwarded through `ProviderRegistry`.
+- **`conductor doctor`'s table output no longer dies part-written on a
+  `cp1252` console** (#401). The Installed/Credentials/Connection/Models
+  columns hardcoded `✓`/`✗`/`○`, none of which cp1252 can encode, so a run
+  on a legacy Windows console raised `UnicodeEncodeError` mid-table, after
+  the Environment section had already printed. `conductor doctor` now
+  resolves each glyph once per invocation against the output console's
+  stream encoding, falling back to `OK`/`X`/`o` when the Unicode glyphs
+  cannot be encoded; the `--json` path was already safe and is unchanged.
 - **MCP tool discovery and structured tool results no longer break with MCP
   2.0** (#419). MCP 2.0 renamed the Python field on `mcp.types.Tool` from
   `inputSchema` to `input_schema` and on `mcp.types.CallToolResult` from
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,12 +42,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   provider and is now forwarded through `ProviderRegistry`.
 - **`conductor doctor`'s table output no longer dies part-written on a
   `cp1252` console** (#401). The Installed/Credentials/Connection/Models
-  columns hardcoded `✓`/`✗`/`○`, none of which cp1252 can encode, so a run
-  on a legacy Windows console raised `UnicodeEncodeError` mid-table, after
-  the Environment section had already printed. `conductor doctor` now
-  resolves each glyph once per invocation against the output console's
-  stream encoding, falling back to `OK`/`X`/`o` when the Unicode glyphs
-  cannot be encoded; the `--json` path was already safe and is unchanged.
+  columns hardcoded `✓`/`✗`/`○`/`⚠`, none of which cp1252 can encode, so a
+  run on a legacy Windows console raised `UnicodeEncodeError` mid-table,
+  after the Environment section had already printed. `conductor doctor`
+  now resolves each glyph once per invocation against the output console's
+  stream encoding, falling back to `OK`/`X`/`o`/`!` when the Unicode
+  glyphs cannot be encoded; the `--json` path was already safe and is
+  unchanged.
 - **MCP tool discovery and structured tool results no longer break with MCP
   2.0** (#419). MCP 2.0 renamed the Python field on `mcp.types.Tool` from
   `inputSchema` to `input_schema` and on `mcp.types.CallToolResult` from
--- a/src/conductor/cli/doctor.py
+++ b/src/conductor/cli/doctor.py
@@ -41,6 +41,7 @@ class _Glyphs(NamedTuple):
     check: Text
     cross: Text
     dash: Text
+    warn: Text
     optional: str
     """Neutral glyph for an absent *optional* credential — deliberately not
     ``cross``, which is reserved for a genuinely missing required credential
@@ -51,23 +52,26 @@ _UNICODE_GLYPHS = _Glyphs(
     check=Text.from_markup("[green]✓[/green]"),
     cross=Text.from_markup("[red]✗[/red]"),
     dash=Text.from_markup("[dim]—[/dim]"),
+    warn=Text.from_markup("[yellow]⚠[/yellow]"),
     optional="○",
 )
 _ASCII_GLYPHS = _Glyphs(
     check=Text.from_markup("[green]OK[/green]"),
     cross=Text.from_markup("[red]X[/red]"),
     dash=Text.from_markup("[dim]-[/dim]"),
+    warn=Text.from_markup("[yellow]![/yellow]"),
     optional="o",
 )
 
 
 def _encodable(text: str, encoding: str | None) -> bool:
-    """Whether *text* survives a round trip through *encoding*.
+    """Whether *text* can be encoded to *encoding*.
 
-    ``encoding`` of ``None`` (the stream has no ``.encoding`` attribute, e.g.
-    an in-memory buffer) is treated as capable: there is nothing to protect
-    against, so it is better to render full-width Unicode than to needlessly
-    downgrade every invocation to ASCII.
+    A falsy ``encoding`` is treated as capable so an in-memory buffer is not
+    needlessly downgraded. ``io.StringIO`` has an ``.encoding`` of ``None``;
+    rich's ``NULL_FILE`` has no such attribute at all. This is a deliberate
+    fail-open: a stream that is lossy *and* silent about its encoding (e.g.
+    ``codecs.getwriter``) will still raise.
     """
     if not encoding:
         return True
@@ -78,21 +82,26 @@ def _encodable(text: str, encoding: str | None) -> bool:
     return True
 
 
-def _resolve_glyphs(console: Console) -> _Glyphs:
+def _resolve_glyphs(console: MarkupFreeConsole) -> _Glyphs:
     """Pick Unicode or ASCII-safe glyphs for *console*'s stream encoding.
 
     Rich hands a rendered line straight to the underlying file's ``write()``;
     it does not check whether the target encoding can represent it. A legacy
-    Windows console (``cp1252``) cannot encode ``✓``/``✗``/``○``, so the table
-    used to die mid-write, part-printed (issue #401). Resolved once per
-    ``run_doctor`` call and passed down, rather than re-checked per cell: the
-    stream's encoding cannot change mid-render.
+    Windows console (``cp1252``) cannot encode ``✓``/``✗``/``○``/``⚠``, so the
+    table dies mid-write, part-printed (issue #401). Resolved once per
+    ``run_doctor`` call and passed down rather than re-checked per cell, so
+    every cell in one report agrees.
+
+    Probed per glyph rather than through rich's ``ConsoleOptions.ascii_only``,
+    which is a ``startswith("utf")`` prefix test: ``gb18030`` encodes all of
+    these and that check would downgrade it for nothing.
     """
-    encoding = getattr(getattr(console, "file", None), "encoding", None)
+    encoding = console.encoding
     return _Glyphs(
         check=_UNICODE_GLYPHS.check if _encodable("✓", encoding) else _ASCII_GLYPHS.check,
         cross=_UNICODE_GLYPHS.cross if _encodable("✗", encoding) else _ASCII_GLYPHS.cross,
         dash=_UNICODE_GLYPHS.dash if _encodable("—", encoding) else _ASCII_GLYPHS.dash,
+        warn=_UNICODE_GLYPHS.warn if _encodable("⚠", encoding) else _ASCII_GLYPHS.warn,
         optional=_UNICODE_GLYPHS.optional if _encodable("○", encoding) else _ASCII_GLYPHS.optional,
     )
 
@@ -322,7 +331,7 @@ def _connection_cell(diag: ProviderDiagnostic, glyphs: _Glyphs) -> Text:
     if not diag.checked or diag.connection_ok is None:
         return glyphs.dash
     if diag.connection_ok and diag.connection_note:
-        return styled("[yellow]⚠[/yellow] {}", diag.connection_note)
+        return styled("{} {}", glyphs.warn, diag.connection_note)
     if diag.connection_ok:
         return styled("{} connected", glyphs.check)
     if diag.connection_error:
@@ -387,9 +396,9 @@ resolution itself failed). Built as module-level constants to avoid
 re-parsing the same markup literal on every table row (matching the
 ``_UNICODE_GLYPHS``/``_ASCII_GLYPHS`` constants above) — each markup argument
 is still a literal template, not an interpolated value, keeping this inside
-the repo's console rules (see AGENTS.md "Console Output"). None of these
-values are drawn from the encoding-sensitive glyph set, so they need no
-ASCII fallback."""
+the repo's console rules (see AGENTS.md "Console Output"). Every value here
+is pure ASCII, so no fallback applies: a property of these four literals,
+not a rule that anything outside :class:`_Glyphs` is safe to print."""
 
 
 def _rate_cell(value: float | None, glyphs: _Glyphs) -> Text:
@@ -430,7 +439,7 @@ def _render_models(providers: list[ProviderDiagnostic], console: Console, glyphs
     for diag in providers:
         if not diag.models:
             continue
-        table = Table(title=f"Models — {diag.name}", show_lines=True)
+        table = Table(title=f"Models {glyphs.dash.plain} {diag.name}", show_lines=True)
         table.add_column("Model", style="cyan", no_wrap=True)
         table.add_column("Reasoning efforts")
         table.add_column("Default")
--- a/src/conductor/cli/doctor.py
+++ b/src/conductor/cli/doctor.py
@@ -12,7 +12,7 @@ import asyncio
 import contextlib
 import logging
 from collections.abc import Iterator
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, NamedTuple
 
 from rich.table import Table
 from rich.text import Text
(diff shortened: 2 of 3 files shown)


src/conductor/cli/doctor.py
```diff
@@ -35,20 +35,66 @@
...
-   ``✓``, ``✗`` and ``○`` are not encodable in cp1252, so the default *table*
-   output of ``conductor doctor`` still fails on such a console — see #401.
-   This module's ``--json`` path is safe (``ensure_ascii=True``); the table path
-   is deliberately out of scope here because every glyph consumer would need the
-   console threaded through it. The em-dash is encodable in cp1252.
-"""
+    check: Text
+    cross: Text
+    dash: Text
+    optional: str
```
jrob5756, line 44: This reads as an inventory of the module's status glyphs, but it isn't one. `_connection_cell` prints a `⚠` that never comes through here, and cp1252 can't encode that either. Adding it as a field is the smallest change that makes the docstring true.
```suggestion
    """The status glyphs a table render uses, resolved for one console."""

    check: Text
    cross: Text
    dash: Text
    warn: Text
    optional: str
```

src/conductor/cli/doctor.py
```diff
@@ -35,20 +35,66 @@
...
+    cross=Text.from_markup("[red]✗[/red]"),
+    dash=Text.from_markup("[dim]—[/dim]"),
+    optional="○",
+)
+_ASCII_GLYPHS = _Glyphs(
+    check=Text.from_markup("[green]OK[/green]"),
+    cross=Text.from_markup("[red]X[/red]"),
+    dash=Text.from_markup("[dim]-[/dim]"),
+    optional="o",
+)
```
jrob5756, line 61: Matching `warn` entries for the field above.

Worth a look while you're here: `OK` is two cells wide where `o` and `X` are one, so the multi-line Credentials cell goes ragged on exactly the console this is meant to serve. Single-character ASCII (`+`/`x`/`o`/`-`) would keep the names lined up.
```suggestion
_UNICODE_GLYPHS = _Glyphs(
    check=Text.from_markup("[green]✓[/green]"),
    cross=Text.from_markup("[red]✗[/red]"),
    dash=Text.from_markup("[dim]—[/dim]"),
    warn=Text.from_markup("[yellow]⚠[/yellow]"),
    optional="○",
)
_ASCII_GLYPHS = _Glyphs(
    check=Text.from_markup("[green]OK[/green]"),
    cross=Text.from_markup("[red]X[/red]"),
    dash=Text.from_markup("[dim]-[/dim]"),
    warn=Text.from_markup("[yellow]![/yellow]"),
    optional="o",
)
```

src/conductor/cli/doctor.py
```diff
@@ -35,20 +35,66 @@
...
+
+
+def _encodable(text: str, encoding: str | None) -> bool:
+    """Whether *text* survives a round trip through *encoding*.
+
+    ``encoding`` of ``None`` (the stream has no ``.encoding`` attribute, e.g.
+    an in-memory buffer) is treated as capable: there is nothing to protect
+    against, so it is better to render full-width Unicode than to needlessly
+    downgrade every invocation to ASCII.
```
jrob5756, line 70: Two small corrections. `io.StringIO` does have an `.encoding` attribute, it is just `None` (your own test on line 934 asserts exactly that). The genuinely attribute-less case is rich's `NULL_FILE`. Also nothing here does a round trip, and none of these glyphs are full-width in the Unicode sense.

The `there is nothing to protect against` claim is the part I'd most want changed. `codecs.getwriter("cp1252")(...)` returns a real, lossy stream that reports no encoding at all, and this branch hands it `✓`. Harmless at today's call site, but it is the reasoning a future reader will lean on.
```suggestion
    """Whether *text* can be encoded to *encoding*.

    A falsy ``encoding`` is treated as capable so an in-memory buffer is not
    needlessly downgraded. ``io.StringIO`` has an ``.encoding`` of ``None``;
    rich's ``NULL_FILE`` has no such attribute at all. This is a deliberate
    fail-open: a stream that is lossy *and* silent about its encoding (e.g.
    ``codecs.getwriter``) will still raise.
    """
```

src/conductor/cli/doctor.py
```diff
@@ -35,20 +35,66 @@
...
+    ``run_doctor`` call and passed down, rather than re-checked per cell: the
+    stream's encoding cannot change mid-render.
+    """
+    encoding = getattr(getattr(console, "file", None), "encoding", None)
+    return _Glyphs(
+        check=_UNICODE_GLYPHS.check if _encodable("✓", encoding) else _ASCII_GLYPHS.check,
+        cross=_UNICODE_GLYPHS.cross if _encodable("✗", encoding) else _ASCII_GLYPHS.cross,
+        dash=_UNICODE_GLYPHS.dash if _encodable("—", encoding) else _ASCII_GLYPHS.dash,
+        optional=_UNICODE_GLYPHS.optional if _encodable("○", encoding) else _ASCII_GLYPHS.optional,
+    )
```
jrob5756, line 97: Three things folded together here.

The annotation should be `MarkupFreeConsole`. `console.py:168-171` calls that "the one place the guarantee can enter the type system", and rule A of the AST guard only catches construction, not annotations. It is already imported at line 20.

`console.encoding` is rich's public property and does the same `getattr(self.file, "encoding", "utf-8")` internally. I checked it against all three cases your tests cover and it agrees on every one, including `StringIO`. It is present in `rich>=13.0.0`, the floor in `pyproject.toml`, so there is no pin risk. The outer `getattr(console, "file", None)` can only fire if the property getter itself raises, which means a half-constructed `Console`.

And `used to die` isn't true yet, per the `⚠` on line 325.

Separately, and not something I'd hold the PR for: each glyph is written twice, once inside `Text.from_markup` and once as the probe argument 40 lines away. Change one and the other silently guards the wrong character. Deriving the probe from the constant would also mean a fifth glyph can't be added without being probed, which is how this one got through.
```suggestion
def _resolve_glyphs(console: MarkupFreeConsole) -> _Glyphs:
    """Pick Unicode or ASCII-safe glyphs for *console*'s stream encoding.

    Rich hands a rendered line straight to the underlying file's ``write()``;
    it does not check whether the target encoding can represent it. A legacy
    Windows console (``cp1252``) cannot encode ``✓``/``✗``/``○``/``⚠``, so the
    table dies mid-write, part-printed (issue #401). Resolved once per
    ``run_doctor`` call and passed down rather than re-checked per cell, so
    every cell in one report agrees.

    Probed per glyph rather than through rich's ``ConsoleOptions.ascii_only``,
    which is a ``startswith("utf")`` prefix test: ``gb18030`` encodes all of
    these and that check would downgrade it for nothing.
    """
    encoding = console.encoding
    return _Glyphs(
        check=_UNICODE_GLYPHS.check if _encodable("✓", encoding) else _ASCII_GLYPHS.check,
        cross=_UNICODE_GLYPHS.cross if _encodable("✗", encoding) else _ASCII_GLYPHS.cross,
        dash=_UNICODE_GLYPHS.dash if _encodable("—", encoding) else _ASCII_GLYPHS.dash,
        warn=_UNICODE_GLYPHS.warn if _encodable("⚠", encoding) else _ASCII_GLYPHS.warn,
        optional=_UNICODE_GLYPHS.optional if _encodable("○", encoding) else _ASCII_GLYPHS.optional,
    )
```

src/conductor/cli/doctor.py
```diff
@@ -257,32 +305,32 @@ def _credentials_cell(diag: ProviderDiagnostic) -> Text:
...
 
 
-def _connection_cell(diag: ProviderDiagnostic) -> Text:
+def _connection_cell(diag: ProviderDiagnostic, glyphs: _Glyphs) -> Text:
     """Format the connection-check result cell."""
     if not diag.checked or diag.connection_ok is None:
-        return _DASH
+        return glyphs.dash
     if diag.connection_ok and diag.connection_note:
         return styled("[yellow]⚠[/yellow] {}", diag.connection_note)
```
jrob5756, line 325: This is the one blocking item. `⚠` is U+26A0, cp1252 can't encode it, and it never goes through `_Glyphs` even though `glyphs` is in scope and used on the line below. `conductor doctor --check` still fails the way #401 describes:

```
conductor doctor --check   (cp1252 console)
-> UnicodeEncodeError: 'charmap' codec can't encode character '\u26a0' in position 376
```

It's a live path. `providers/claude.py:456` sets `_connection_probe_note` whenever the probe is inconclusive (#455), `providers/diagnostics.py:617-618` copies it into `connection_note`, and this branch renders it. So it fires for any Claude-compatible endpoint that doesn't implement `/v1/models`.

The outcome is also worse than the original bug rather than equal to it. The exception reaches `cli/app.py:182`, whose error panel title carries `❌` (U+274C, also not cp1252), so that write fails too and the user gets zero bytes and exit 1 instead of a partial table.
```suggestion
        return styled("{} {}", glyphs.warn, diag.connection_note)
```

src/conductor/cli/doctor.py
```diff
@@ -337,20 +385,22 @@ def _default_effort_cell(model: ModelDiagnostic) -> Text:
...
 resolution itself failed). Built as module-level constants to avoid
 re-parsing the same markup literal on every table row (matching the
-``_CHECK``/``_CROSS``/``_DASH`` constants above) — each markup argument is
-still a literal template, not an interpolated value, keeping this inside the
-repo's console rules (see AGENTS.md "Console Output")."""
+``_UNICODE_GLYPHS``/``_ASCII_GLYPHS`` constants above) — each markup argument
+is still a literal template, not an interpolated value, keeping this inside
+the repo's console rules (see AGENTS.md "Console Output"). None of these
+values are drawn from the encoding-sensitive glyph set, so they need no
+ASCII fallback."""
```
jrob5756, line 392: The claim is true of these four values, but it's phrased as a general rule: not from the glyph set, so no fallback needed. Apply that reasoning to the `⚠` on line 325 and you get the wrong answer. Worth pinning it to the literals rather than leaving it as an inference someone can follow into the same bug.
```suggestion
the repo's console rules (see AGENTS.md "Console Output"). Every value here
is pure ASCII, so no fallback applies — a property of these four literals,
not a rule that anything outside ``_Glyphs`` is safe to print."""
```

src/conductor/cli/doctor.py
```diff
@@ -368,7 +418,7 @@ def _pricing_source_cell(model: ModelDiagnostic) -> Text:
     return _PRICING_SOURCE_CELLS.get(model.pricing_source, Text(model.pricing_source))
 
 
-def _render_models(providers: list[ProviderDiagnostic], console: Console) -> None:
+def _render_models(providers: list[ProviderDiagnostic], console: Console, glyphs: _Glyphs) -> None:
```
jrob5756, line 421: Not on a line I can attach a suggestion to, so noting it here: line 433 is `Table(title=f"Models — {diag.name}", ...)` with a bare em dash. cp1252 encodes U+2014 so #401's scenario is unaffected, but `--models` still crashes on `ascii`, `latin-1` and `cp437`:

```
ascii + --models -> UnicodeEncodeError: '\u2014' in position 53
```

The irony is that `_resolve_glyphs` correctly downgrades `dash` to `-` on those encodings and the render dies anyway, on the same codepoint, in a function that already receives `glyphs`.

One caution if you fix it: I checked, and a `str` title picks up the table's italic `title_style` while a `Text` title silently loses it. So `styled(...)` is the wrong tool here. A plain `f"Models - {diag.name}"` keeps the styling and needs no glyph resolution.

CHANGELOG.md
```diff
@@ -9,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ### Fixed
 
+- **`conductor doctor`'s table output no longer dies part-written on a
+  `cp1252` console** (#401). The Installed/Credentials/Connection/Models
+  columns hardcoded `✓`/`✗`/`○`, none of which cp1252 can encode, so a run
+  on a legacy Windows console raised `UnicodeEncodeError` mid-table, after
+  the Environment section had already printed. `conductor doctor` now
+  resolves each glyph once per invocation against the output console's
+  stream encoding, falling back to `OK`/`X`/`o` when the Unicode glyphs
+  cannot be encoded; the `--json` path was already safe and is unchanged.
```
jrob5756, line 50: As written this is false for `--check` until line 325 is fixed, and a user who hits that crash will have read here that it was solved. The fallback list also omits `—` to `-`, which matters on `ascii` and `latin-1` even though it doesn't on cp1252.

If you take the `warn` fix, the original headline becomes accurate and you can drop the last sentence of the suggestion.
```suggestion
- **`conductor doctor`'s status glyphs degrade to ASCII on a console that
  cannot encode them** (#401). The Installed/Credentials/Connection/Models
  columns hardcoded `✓`/`✗`/`○`, none of which cp1252 can encode, so a run
  on a legacy Windows console raised `UnicodeEncodeError` mid-table, after
  the Environment section had already printed. `conductor doctor` now
  resolves each glyph once per invocation against the output console's
  stream encoding, falling back to `OK`/`X`/`o`/`-` when the Unicode glyphs
  cannot be encoded; the `--json` path was already safe and is unchanged.
```

tests/test_cli/test_doctor.py
```diff
@@ -859,3 +860,110 @@ def test_registries_load_error_renders(self, monkeypatch: pytest.MonkeyPatch) ->
...
+    def _bind_console(
+        self, monkeypatch: pytest.MonkeyPatch, encoding: str
+    ) -> tuple[io.BytesIO, io.TextIOWrapper]:
+        """Point the CLI's output console at a fresh buffer with *encoding*.
+
+        Returns the raw byte buffer rather than relying on ``result.output``:
+        the CLI's real console is bound directly to it, decoded with the same
+        *encoding* it was written with, so a character neither side can
+        represent surfaces as a decode error rather than being silently
+        swallowed by pytest's own UTF-8 capture.
```
jrob5756, line 883: The harness itself is the best part of this PR: binding the real console to a strict `TextIOWrapper` makes the stream do the asserting. The docstring describes it backwards though. A leaked glyph raises `UnicodeEncodeError` on the way *in*, at `write()` time, and the buffer stays empty, so the decode on line 911 never sees it. And the UTF-8 capture that would otherwise swallow it belongs to Click's `CliRunner`, not pytest.
```suggestion
        Returns the raw byte buffer rather than relying on ``result.output``:
        Click's ``CliRunner`` captures through a UTF-8 stream that would accept
        a glyph a real cp1252 console rejects, so a test could pass on output
        the user never gets. Binding the CLI's console to a ``TextIOWrapper``
        in the target *encoding* makes a leaked glyph raise
        ``UnicodeEncodeError`` at write time, surfacing via
        ``result.exception``.
```

tests/test_cli/test_doctor.py
```diff
@@ -859,3 +860,110 @@ def test_registries_load_error_renders(self, monkeypatch: pytest.MonkeyPatch) ->
...
+            providers=[_prov("copilot", installed=True)],
+            registries=RegistryDiagnostic(
+                default="local",
+                registries=[
+                    RegistryInfo(name="local", type="path", source="~/.conductor", is_default=True)
+                ],
+            ),
+        )
+        _patch_gather(monkeypatch, report)
+        result = runner.invoke(app, ["doctor"])
```
jrob5756, line 907: This is where the `⚠` slipped through, and I don't think the assertions are to blame. `assert result.exception is None` on line 909 already *is* a complete "every glyph survives" check, enforced by the codec at write time, with nobody having to guess which character is at fault. It just only ever runs against one fixture under one invocation.

Two gaps behind that. `_prov` (line 62) forwards eleven of `ProviderDiagnostic`'s twelve fields and the one it omits is `connection_note`, so no test in this file can construct the crashing branch. `grep -rn connection_note tests/` returns nothing for the whole repo. And `--check` and `--models`, the two flags that add the columns this PR rewrote, never touch the cp1252 stream at all.

I'd add `connection_note` to `_prov` and parametrize this test over diagnostic shapes crossed with `[]`, `--check` and `--models`. Enumerating shapes rather than characters is what makes it catch the next one. Note that once `connection_ok=False` shapes are in the matrix the assertion needs to become `assert not isinstance(result.exception, UnicodeEncodeError)`, since those exit 1 legitimately.

tests/test_cli/test_doctor.py
```diff
@@ -859,3 +860,110 @@ def test_registries_load_error_renders(self, monkeypatch: pytest.MonkeyPatch) ->
...
+        stream.flush()
+        assert result.exception is None
+        assert result.exit_code == 0
+        output = buffer.getvalue().decode("cp1252")  # raises if a raw glyph leaked through
+        assert "OK" in output
+        assert "✓" not in output
```
jrob5756, line 913: The comment credits the wrong line. The decode can't raise: only five byte values are undefined in cp1252 and a cp1252 encode can't produce any of them. Line 909 is what actually catches a leak, so a future cleanup that drops it as redundant would quietly remove the only real check. `assert "✓" not in output` is a tautology for the same reason.

Separately, `assert "OK" not in output` on line 925 is a whole-report negative on a two-letter string. I confirmed it breaks on a fixture as ordinary as `note="HTTP 200 OK from probe"`. Asserting on `_resolve_glyphs(...).check.plain`, or scoping to the Installed cell, would be sturdier.
```suggestion
        # Decode is exact, not lossy; a leaked glyph is caught by the
        # result.exception assertion above, which fails at write time.
        output = buffer.getvalue().decode("cp1252")
        assert "OK" in output
```

tests/test_cli/test_doctor.py
```diff
@@ -859,3 +860,110 @@ def test_registries_load_error_renders(self, monkeypatch: pytest.MonkeyPatch) ->
...
+class TestDoctorTierAndModelsErrorCells:
+    """Two more per-cell branches driven by the same glyph set as the
+    encoding-fallback tests above, exercised on the default UTF-8 path."""
+
+    def test_missing_tier_renders_dash(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        report = DoctorReport(providers=[_prov("copilot", tier=None)])
+        _patch_gather(monkeypatch, report)
+        result = runner.invoke(app, ["doctor"])
+        assert result.exit_code == 0
+        assert "—" in result.output
```
jrob5756, line 957: This one passes whether or not the branch it names exists. With `_prov`'s defaults the Credentials and Notes cells also render `—`, so the assertion holds with `tier="stable"` too. I checked. Deleting `_tier_cell`'s `if tier is None` branch wouldn't fail it.

Filling the other two cells makes the tier cell the only possible dash. Worth noting the class docstring above describes these as driven by the same glyph set as the encoding tests, but both run on UTF-8 against `result.output`; moving them onto the cp1252 stream would turn them into real regression tests for free.
```suggestion
    def test_missing_tier_renders_dash(self, monkeypatch: pytest.MonkeyPatch) -> None:
        report = DoctorReport(
            providers=[
                _prov(
                    "copilot",
                    tier=None,
                    creds=[CredentialEnvVar(name="COPILOT_TOKEN", present=True)],
                    note="see docs",
                )
            ]
        )
        _patch_gather(monkeypatch, report)
        result = runner.invoke(app, ["doctor"])
        assert result.exit_code == 0
        assert "—" in result.output
```

jrob5756 (commented):
Nice fix, and the shape is right: resolve once from the stream encoding and thread it explicitly instead of reaching for module globals. Probing per glyph rather than using rich's `ascii_only` is a better call than it might look, too. I checked `gb18030`, which encodes all of these but reports `ascii_only=True`, so rich's prefix test would have downgraded it for nothing. Your `_bind_console` harness is genuinely good: binding the real console to a strict `TextIOWrapper` makes the stream itself the assertion.

The blocker is that the glyph set isn't closed. `_connection_cell` still prints a bare `⚠` (U+26A0), which cp1252 can't encode either, so `conductor doctor --check` fails exactly the way #401 describes:

```
conductor doctor --check   (cp1252 console)
-> UnicodeEncodeError: 'charmap' codec can't encode character '\u26a0' in position 376
```

That's a shipped path, not a corner case. `providers/claude.py:456` sets `_connection_probe_note` on an inconclusive probe (#455), `providers/diagnostics.py:617-618` copies it to `connection_note`, and the branch renders it. The result is also worse than the bug you started from: the exception reaches `cli/app.py:182`, whose error panel title carries `❌` (also not cp1252), so that write fails as well and the user gets zero bytes and exit 1 rather than a partial table.

No test could have caught it. `_prov` has no `connection_note` parameter, and `grep -rn connection_note tests/` comes back empty across the whole repo.

A second one is sitting at line 433, `Table(title=f"Models — {diag.name}")`. cp1252 handles the em dash so #401's scenario is fine, but `--models` still dies on `ascii`, `latin-1` and `cp437` while `_resolve_glyphs` is busy correctly downgrading `dash` to `-` two hundred lines away.

That pattern is worth a moment. `test_markup_guards.py` opens by saying the behavioural tests can't prove the *next* call site is correct, and that #387 fixed `cli/run.py` thoroughly and still left a `Panel(title=)` in the function it changed. This is the same story with a different codepoint. A guard rule rejecting non-ASCII literals outside the glyph constants would have flagged both lines with file and line the day it was written, and it fits the existing rule A-H idiom exactly. Worth considering as a follow-up if not here.

Everything else below is smaller: a few docstrings that now claim more than the code delivers, and some test assertions that pass for the wrong reason.

Gates are all clean on my end: `ruff check`, `ruff format --check`, `ty check src` (fully green, no pre-existing warnings), 1106 passed in `tests/test_cli`, 164 markup-guard tests passed. No Console Output convention violations in the diff, and the removal half left no orphans behind, which is the usual failure mode for a change of this shape.

AmirF194:
Thanks for the thorough pass. Took the blocking fix and every suggestion: added a warn glyph to _Glyphs (falls back to !), routed the connection-note branch through it, switched _resolve_glyphs to MarkupFreeConsole.encoding, corrected the _encodable and _PRICING_SOURCE_CELLS docstrings, added connection_note to the test helper and parametrized the crash test over --check/--models (that's the actual gap, since those are the only flags that render the Connection column), fixed the misattributed comment and the tautological assert, and tightened the missing-tier test so it can only pass on the branch under test.

One divergence: for the Models title I kept it dash-resolving rather than hardcoding an ASCII hyphen, using glyphs.dash.plain so it's still a plain str (no title_style loss, verified). Preserves the unicode em dash on consoles that can take it while still downgrading where needed.

Left the AGENTS.md rule-A/H guard and the app.py:182 panel-title glyph as follow-ups, not in this PR. Full suite plus the doctor.py coverage report both green, pushed to 0ef11e3.

jrob5756 (approved):

AmirF194:
Thanks for merging, and for the detailed first pass, it made the fix better.

PR #473, opened 2026-08-21
feat(cli): add runtime.log_file YAML default for --log-file

### LINKED ISSUE #472: feat(cli): allow default log_file to be configured in workflow.yaml
# Summary
--log-file can only be specified as a CLI argument Workflow authors cannot set a default in the workflow definition.
Users must remember to pass --log-file auto every time -- when forgotten, crashes leave no diagnostic log.

# Problem
・When --log-file is not passed, crashes leave no diagnostic information.
・This is especially painful for long-running workflows(30-60 min) that crash near the end.
・default_model, max_tokens, temperature, timeout can already be configured in YAML, but log_file is the exception.

# Proposal
Add a log_file field to the runtime: block.

```
workflow:
  runtime:
    log_file: auto
```

Accepts the same values as CLI --log-file:
・auto -- Generate a timestamped path in $TMPDIR/conductor/ (same as --log-file auto).
・A file path (e.g. ./logs/run.log) -- Write to that path directly.
・Omitted -- No file logging (current default, unchanged).

# Precedence
CLI --log-file overrides the YAML value when both are specified.

・CLI --log-file specified -> Use CLI value
・CLI unset + YAML log_file set -> Use YAML value
・Neither set -> No logging (current behavior)

nskun:
# Summary
Add a log_file field to RuntimeConfig so workflow authors can set a default log destination in YAML.
The CLI --log-file flag continues to take precedence when specified.

# Motivation
Currently, file logging requires passing --log-file on every invocation.
When forgotten, crashes leave no diagnostic log.
This is especially painful for long-running workflows(30-60 min).
Other runtime fields(default_model, max_tokens, temperature, timeout) already support YAML defaults -- log_file should follow the same pattern.

# Usage
```
workflow:
  runtime:
    log_file: auto
```
Accepts the same values as CLI --log-file:
・auto -- Generate a timestamped path in $TMPDIR/conductor/.
・An explicit path (e.g. ./logs/run.log) -- Write to that path directly.
・Omitted -- No file logging(current default, unchanged).

# Design
CLI --log-file and YAML runtime.log_file become available at different times.
CLI is known before YAML parse: YAML is known only after load_config(). So:
・CLI --log-file specified -> log init before YAML parse(existing behavior, unchanged)
・CLI unset -> after load_config(), check runtime.log_file and init if present.
・CLI takes precedence when both are set.

Closes #472

### DIFF
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -139,6 +139,24 @@ def init_file_logging(log_path: Path) -> None:
     )
 
 
+def _try_init_file_logging(log_path: Path | None) -> bool:
+    """Initialize file logging, returning whether initialization succeeded."""
+    if log_path is None:
+        return False
+
+    try:
+        init_file_logging(log_path)
+    except OSError as e:
+        _verbose_console.print(
+            styled(
+                "[bold yellow]Warning:[/bold yellow] Cannot open log file {}: {}", log_path, e
+            )
+        )
+        return False
+
+    return True
+
+
 def close_file_logging() -> None:
     """Close file logging and clean up resources."""
     global _file_console, _file_handle
@@ -1959,15 +1977,7 @@ async def run_workflow_async(
     start_time = time.time()
 
     # Initialize file logging if requested
-    if log_file is not None:
-        try:
-            init_file_logging(log_file)
-        except OSError as e:
-            _verbose_console.print(
-                styled(
-                    "[bold yellow]Warning:[/bold yellow] Cannot open log file {}: {}", log_file, e
-                )
-            )
+    _try_init_file_logging(log_file)
 
     # Always create event emitter and JSONL log subscriber
     emitter = WorkflowEventEmitter()
@@ -1994,6 +2004,15 @@ async def run_workflow_async(
         load_start = time.time()
         config = load_config(workflow_path)
         verbose_log_timing("Configuration loaded", time.time() - load_start)
+
+        if log_file is None:
+            configured = config.workflow.runtime.log_file
+            if configured is not None:
+                if configured.lower() == "auto":
+                    log_file = generate_log_path(workflow_path.stem)
+                else:
+                    log_file = Path(configured)
+                _try_init_file_logging(log_file)
 
         # Merge CLI metadata on top of YAML-declared metadata
         if metadata:
@@ -2592,15 +2611,7 @@ async def resume_workflow_async(
     start_time = time.time()
 
     # Initialize file logging if requested
-    if log_file is not None:
-        try:
-            init_file_logging(log_file)
-        except OSError as e:
-            _verbose_console.print(
-                styled(
-                    "[bold yellow]Warning:[/bold yellow] Cannot open log file {}: {}", log_file, e
-                )
-            )
+    _try_init_file_logging(log_file)
 
     # Always create event emitter and JSONL log subscriber (parity with run)
     emitter = WorkflowEventEmitter()
@@ -2658,6 +2669,15 @@ async def resume_workflow_async(
         # Load workflow config first — needed both to construct the dashboard
         # (workflow_root) and to seed the synthetic replay fallback.
         config = load_config(resolved_workflow_path)
+
+        if log_file is None:
+            configured = config.workflow.runtime.log_file
+            if configured is not None:
+                if configured.lower() == "auto":
+                    log_file = generate_log_path(resolved_workflow_path.stem)
+                else:
+                    log_file = Path(configured)
+                _try_init_file_logging(log_file)
 
         # Merge CLI metadata on top of YAML-declared metadata (parity with run)
         if metadata:
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -148,9 +148,7 @@ def _try_init_file_logging(log_path: Path | None) -> bool:
         init_file_logging(log_path)
     except OSError as e:
         _verbose_console.print(
-            styled(
-                "[bold yellow]Warning:[/bold yellow] Cannot open log file {}: {}", log_path, e
-            )
+            styled("[bold yellow]Warning:[/bold yellow] Cannot open log file {}: {}", log_path, e)
         )
         return False
 
--- a/src/conductor/config/schema.py
+++ b/src/conductor/config/schema.py
@@ -3332,6 +3332,9 @@ class RuntimeConfig(BaseModel):
     default_model: str | None = None
     """Default model for agents that don't specify one."""
 
+    log_file: str | None = None
+    """Default log file path, or ``auto`` to generate one automatically."""
+
     mcp_servers: dict[str, MCPServerDef] = Field(default_factory=dict)
     """MCP server configurations keyed by server name."""
 
--- a/tests/test_cli/test_logging.py
+++ b/tests/test_cli/test_logging.py
@@ -395,6 +395,68 @@ class TestGenerateLogPath:
         assert str(path).startswith(tempfile.gettempdir())
 
 
+class TestTryInitFileLogging:
+    """Tests for _try_init_file_logging helper."""
+
+    def test_try_init_with_valid_path(self, tmp_path: Path) -> None:
+        """Test successful initialization with a valid path."""
+        from conductor.cli.run import (
+            _try_init_file_logging,
+            close_file_logging,
+            verbose_log,
+        )
+
+        log_path = tmp_path / "test.log"
+        token = verbose_mode.set(False)
+        try:
+            result = _try_init_file_logging(log_path)
+            assert result is True
+            assert log_path.exists()
+
+            verbose_log("hello from try_init")
+            close_file_logging()
+
+            content = log_path.read_text(encoding="utf-8")
+            assert "hello from try_init" in content
+        finally:
+            close_file_logging()
+            verbose_mode.reset(token)
+
+    def test_try_init_with_none(self) -> None:
+        """Test no-op when path is None."""
+        from conductor.cli.run import _try_init_file_logging
+
+        result = _try_init_file_logging(None)
+        assert result is False
+
+    @pytest.mark.skipif(sys.platform == "win32", reason="Unix-specific chmod permissions")
+    def test_try_init_with_permission_error(self, tmp_path: Path) -> None:
+        """Test graceful handling of OSError (warning to stderr, no raise)."""
+        import os
+        from io import StringIO
+
+        from rich.console import Console
+
+        from conductor.cli.run import _try_init_file_logging, close_file_logging
+
+        readonly_dir = tmp_path / "readonly"
+        readonly_dir.mkdir()
+        os.chmod(readonly_dir, 0o444)
+
+        stderr_output = StringIO()
+        mock_console = Console(file=stderr_output, no_color=True, highlight=False)
+
+        try:
+            with patch("conductor.cli.run._verbose_console", mock_console):
+                result = _try_init_file_logging(readonly_dir / "test.log")
+
+            assert result is False
+            assert "Cannot open log file" in stderr_output.getvalue()
+        finally:
+            close_file_logging()
+            os.chmod(readonly_dir, 0o755)
+
+
 class TestVerboseLogging:
     """Tests for verbose logging functions (migrated from test_verbose.py)."""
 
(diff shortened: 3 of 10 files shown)


src/conductor/cli/run.py
```diff
@@ -139,6 +139,24 @@ def init_file_logging(log_path: Path) -> None:
...
+    """Initialize file logging, returning whether initialization succeeded."""
+    if log_path is None:
+        return False
+
+    try:
+        init_file_logging(log_path)
+    except OSError as e:
+        _verbose_console.print(
+            styled(
+                "[bold yellow]Warning:[/bold yellow] Cannot open log file {}: {}", log_path, e
```
jrob5756: `ruff format` wants this on one line now that it's out of the nested `if` and fits in 100 chars. It's the only formatting deviation in the tree, and `.github/workflows/ci.yml:61` runs `ruff format --check`, so the lint job is failing on this alone.

```suggestion
            styled("[bold yellow]Warning:[/bold yellow] Cannot open log file {}: {}", log_path, e)
```

src/conductor/cli/run.py
```diff
@@ -1995,6 +2005,15 @@ async def run_workflow_async(
...
         verbose_log_timing("Configuration loaded", time.time() - load_start)
 
+        if log_file is None:
+            configured = config.workflow.runtime.log_file
+            if configured is not None:
+                if configured.lower() == "auto":
+                    log_file = generate_log_path(workflow_path.stem)
+                else:
+                    log_file = Path(configured)
+                _try_init_file_logging(log_file)
```
jrob5756, line 2013: Four things land in this block.

`generate_log_path()` is outside the guard. Its `mkdir(parents=True, exist_ok=True)` still raises `FileExistsError` when `$TMPDIR/conductor` is a file, and `PermissionError` or `ENOSPC` on a constrained runner. A logging setting shouldn't be able to abort the run, and it's inconsistent with the explicit-path branch one line below, which warns and continues.

`Path(configured)` doesn't expand `~`. Since `init_file_logging` creates parents, `log_file: ~/logs/run.log` quietly writes into a directory named `~` under the CWD and reports success. The shell saves the CLI flag from this; YAML doesn't.

A NUL byte in the value raises `ValueError`, which `except OSError` doesn't catch, so it escapes as an unhandled traceback.

And the `bool` from `_try_init_file_logging` is dropped here, so `log_file` keeps naming a file that was never opened. Nothing reads it wrongly today, but only because the `_file_console is not None` check at line 2313 happens to cover for it.

```suggestion
            if configured is not None:
                try:
                    if configured.strip().lower() == "auto":
                        log_file = generate_log_path(workflow_path.stem)
                    else:
                        log_file = Path(configured).expanduser()
                except (OSError, ValueError) as e:
                    _verbose_console.print(
                        styled(
                            "[bold yellow]Warning:[/bold yellow] Cannot prepare log file {}: {}",
                            configured,
                            e,
                        )
                    )
                    log_file = None
                if log_file is not None and not _try_init_file_logging(log_file):
                    log_file = None
```

src/conductor/cli/run.py
```diff
@@ -2659,6 +2670,15 @@ async def resume_workflow_async(
...
         config = load_config(resolved_workflow_path)
 
+        if log_file is None:
+            configured = config.workflow.runtime.log_file
+            if configured is not None:
+                if configured.lower() == "auto":
+                    log_file = generate_log_path(resolved_workflow_path.stem)
+                else:
+                    log_file = Path(configured)
+                _try_init_file_logging(log_file)
```
jrob5756, line 2678: Same four issues as the run side. Given these two blocks are identical apart from the path variable, and AGENTS.md names drift between `run_workflow_async` and `resume_workflow_async` as a hazard, I'd lift this into one helper rather than fix it twice.

A shared `resolve_log_file(value, stem)` next to `generate_log_path` would also absorb the two copies in `cli/app.py` (`:663` and `:1157`). Those have already drifted: `app.py:1158` falls back to a `"resume"` stem where this line uses the checkpoint-derived one. Four copies of the `auto` vocabulary means the `expanduser()` and `strip()` fixes have to be applied four times to hold.

One more thing specific to resume. Because the YAML value is read after the checkpoint block, this log misses `Loading checkpoint`, the workflow-hash-changed warning, and `Resuming from agent`, all of which `--log-file` captures. Two spellings of the same setting produce different logs, and the docs present them as equivalent.

src/conductor/config/schema.py
```diff
@@ -3332,6 +3332,9 @@ def _coerce_provider(cls, value: Any) -> Any:
     default_model: str | None = None
     """Default model for agents that don't specify one."""
 
+    log_file: str | None = None
+    """Default log file path, or ``auto`` to generate one automatically."""
```
jrob5756, line 3336: A couple of invariants the type could carry instead of the call site.

An empty or whitespace-only value clears the `is not None` guard downstream and becomes `Path(".")`, which fails as `IsADirectoryError` and prints `Cannot open log file .:` with an errno naming `''`. Neither half points back at this field. `SkillDiscoveryConfig.validate_exclude` a few lines below rejects blank entries for the same reason, and `session_key` at line 1532 already uses the constraint below.

Stripping also fixes `" auto "`, which currently becomes a file named `␣auto␣` rather than matching the sentinel.

The docstring is worth two additions as well: `auto` is matched case-insensitively, and only the top-level workflow's value is read. A sub-workflow setting this gets no log, no warning and no error, which is surprising when `working_dir`, `skills` and `default_model` all apply per sub-workflow. `runtime.checkpoint` has the same root-only scope and says so.

```suggestion
    log_file: Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)] | None = None
    """Default log file path, or ``auto`` (case-insensitive) to generate one automatically.

    Honoured only for the top-level workflow; ignored in a ``type: workflow`` sub-workflow.
    """
```

tests/test_cli/test_logging.py
```diff
@@ -1622,6 +1684,161 @@ def test_section_writes_full_content_to_file_in_silent_mode(self, tmp_path: Path
...
+
+    @staticmethod
+    def _write_workflow(tmp_path: Path, log_path: Path) -> Path:
+        workflow_file = tmp_path / "test.yaml"
+        workflow_yaml = f"""\
+workflow:
+  name: test-workflow
+  entry_point: agent1
+  runtime:
+    log_file: "{log_path.as_posix()}"
```
jrob5756, line 1698: `_write_workflow` always interpolates an explicit path, so `configured.lower() == "auto"` is never taken on run or resume. That's the value in every doc example this PR adds, and it's the branch with the most room to go wrong: dropping the `if`/`else`, or just the `.lower()`, leaves the whole suite green.

`tests/conftest.py` already redirects `tempfile.gettempdir()` to `tmp_path` for every test and `TestGenerateLogPath` relies on it, so this is a small addition. Parameterising the helper to take the raw value would let one test assert a single `*.log` lands under `gettempdir()/conductor/` with the workflow stem in its name.

While you're here, both integration tests assert only `log_path.exists()`. Mode `"w"` creates the file at init regardless, so neither would notice if no log line ever reached it. `test_try_init_with_valid_path` gets this right by reading the content back.

tests/test_cli/test_logging.py
```diff
@@ -395,6 +395,68 @@ def test_generate_log_path_uses_tmpdir(self) -> None:
...
+        from rich.console import Console
+
+        from conductor.cli.run import _try_init_file_logging, close_file_logging
+
+        readonly_dir = tmp_path / "readonly"
+        readonly_dir.mkdir()
+        os.chmod(readonly_dir, 0o444)
+
+        stderr_output = StringIO()
+        mock_console = Console(file=stderr_output, no_color=True, highlight=False)
```
jrob5756, line 447: This patches in a plain `rich.Console`, but production uses `_SilentAwareConsole`, whose `print` is gated on `is_verbose()`. The assertion therefore passes against a console that never suppresses anything.

Under `--silent` the real warning is dropped and the user gets nothing at all: no file, no message, exit 0, and nothing in the JSONL event log either. So this test currently vouches for feedback that doesn't exist in the configuration the docs actually recommend, since `execution.md` suggests pairing file logging with `--silent`.

Patching with `_SilentAwareConsole` and asserting both verbosity states would catch it. Separately, I'd argue a diagnostics subsystem failing to start isn't progress output and shouldn't be silenced by `--silent` at all; routing it to `console` or emitting it as a workflow event would make it durable.

jrob5756 (commented):
Took a careful read through this. The gap you're closing is real, and a few things here are done better than average: the run/resume parity is genuine rather than just claimed, the new warning correctly passes the path and `str(e)` through `styled()` as values instead of into the template, and you updated the bundled skill references alongside the user docs, which people usually forget.

A few things I'd want sorted before it merges.

CI is red right now. `ruff format --check` wants the new warning block collapsed onto one line, and it's the only formatting deviation in the tree. `make format` handles it.

The one I'd fix first is that `Path(configured)` never calls `.expanduser()`. Writing `log_file: ~/logs/run.log` in YAML is a natural thing to do, and because `init_file_logging` runs `mkdir(parents=True)` it succeeds: you get a directory named literally `~` under the process CWD, the log inside it, and no warning, because nothing raised. The CLI flag dodges this only because the shell expands `~` before Python sees it. YAML has no shell, so this field is what makes the case reachable. Every other user-supplied path in the tree expands it, `runtime.working_dir` included.

Close behind: `generate_log_path()` sits outside the try/except, and its `mkdir(parents=True, exist_ok=True)` still raises if `$TMPDIR/conductor` happens to be a file, or on a read-only or full runner. So `log_file: auto` can take down a workflow that would otherwise run fine, with an error panel naming neither `log_file` nor the workflow. Right now an explicit bad path warns and continues while `auto` aborts, and those should match.

Worth deciding separately: `init_file_logging` opens with mode `"w"`. That was fine when the only way in was typing `--log-file` at the prompt. As a persistent YAML default it means `conductor resume` truncates the crash log before writing to it, which is the one moment you most want that file intact. `auto` is safe because of the random suffix; explicit paths aren't, and no doc mentions it.

Finally, `log_file: auto` is the example value in all four doc surfaces you added and no test exercises it on either path. Deleting the `if`/`else` and keeping only `Path(configured)` leaves the suite green.

Rest is inline. None of it is about the idea, which I think is worth having. There's also no `CHANGELOG.md` entry under `[Unreleased]`, which recent feature PRs have all added.

nskun:

src/conductor/cli/run.py
```diff
@@ -1995,6 +2003,15 @@ async def run_workflow_async(
         config = load_config(workflow_path)
         verbose_log_timing("Configuration loaded", time.time() - load_start)
 
+        if log_file is None:
+            configured = config.workflow.runtime.log_file
+            if configured is not None:
+                if configured.lower() == "auto":
+                    log_file = generate_log_path(workflow_path.stem)
```
jrob5756: **BLOCKING**: Still unguarded. `log_file = generate_log_path(workflow_path.stem)` runs with nothing around it, so `mkdir(parents=True, exist_ok=True)` failing — read-only runner, `$TMPDIR/conductor` existing as a file, disk full — still takes the whole run down with it. The explicit-path branch one line below already warns and continues; `auto` should do the same instead of being the one path that's allowed to be fatal.

Wrap both branches in one try, warn through `_verbose_console`, and set `log_file = None` on failure.

src/conductor/cli/run.py
```diff
@@ -2659,6 +2668,15 @@ async def resume_workflow_async(
...
         config = load_config(resolved_workflow_path)
 
+        if log_file is None:
+            configured = config.workflow.runtime.log_file
+            if configured is not None:
+                if configured.lower() == "auto":
+                    log_file = generate_log_path(resolved_workflow_path.stem)
+                else:
+                    log_file = Path(configured)
+                _try_init_file_logging(log_file)
```
jrob5756: **BLOCKING**: This block is still a verbatim copy of the run-path version at 2006-2013, just with a different variable name. Every defect above now exists in two places, and the two `cli/app.py` copies of the same `auto` handling are untouched, so there are four separate spots that all need the same `expanduser`/error-handling fix.

Pull this into one shared helper (`resolve_log_file(value, stem)` next to `generate_log_path` is a reasonable shape) and call it from both async functions.

tests/test_cli/test_logging.py
jrob5756: **BLOCKING**: This still patches a plain `rich.Console`, not `_SilentAwareConsole`. Production gates this warning on `is_verbose()`, so under `--silent` a failed log init produces no file, no message, and exit 0 — and this test can't see that because it isn't using the same console class.

Patch `_SilentAwareConsole` instead and assert the warning in both verbose and silent modes.

src/conductor/cli/run.py
```diff
@@ -1995,6 +2003,15 @@ async def run_workflow_async(
...
         verbose_log_timing("Configuration loaded", time.time() - load_start)
 
+        if log_file is None:
+            configured = config.workflow.runtime.log_file
+            if configured is not None:
+                if configured.lower() == "auto":
+                    log_file = generate_log_path(workflow_path.stem)
+                else:
+                    log_file = Path(configured)
+                _try_init_file_logging(log_file)
```
jrob5756: **RECOMMENDED**: A value with a NUL byte raises `ValueError` from `open()`, and `_try_init_file_logging`'s `except OSError` (lines 145-152) doesn't catch that — it escapes as an unhandled traceback.

Catch `(OSError, ValueError)`, either here or inside `_try_init_file_logging` itself.

jrob5756 (commented):
Good progress since the last pass! I started to review but didn't realize you were working on it still. My apologies for that. Please let me know when its ready for another pass. Cheers!

src/conductor/cli/run.py
jrob5756: **RECOMMENDED**: The bool `_try_init_file_logging` returns is still discarded here. That return value exists specifically so a failed init can clear `log_file`, and neither new call site reads it — the only reason nothing misbehaves today is that the `_file_console is not None` check happens to paper over it.

```suggestion
            if not _try_init_file_logging(log_file):
                log_file = None
```

src/conductor/config/schema.py
```diff
@@ -3332,6 +3332,9 @@ def _coerce_provider(cls, value: Any) -> Any:
     default_model: str | None = None
     """Default model for agents that don't specify one."""
 
+    log_file: str | None = None
```
jrob5756: **RECOMMENDED**: `log_file` still has no constraint, so an empty or blank value becomes `Path(".")` with an error that names neither the field nor the workflow, and `" auto "` falls through to the explicit-path branch instead of matching the sentinel.

```suggestion
    log_file: Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)] | None = None
```

Matches `SkillDiscoveryConfig.validate_exclude` and `session_key` elsewhere in this file.

src/conductor/cli/run.py
```diff
@@ -139,6 +139,22 @@ def init_file_logging(log_path: Path) -> None:
     )
 
 
+def _try_init_file_logging(log_path: Path | None) -> bool:
+    """Initialize file logging, returning whether initialization succeeded."""
+    if log_path is None:
+        return False
+
+    try:
+        init_file_logging(log_path)
```
jrob5756: **RECOMMENDED**: Still routed through `_verbose_console`, so under `--silent` a failure to open the log file is completely invisible — no message, no event in the JSONL log either. A diagnostics subsystem failing to start isn't progress output; it should reach the user regardless of verbosity.

Decide deliberately: print through the always-on `console`, or emit a workflow event, so the failure is durable one way or the other.

jrob5756:
Oh sorry @nskun! I did another round of review and didn't realize you weren't ready yet. My apologies! Let me know when you're ready for another round

PR #474, opened 2026-08-21
feat(runtime): add event_log_dir to configure event log output directory

### LINKED ISSUE #475: feat(runtime): allow event log output directory to be configured in workflow.yaml
# Summary
Event logs are always written to $TMPDIR/conductor/ (typically /tmp/conductor/) with no way to configure the output directory.

# Problem
・Event logs in tmp are lost on reboot or tmpwatch cleanup
・Users must manually copy event logs to a persistent directory after each run
・When a workflow crashes without a workflow_failed event, the only diagnostic is the event log -- which may already be gone
・The event log subscriber is always-on by design (correct -- it's a flight recorder), but the output location should be user-configurable
・Configurable log output directories are standard in self-hosted orchestration tools: Airflow(base_log_folder), Argo workflows(artifactRepository), Temporal, Prefect, Docker(log-driver), Kubernetes(--log-dir)

# Proposal
Add an event_log_dir field to the runtime: block in workflow YAML.
```
workflow:
  runtime:
    event_log_dir: ./logs
```

When omitted, behavior is unchanged($TMPDIR/conductor/). 
Relative paths are resolved against the workflow file's directory.

nskun:
# Summary
Add a new event_log_dir field to RuntimeConfig so users can configure where event logs are written, instead of the hardcoded $TMPDIR/conductor/ .

# Motivation
Event log are always written to $TMPDIR/conductor/ (typically /tmp/conductor/ ). 
On Linux, tmp is cleared on reboot or by tmpwatch, so diagnostic data from long-running workflows can be lost before it is reviewed.
Users currently need awrapper script to copy event logs to a persistent directory after each run.

Configurable log output directories are standard in self-hosted orchestration tools (Airflow base_log_folder, Argo Workflows artifactRepository, Docker log-driver, Kubernetes --log-dir).

# Usage
Relative paths are resolved against the workflow file's directory. 
When omitted, writes to $TMPDIR/conductor as before.
Filename pattern (conductor-{name}-{ts}-{run_id}.events.jsonl) is unchanged -- only the parent directory changes.

Closes #475

### DIFF
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -113,6 +113,20 @@ def generate_log_path(workflow_name: str) -> Path:
     return path
 
 
+def _resolve_event_log_dir(
+    event_log_dir: str | None,
+    workflow_path: Path,
+) -> Path | None:
+    """Resolve an event-log directory relative to its workflow file."""
+    if not event_log_dir:
+        return None
+
+    path = Path(event_log_dir)
+    if not path.is_absolute():
+        path = workflow_path.resolve().parent / path
+    return path.resolve()
+
+
 def init_file_logging(log_path: Path) -> None:
     """Initialize file logging to the given path.
 
@@ -2054,8 +2068,10 @@ async def run_workflow_async(
         # Start JSONL event log subscriber (always-on structured diagnostics)
         from conductor.engine.event_log import EventLogSubscriber
 
-        _eld = config.workflow.runtime.event_log_dir
-        _event_log_dir = Path(_eld).resolve() if _eld else None
+        _event_log_dir = _resolve_event_log_dir(
+            config.workflow.runtime.event_log_dir,
+            workflow_path,
+        )
 
         event_log_subscriber = EventLogSubscriber(
             config.workflow.name,
@@ -2771,8 +2787,10 @@ async def resume_workflow_async(
             # appends to it and reuses run_id; otherwise it generates fresh.
             from conductor.engine.event_log import EventLogSubscriber
 
-            _eld = config.workflow.runtime.event_log_dir
-            _event_log_dir = Path(_eld).resolve() if _eld else None
+            _event_log_dir = _resolve_event_log_dir(
+                config.workflow.runtime.event_log_dir,
+                resolved_workflow_path,
+            )
 
             event_log_subscriber = EventLogSubscriber(
                 config.workflow.name,
--- a/src/conductor/cli/run.py
+++ b/src/conductor/cli/run.py
@@ -2054,7 +2054,13 @@ async def run_workflow_async(
         # Start JSONL event log subscriber (always-on structured diagnostics)
         from conductor.engine.event_log import EventLogSubscriber
 
-        event_log_subscriber = EventLogSubscriber(config.workflow.name)
+        _eld = config.workflow.runtime.event_log_dir
+        _event_log_dir = Path(_eld).resolve() if _eld else None
+
+        event_log_subscriber = EventLogSubscriber(
+            config.workflow.name,
+            event_log_dir=_event_log_dir,
+        )
         emitter.subscribe(event_log_subscriber.on_event)
 
         # Write the Fleet Manager run record (E2): this is the first point
@@ -2765,10 +2771,14 @@ async def resume_workflow_async(
             # appends to it and reuses run_id; otherwise it generates fresh.
             from conductor.engine.event_log import EventLogSubscriber
 
+            _eld = config.workflow.runtime.event_log_dir
+            _event_log_dir = Path(_eld).resolve() if _eld else None
+
             event_log_subscriber = EventLogSubscriber(
                 config.workflow.name,
                 existing_path=existing_log_path,
                 existing_run_id=cp.run_id or None,
+                event_log_dir=_event_log_dir,
             )
             emitter.subscribe(event_log_subscriber.on_event)
 
--- a/src/conductor/config/schema.py
+++ b/src/conductor/config/schema.py
@@ -3452,6 +3452,16 @@ class RuntimeConfig(BaseModel):
     provider declares ``capabilities.working_dir=False``.
     """
 
+    event_log_dir: str | None = None
+    """Directory for event log output.
+
+    When set, event logs are written to this directory instead of
+    ``$TMPDIR/conductor/``. Relative paths are resolved against the
+    workflow file's directory.
+
+    When omitted, behavior is unchanged (writes to ``$TMPDIR/conductor/``).
+    """
+
     skills: list[str] = Field(default_factory=list)
     """Workflow-wide default skills for every provider-backed agent.
 
--- a/src/conductor/engine/event_log.py
+++ b/src/conductor/engine/event_log.py
@@ -79,6 +79,7 @@ class EventLogSubscriber:
         *,
         existing_path: Path | None = None,
         existing_run_id: str | None = None,
+        event_log_dir: Path | None = None,
     ) -> None:
         """Initialise the subscriber.
 
@@ -150,11 +151,12 @@ class EventLogSubscriber:
                 )
             self._run_id = new_run_id()
         ts = time.strftime("%Y%m%d-%H%M%S")
-        self._path = (
-            Path(tempfile.gettempdir())
-            / "conductor"
-            / f"conductor-{workflow_name}-{ts}-{self._run_id}.events.jsonl"
+        base_dir = (
+            event_log_dir
+            if event_log_dir is not None
+            else Path(tempfile.gettempdir()) / "conductor"
         )
+        self._path = base_dir / f"conductor-{workflow_name}-{ts}-{self._run_id}.events.jsonl"
         self._path.parent.mkdir(parents=True, exist_ok=True)
         self._handle = open(self._path, "w", encoding="utf-8")  # noqa: SIM115
 
--- a/tests/test_cli/test_event_log_dir.py
+++ b/tests/test_cli/test_event_log_dir.py
@@ -1,0 +1,77 @@
+"""Tests for event log directory path resolution."""
+
+from pathlib import Path
+
+import pytest
+
+from conductor.cli.run import _resolve_event_log_dir
+
+
+def test_relative_event_log_dir_is_resolved_from_workflow_directory(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Relative paths use the workflow file directory, not process CWD."""
+    workflow_dir = tmp_path / "workflow"
+    workflow_dir.mkdir()
+    workflow_path = workflow_dir / "workflow.yaml"
+    workflow_path.write_text("", encoding="utf-8")
+
+    other_cwd = tmp_path / "elsewhere"
+    other_cwd.mkdir()
+    monkeypatch.chdir(other_cwd)
+
+    result = _resolve_event_log_dir("./logs", workflow_path)
+
+    assert result == (workflow_dir / "logs").resolve()
+    assert result != (other_cwd / "logs").resolve()
+
+
+def test_parent_relative_event_log_dir_is_workflow_relative(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Parent components are evaluated from the workflow directory."""
+    workflow_dir = tmp_path / "project" / "workflows"
+    workflow_dir.mkdir(parents=True)
+    workflow_path = workflow_dir / "workflow.yaml"
+    workflow_path.write_text("", encoding="utf-8")
+
+    other_cwd = tmp_path / "elsewhere"
+    other_cwd.mkdir()
+    monkeypatch.chdir(other_cwd)
+
+    result = _resolve_event_log_dir("../logs", workflow_path)
+
+    assert result == (workflow_dir.parent / "logs").resolve()
+
+
+def test_absolute_event_log_dir_is_unchanged(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Absolute paths do not depend on workflow location or process CWD."""
+    workflow_dir = tmp_path / "workflow"
+    workflow_dir.mkdir()
+    workflow_path = workflow_dir / "workflow.yaml"
+    workflow_path.write_text("", encoding="utf-8")
+
+    other_cwd = tmp_path / "elsewhere"
+    other_cwd.mkdir()
+    monkeypatch.chdir(other_cwd)
+
+    absolute_dir = tmp_path / "absolute-logs"
+
+    assert _resolve_event_log_dir(str(absolute_dir), workflow_path) == absolute_dir.resolve()
+
+
+@pytest.mark.parametrize("value", [None, ""])
+def test_missing_event_log_dir_remains_unset(
+    tmp_path: Path,
+    value: str | None,
+) -> None:
+    """Omitted and empty values retain the default TMPDIR behavior."""
+    workflow_path = tmp_path / "workflow.yaml"
+    workflow_path.write_text("", encoding="utf-8")
+
+    assert _resolve_event_log_dir(value, workflow_path) is None
--- a/tests/test_engine/test_event_log_dir.py
+++ b/tests/test_engine/test_event_log_dir.py
@@ -1,0 +1,69 @@
+"""Tests for configurable event log output directory."""
+
+import tempfile
+import time
+from pathlib import Path
+
+from conductor.engine.event_log import EventLogSubscriber
+from conductor.events import WorkflowEvent
+
+
+def test_runtime_config_event_log_dir():
+    """RuntimeConfig accepts event_log_dir field, defaults to None."""
+    from conductor.config.schema import RuntimeConfig
+
+    assert RuntimeConfig().event_log_dir is None
+    assert RuntimeConfig(event_log_dir="./logs").event_log_dir == "./logs"
+    assert RuntimeConfig(event_log_dir="/var/log/conductor").event_log_dir == ("/var/log/conductor")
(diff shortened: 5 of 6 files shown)


src/conductor/cli/run.py
```diff
@@ -2054,7 +2054,13 @@ async def run_workflow_async(
         # Start JSONL event log subscriber (always-on structured diagnostics)
         from conductor.engine.event_log import EventLogSubscriber
 
-        event_log_subscriber = EventLogSubscriber(config.workflow.name)
+        _eld = config.workflow.runtime.event_log_dir
+        _event_log_dir = Path(_eld).resolve() if _eld else None
```
jrob5756: `Path(...).resolve()` anchors a relative path to the process working directory, but the docstring for this field at `schema.py:3459` says relative paths resolve against the workflow file's directory. I checked: a workflow at `/tmp/eld-test/wf.yaml` with `event_log_dir: ./logs`, invoked from `/tmp/eld-cwd`, wrote to `/tmp/eld-cwd/logs`. The same workflow file puts its logs somewhere different depending on which shell you start it from.

Two related problems on the same line. There is no `expanduser()`, so `event_log_dir: ~/logs` creates a directory named `~` under the current directory (verified). And `resolve()` follows symlinks, which `_resolve_agent_working_dir` at `engine/workflow.py:620` deliberately avoids so symlink aliases stay distinct.

`workflow_path` is already a parameter of both functions, so the workflow directory is available right here.

```suggestion
        configured_dir = config.workflow.runtime.event_log_dir
        _event_log_dir = None
        if configured_dir:
            candidate = Path(configured_dir).expanduser()
            if not candidate.is_absolute():
                candidate = Path(workflow_path).resolve().parent / candidate
            _event_log_dir = Path(os.path.normpath(candidate))
```

Better still, put this in one helper instead of applying it twice. The same two lines appear again at 2774, and the Run/Resume parity rule in AGENTS.md names the event log subscriber specifically. Two copies of a resolution policy is how those two paths drift apart.

src/conductor/config/schema.py
```diff
@@ -3452,6 +3452,16 @@ def _coerce_provider(cls, value: Any) -> Any:
...
 
+    event_log_dir: str | None = None
+    """Directory for event log output.
+
+    When set, event logs are written to this directory instead of
+    ``$TMPDIR/conductor/``. Relative paths are resolved against the
+    workflow file's directory.
+
+    When omitted, behavior is unchanged (writes to ``$TMPDIR/conductor/``).
+    """
```
jrob5756, line 3463: The sentence about relative paths does not match `cli/run.py:2058`, which resolves against the process working directory. See my note there. Whichever way that goes, this text has to agree with it. Worth noting `spill_dir` a few hundred lines up has the same mechanics and documents the opposite rule, so right now two adjacent fields contradict each other.

There is also no validator. `event_log_dir: "   "` is truthy, so it creates a directory named with three spaces. `""` happens to behave, but only because `run.py` tests truthiness while `event_log.py:155` tests `is not None`. That split is what `_normalize_spill_dir` at line 3123 exists to prevent, and its docstring says so directly.

```python
@field_validator("event_log_dir")
@classmethod
def _normalize_event_log_dir(cls, v: str | None) -> str | None:
    """Normalize an empty or whitespace-only value to None.

    cli/run.py tests this field for truthiness while EventLogSubscriber
    tests it for `is not None`, so without one normalization point an
    empty string means "default directory" to one consumer and
    "process cwd" to the other.
    """
    if v is None:
        return None
    return v.strip() or None
```

Two things the docstring should also say: a resumed run keeps writing to the log recorded in its checkpoint and ignores this setting, and a log written outside `$TMPDIR/conductor/` is skipped by the retention sweep and by `conductor fleet` History.

`grep event_log_dir docs/` currently comes back empty, and `docs/cli-reference.md:333` still states the log lives under `$TMPDIR/conductor/`.

src/conductor/engine/event_log.py
```diff
@@ -79,6 +79,7 @@ def __init__(
         *,
         existing_path: Path | None = None,
         existing_run_id: str | None = None,
+        event_log_dir: Path | None = None,
```
jrob5756, line 82: The `Args:` block below documents the other three parameters but not this one. The precedence rule is missing too: `event_log_dir` is ignored when the subscriber appends to `existing_path`. That behaviour is correct, since a resumed run should keep one continuous log, but a test is currently the only place it is recorded.

Something like:

```
    event_log_dir: Base directory for a freshly created log, replacing
        the default ``$TMPDIR/conductor/``. Ignored when the
        ``existing_path`` append branch is taken, so a resumed run keeps
        writing to the log its checkpoint points at. Callers resolve
        relative paths and ``~`` before passing a value here.
```

The case this leaves confusing: set `event_log_dir` after an initial run, then resume. The new directory stays empty, the log keeps growing in `$TMPDIR/conductor/` where the retention sweep will eventually remove it, and nothing says why. A `logger.info` on that branch would cover it. The sibling branch already logs when it cannot append, so the asymmetry stands out.

The module docstring at lines 3-4 also still says the log goes to `$TMPDIR/conductor/` without qualification.

src/conductor/engine/event_log.py
```diff
@@ -150,11 +151,12 @@ def __init__(
...
         ts = time.strftime("%Y%m%d-%H%M%S")
-        self._path = (
-            Path(tempfile.gettempdir())
-            / "conductor"
-            / f"conductor-{workflow_name}-{ts}-{self._run_id}.events.jsonl"
+        base_dir = (
+            event_log_dir
+            if event_log_dir is not None
+            else Path(tempfile.gettempdir()) / "conductor"
         )
```
jrob5756, line 158: Minor: this collapses to one line. `PurePath` defines neither `__bool__` nor `__len__`, so every `Path` is truthy and `None` is the only falsy input, which makes `or` equivalent to the `is not None` test here.

```suggestion
        base_dir = event_log_dir or Path(tempfile.gettempdir()) / "conductor"
```

That equivalence depends on the parameter staying `Path | None`. If it is ever widened to accept `str`, an empty string becomes falsy and `or` would quietly swallow it.

src/conductor/engine/event_log.py
```diff
@@ -150,11 +151,12 @@ def __init__(
...
+        self._path = base_dir / f"conductor-{workflow_name}-{ts}-{self._run_id}.events.jsonl"
```
jrob5756, line 159: Worth deciding explicitly what should happen to a run whose log lands outside `$TMPDIR/conductor/`. Five places derive that directory independently and none of them consult the new setting:

- `fleet/retention.py:76` (`event_log_root`), so the log is never pruned and the directory grows without bound
- `fleet/history.py:476`, so the run never appears in `conductor fleet` History
- `fleet/resume.py`, which joins checkpoints against History entries, so the Resume action is never offered for it
- `fleet/records.py:889`, the event-log recovery lookup
- `cli/bg_runner.py:1058`, which still writes `.bg.stderr.log` and `.bg.stdout.log` to `$TMPDIR/conductor/` while the events log moves elsewhere

I confirmed the History case: one entry from `build_history_entries()` with the default location, zero with `event_log_dir` set.

That last item has a user-visible consequence of its own. `cli/app.py:219` locates the capture logs with `Path(record.event_log_path).parent`, so `conductor status --json` reports `stderr_log: null` for a background run whose capture logs are sitting one directory over. It also breaks the `--web-bg` debugging procedure in AGENTS.md, which depends on the three artifacts sharing a directory, and retention's guarantee that they are kept or removed together.

Either route these readers through a shared accessor, or say plainly in the field docstring that a custom directory opts the run out of History, retention, and background log correlation. The empty History screen is the part I would not leave silent. AGENTS.md is explicit that showing an empty state for a directory that was never read is a claim of absence that reads like success.

src/conductor/engine/event_log.py
```diff
@@ -150,11 +151,12 @@ def __init__(
...
         self._path.parent.mkdir(parents=True, exist_ok=True)
         self._handle = open(self._path, "w", encoding="utf-8")  # noqa: SIM115
```
jrob5756, line 161: Now that this path can come from user config, these two lines can raise. I checked both cases: pointing `event_log_dir` at an existing file gives `FileExistsError`, and an unwritable directory gives `PermissionError`. Neither `run_workflow_async` nor `resume_workflow_async` wraps the call in an `except`, so it reaches `print_error` and the user gets a panel titled `PermissionError` containing `[Errno 13] Permission denied: '/sys/conductor-logs'`. The field name never appears, and the workflow exits 1 before running a single step.

So a typo in a diagnostics setting now stops work that would otherwise have succeeded, which is the opposite of what an always-on flight recorder should do.

Everything comparable in this codebase degrades instead. The `existing_path` branch at lines 112-125 catches `OSError` and creates a fresh log with a warning. `--log-file` warns and continues (`run.py:1962`). The dashboard warns and continues (`run.py:2016`). `spill_dir` in `mcp/manager.py:450` is documented as best effort that must never raise.

Catching `OSError` here and falling back to `$TMPDIR/conductor/` with a warning naming the field would match the rest, and it keeps the log alive rather than trading it for a dead run. If strict failure is preferred instead, it should be a config error raised by `conductor validate`, which currently passes clean on all of these values.

tests/test_engine/test_event_log_dir.py
```diff
@@ -0,0 +1,69 @@
...
+
+import tempfile
+import time
+from pathlib import Path
+
+from conductor.engine.event_log import EventLogSubscriber
+from conductor.events import WorkflowEvent
+
+
+def test_runtime_config_event_log_dir():
```
jrob5756, line 11: Every test in this file hands the subscriber a `Path` that is already absolute, so the piece this PR actually adds, the conversion from the YAML string in `cli/run.py`, is never exercised. That is why the relative-path and `~` behaviour went unnoticed.

Cases worth adding: a relative value where the process runs somewhere other than the workflow file's directory, `~`, whitespace, a path that already exists as a file, and the resume wiring. `tests/test_fleet/test_run_record_wiring.py` is a natural home for the last one, since it already drives both async helpers and can assert where `record.event_log_path` lands.

Two smaller notes. This test exercises `RuntimeConfig` rather than the engine, so `tests/test_config/` fits better. And `test_default_writes_to_tmpdir` largely repeats `test_event_log.py:16`, whose `TestEventLogSubscriber` class already owns this surface, so these might be more discoverable as methods there.

Line 17 also has a stray pair of parens around the string literal.

tests/test_engine/test_event_log_dir.py
```diff
@@ -0,0 +1,69 @@
...
+                timestamp=time.time(),
+                data={},
+            )
+        )
+        sub.close()
+
+        assert sub.path.read_text().strip()
+    finally:
+        if not sub._handle.closed:
+            sub.close()
```
jrob5756, line 51: `close()` is already idempotent. `event_log.py:194` guards with `self._handle is not None and not self._handle.closed`, and `test_event_log.py:73` already pins that with a double close. This guard re-implements the check and reaches into a private attribute to do it.

```suggestion
    finally:
        sub.close()
```

nskun:
The Windows CI failure is in an unrelated, timing-sensitive TUI timer test:

tests/test_fleet/test_tui_runs.py::TestRunsScreenPausesWhileNotOnTop::test_the_animation_timer_is_paused_and_resumed

This PR does not modify the TUI code, and the same Windows check passes on the base commit. Could a maintainer please rerun the failed job?

nskun:
I pushed an empty commit to retrigger CI.
The new workflow run is awaiting approval.
Could a maintainer approve and run it?

nskun:
I also fixed relative path resolution so runtime.event_log_dir is now resolved against the workflow file's directory, as documented, for both run and resume.

Added regression coverage for cases where the process CWD differs from the workflow directory.

jrob5756 (commented):
Reviewed commit 13cba6c in a worktree with a set of focused agents. Lint and format pass, `ty` adds no new diagnostics, and the four new tests pass.

The problem this solves is real. The event log is the primary post-mortem artifact for a crashed run and `$TMPDIR` gets swept, so making the location configurable is worth doing. My comments are about how the configured value is resolved, and about the parts of the system that still assume the old location.

One finding has no line to attach to, because the file is not in this diff:

`tests/test_fleet/test_run_record_wiring.py` builds its config with `_mock_config()`, which returns a `MagicMock`. `MagicMock` implements `__fspath__`, so `_eld` is truthy and `Path(_eld).resolve()` produces a path inside the repository, which `event_log.py:160` then creates. From a clean tree:

```
$ pytest tests/test_fleet/test_run_record_wiring.py
10 passed
$ git status --porcelain
?? MagicMock/
$ find MagicMock -type f | head -1
MagicMock/mock.workflow.runtime.event_log_dir/137076342320704/conductor-wiring-test-...events.jsonl
```

Twelve test files drive `run_workflow_async` or `resume_workflow_async`, so it is not confined to one suite. It gets past both autouse guards in `tests/conftest.py` because they patch `$TMPDIR` and `CONDUCTOR_HOME`, while this path anchors on the process working directory. It also weakens an existing assertion: `test_run_record_wiring.py:202` used to prove the default `$TMPDIR` path worked, and now only proves that a `MagicMock/` path exists. Setting `event_log_dir = None` in `_mock_config()` clears it.

Two behaviours I verified by running a workflow from a directory other than the one holding the YAML: `./logs` resolved against the shell's working directory rather than the workflow file's directory, and `~/conductor-logs` created a directory literally named `~`. On the fleet side, a completed run appears in `build_history_entries()` with the default location and gives zero entries with `event_log_dir` set.

Nothing here is hard to fix, and most of it collapses into one shared resolution helper.

nskun:

jrob5756 (commented):
Good progress since the last pass! I started to review but didn't realize you were working on it still. My apologies for that. Please let me know when its ready for another pass. Cheers!

src/conductor/config/schema.py
```diff
@@ -3452,6 +3452,16 @@ def _coerce_provider(cls, value: Any) -> Any:
     provider declares ``capabilities.working_dir=False``.
     """
 
+    event_log_dir: str | None = None
```
jrob5756: **RECOMMENDED**: No validator normalising a whitespace-only value

`event_log_dir: "   "` is truthy, so it creates a directory named with three spaces; `_normalize_spill_dir` a few hundred lines up exists to prevent exactly this.

What the code shows now: there's no `field_validator` for `event_log_dir` here. Checked against current code: `RuntimeConfig(event_log_dir='   ').event_log_dir` is `'   '`, and `_resolve_event_log_dir('   ', wf)` returns a path ending in three spaces under the workflow directory. The empty-string case is only safe by accident, since the helper at `run.py:121` tests truthiness while `event_log.py:156` tests `is not None`.

What's still needed: add a `_normalize_event_log_dir` validator that strips and folds an empty result to `None`, so both consumers see one value.

src/conductor/engine/event_log.py
jrob5756: **RECOMMENDED**: New parameter is undocumented, and the `existing_path` precedence is unrecorded

The `Args:` block documents the other three parameters but not `event_log_dir`, and nothing states that the setting is ignored when the subscriber appends to `existing_path`. The module docstring also still says the log goes to `$TMPDIR/conductor/` unqualified.

What the code shows now: lines 85-96 still list only `workflow_name`, `existing_path`, and `existing_run_id`. The module docstring at lines 3-4 is unchanged. The append branch at lines 110-125 still returns early with no log line explaining that a configured directory was bypassed, while its sibling failure path does log.

What's still needed: document `event_log_dir` in the `Args:` block including the append-branch precedence, qualify the module docstring, and consider a `logger.info` on the branch that ignores the setting.

tests/test_cli/test_event_log_dir.py
```diff
@@ -0,0 +1,77 @@
...
+
+from pathlib import Path
+
+import pytest
+
+from conductor.cli.run import _resolve_event_log_dir
+
+
+def test_relative_event_log_dir_is_resolved_from_workflow_directory(
+    tmp_path: Path,
```
jrob5756: **RECOMMENDED** (partially addressed): String-to-path conversion is now tested; several suggested cases are not

What's landed: this file now covers relative, `../`, absolute, `None`, and `""` with the process cwd deliberately elsewhere, and `tests/test_fleet/test_run_record_wiring.py` asserts the resolved directory on both the run and resume paths.

What's still needed: `~`, a whitespace-only value, and a configured path that already exists as a file — the three cases that still misbehave per `b1` and `r2` above. Add those once the corresponding behaviour is decided.

tests/test_engine/test_event_log_dir.py
```diff
@@ -0,0 +1,69 @@
...
+from conductor.events import WorkflowEvent
+
+
+def test_runtime_config_event_log_dir():
+    """RuntimeConfig accepts event_log_dir field, defaults to None."""
+    from conductor.config.schema import RuntimeConfig
+
+    assert RuntimeConfig().event_log_dir is None
+    assert RuntimeConfig(event_log_dir="./logs").event_log_dir == "./logs"
+    assert RuntimeConfig(event_log_dir="/var/log/conductor").event_log_dir == ("/var/log/conductor")
```
jrob5756: **RECOMMENDED**: Test placement and a stray pair of parens

`test_runtime_config_event_log_dir` exercises `RuntimeConfig` rather than the engine, so it fits `tests/test_config/` better; `test_default_writes_to_tmpdir` largely repeats the existing `TestEventLogSubscriber` coverage, and this line has redundant parens around a string literal.

What the code shows now: this file is unchanged — the schema test is still at lines 11-17 with `== ("/var/log/conductor")` on line 17, and `test_default_writes_to_tmpdir` still sits outside `test_event_log.py`'s existing class.

What's still needed: cosmetic. Drop the parens and consider folding the subscriber tests into `TestEventLogSubscriber`.

```suggestion
        assert config.event_log_dir == "/var/log/conductor"
```

PR #480, opened 2026-08-23
refactor(engine): remove dead workflow lifecycle hooks (#476)

### LINKED ISSUE #476: Remove dead workflow lifecycle hooks (on_start / on_complete / on_error)
## Summary

The workflow schema documents and parses a `hooks:` block with `on_start`, `on_complete`, and `on_error` lifecycle hooks (see `docs/workflow-syntax.md`, "Hooks"), but the feature is effectively **dead code**: hook templates are rendered and then discarded. Nothing observes the result.

## Current state

**Schema & docs** — `HooksConfig` is defined in `src/conductor/config/schema.py` (~line 597) with three optional template strings, and documented as:

```yaml
hooks:
    on_start: "{{ template }}"      # Optional: Expression evaluated on start
    on_complete: "{{ template }}"   # Optional: Expression evaluated on success
    on_error: "{{ template }}"      # Optional: Expression evaluated on error
```

**Engine** — `WorkflowEngine._execute_hook` (`src/conductor/engine/workflow.py`, ~line 5477) is called from ~12 sites (`run()`, `resume()`, the `on_complete` paths in `_execute_loop`, and `on_error`). It:

1. Renders the Jinja2 template against the workflow context (plus `result` / `error` for the respective hooks).
2. Returns a `LifecycleHookResult(executed=True, result=rendered)`.

**The gap** — every call site ignores the returned value. The rendered string is not:

- emitted as an event (no `hook_executed` / similar in the event vocabulary),
- written to the JSONL event log,
- shown on the dashboard or console,
- stored in the workflow context,
- used for any side effect (no notification, no script execution, nothing).

Hook *render* errors are swallowed by design ("should not fail the workflow"), which is reasonable — but combined with the discarded result it means a hook is completely unobservable, including when it fails. There is also no test coverage asserting any observable behavior, because there is none to assert.

## History

The hooks block was introduced in the very first commit of the repository and has never been wired to anything. No open issue tracks plans for it.

## Proposal

Remove the dead code:

1. Drop `HooksConfig` and `WorkflowDef.hooks` from `config/schema.py` (plus re-exports in `config/__init__.py`).
2. Drop `WorkflowEngine._execute_hook`, `LifecycleHookResult`, and all call sites in `engine/workflow.py`.
3. Remove the Hooks section from `docs/workflow-syntax.md`.

Removing unused, unobservable surface area is good hygiene: it stops misleading workflow authors into believing the block does something, and shrinks the schema the validator and docs must maintain.

## If hooks are needed later

Lifecycle hooks are plausibly useful one day (notifications on completion, error alerting, audit trails). But when a concrete requirement arrives, the right syntax will likely differ from "render a template and throw it away" — e.g. emitting an event, invoking a `type: script` step, or calling a webhook. It is cheaper to design that syntax against a real requirement than to preserve a placeholder whose semantics were never defined.

## Acceptance criteria

- [ ] `hooks:` removed from the workflow schema (parsing a workflow that declares it produces a clear validation error).
- [ ] `_execute_hook` / `LifecycleHookResult` and all call sites removed from the engine.
- [ ] Hooks section removed from `docs/workflow-syntax.md`.
- [ ] CHANGELOG entry noting the removal.

RounakKumarAgarwal:
Closes #476.

## What

The `hooks:` block (`on_start` / `on_complete` / `on_error`) was documented and parsed, but the feature was dead code: hook templates were rendered and the result discarded. Nothing observed it — no event, no log entry, no dashboard/console output, no context write, no side effect — so a hook was completely unobservable, including when it failed.

## Changes

- **Schema** — removed `HooksConfig` and `WorkflowDef.hooks` from `config/schema.py`, and the re-exports from `config/__init__.py`.
- **Engine** — removed `_execute_hook`, `LifecycleHookResult`, and all call sites from `engine/workflow.py`, plus now-stale docstring references.
- **Docs** — removed the Hooks section (and TOC entry / schema-overview snippet) from `docs/workflow-syntax.md`.
- **Tests** — removed the hook-specific tests and fixture (`tests/fixtures/valid_full.yaml`); added `TestHooksRemoved` asserting a workflow declaring `hooks:` now fails validation with a clear error.
- **CHANGELOG** — noted the removal.

## Acceptance criteria

- [x] `hooks:` removed from the schema — declaring it produces a clear validation error (`extra_forbidden`).
- [x] `_execute_hook` / `LifecycleHookResult` and all call sites removed from the engine.
- [x] Hooks section removed from `docs/workflow-syntax.md`.
- [x] CHANGELOG entry added.

## Notes

If lifecycle hooks are wanted later, the right syntax should be designed against a concrete requirement (emitting an event, invoking a `type: script` step, or calling a webhook) rather than restoring a placeholder that renders a template and throws it away.

Because `hooks:` was documented and previously parsed (silently doing nothing), this is technically a breaking change for any workflow that declares it — those will now fail validation, which is the intended behavior per the issue.

Verified locally: `ruff check`/`format` clean, affected test suites green. The remaining `ty` diagnostics are pre-existing Windows-only platform issues (`termios`/`signal.SIGKILL`/etc.) present on a clean checkout, unrelated to this change.


RounakKumarAgarwal:

src/conductor/engine/workflow.py
```diff
@@ -5379,16 +5340,10 @@ async def _execute_loop(self, current_agent_name: str) -> dict[str, Any]:
...
-            # NOTE: we deliberately do NOT invoke ``on_error`` lifecycle
-            # hooks here. Their declared signature accepts ``Exception |
-            # None``, so user-defined hooks may assume ``e.args`` /
-            # ``traceback`` semantics that don't hold for ``SystemExit``,
-            # and surfacing a process-exit signal to them would change
-            # their contract. Users who need hook-like notification for
-            # ``BaseException`` failures should subscribe to the
-            # ``workflow_failed`` event and check ``is_base_exception``.
-            #
+            # Windows startup crash (#116) leaves no ``workflow_failed``
```
jrob5756: **BLOCKING**: Removing the `NOTE:` paragraph about why `on_error` hooks were skipped for `BaseException` corrupted the surrounding comment. Lines 5341-5342 are duplicated verbatim at 5343-5344, and 5345-5346 repeat at 5347-5348. Line 5346 cuts off mid-clause ("...are wrapped in their own") and then restarts the same sentence — the paragraph no longer reads as prose, and both `#`-only separators got eaten in the process, so the two surviving paragraphs run together.

No executable line moved — I diffed the handler against `origin/main` and both `try/except` guards, both `noqa: BLE001` stderr warnings, and the terminal `raise` are all intact. Ruff and the tests pass, so nothing catches this automatically.

Still worth blocking on: this arm exists specifically to stop failures from disappearing silently (issue #116), and the comment is the only thing telling a future maintainer why `_emit` and `_save_checkpoint_on_failure` each need their own wrapper — so neither one can swallow the original exception. Leave it garbled and someone will eventually "simplify" it back into exception-masking.

One more thing worth keeping from the deleted NOTE: "consumers that need notification for `BaseException` failures should subscribe to `workflow_failed` and check `is_base_exception`." That flag gets set about 15 lines below and now has no documentation anywhere explaining what it's for.

```suggestion
        except BaseException as e:
            # Catch-all for exception classes that don't derive from
            # ``Exception`` (e.g. ``SystemExit`` raised by a misbehaving
            # library, fatal runtime errors). Without this arm the silent
            # Windows startup crash (#116) leaves no ``workflow_failed``
            # event in the JSONL log, making the failure invisible.
            #
            # Consumers that need notification for ``BaseException``
            # failures should subscribe to ``workflow_failed`` and check
            # the ``is_base_exception`` flag set below.
            #
            # The diagnostic side effects below (``_emit``,
            # ``_save_checkpoint_on_failure``) are wrapped in their own
            # ``try/except`` blocks: if either of them raised here, the
            # raised side-effect exception would replace the original
            # ``BaseException`` on its way out of this handler, masking
            # the exact crash we're trying to surface. Print a warning to
            # stderr instead so the diagnostic regression is visible, but
            # the ``raise`` at the end always re-raises the *original* ``e``.
```

jrob5756 (commented):
Two blocking issues here, both real: a mangled comment that undoes the readability this refactor was supposed to deliver, and a bundled skill that still teaches agents to write a `hooks:` block that now hard-fails validation. Once those are fixed this is fine to merge — the three recommended items are cleanup, not blockers.

**Blocking:**
- `src/conductor/engine/workflow.py:5343` — duplicated lines and a broken sentence in the `except BaseException` comment (see inline).
- No single line — the `conductor` skill (`SKILL.md`, `authoring.md`, `yaml-schema.md`) still documents `hooks:`, and a workflow using it now fails to load instead of just silently no-opping like before this PR.

### b2 — BLOCKING: the bundled `conductor` skill still teaches agents to write `hooks:`, which now hard-fails validation

`docs/workflow-syntax.md` got cleaned up, but the skill that ships with the wheel didn't. `plugins/conductor/skills/conductor/skills/conductor/SKILL.md:111`, `references/authoring.md:51-54`, and `references/yaml-schema.md:133-138` all still document the removed `hooks:` block — and this is the file `skills/registry.py` loads whenever a workflow declares `skills: [conductor]`.

This is worse than typical doc drift because of how the removal was implemented. `WorkflowDef` uses `extra="forbid"`, so `hooks:` doesn't degrade to a no-op anymore — it aborts the load. I confirmed it: a workflow with `workflow.hooks:` fails `conductor validate` with exit code 1 and `workflow.hooks: Extra inputs are not permitted`. So an agent following this skill will generate a workflow that can't run at all. Before this PR, the same generated YAML at least executed.

I checked that these are genuinely the removed workflow-level feature and not the unrelated plugin `hooks/` directories (`plugins/manifest.py`, `config/validator.py`, `README.md:229`, `examples/plugins.yaml:39` are correctly untouched) — these three references are specifically `workflow:`-scoped `hooks:` keys with `on_start`/`on_complete`/`on_error` children.

Worth flagging as a process gap too: `docs/workflow-syntax.md` and these skill references are near-duplicate YAML references, and only the first got updated here. That's exactly how this slipped through, and it'll happen again on the next syntax change unless there's a check for it. A `grep -rn 'hooks:' plugins/ docs/ examples/` before merge would have caught it.

**Suggested fix:** Delete the table row at `SKILL.md:111`; delete lines 51-54 of `authoring.md` (and the blank line that's now orphaned); delete lines 133-138 of `yaml-schema.md` including the `# Lifecycle hooks` comment, leaving one blank line between the `cost:` block and the `# Arbitrary metadata for downstream tooling` comment.

### r1 — RECOMMENDED: stale "on_error hooks" prose survives in three places the sweep missed

The parallel wording got scrubbed correctly at `engine/workflow.py:2109`, `engine/workflow.py:2136`, and `docs/workflow-syntax.md:1158`, but three mirrors of the same sentences didn't:

- `src/conductor/exceptions.py:664` — the `SubworkflowTerminatedError` docstring still says the attributes are carried "...so on_error hooks, debugging surfaces, and the CLI can inspect it without walking `__cause__`." This one's worth more attention than a normal stale comment, since it's the stated reason `terminated_output`/`terminated_reason`/`terminated_by` live on the exception at all. Point at a consumer that no longer exists and someone will eventually read this as dead code and strip real diagnostic payload off a sub-workflow failure.
- `AGENTS.md:216` — the Terminate steps bullet still says "; runs `on_complete` hook" and ", runs `on_error` hook." This file gets loaded into agent context on basically every task in this repo, so the stale claim has wide reach.
- `plugins/conductor/skills/conductor/references/authoring.md:657, 658, 660` — "Runs the `on_complete` hook.", "Runs the `on_error` hook.", and "...for `on_error` hooks and debugging surfaces." Line 660 is the exact sentence already fixed at `docs/workflow-syntax.md:1158` — just not here.

All three describe behavior that doesn't happen anymore.

**Suggested fix:** Match the wording already used in `docs/workflow-syntax.md:1158`. In `exceptions.py:664`: "the wrapper so debugging surfaces and the CLI can inspect it without walking `__cause__`." In `AGENTS.md:216`: drop "; runs `on_complete` hook" and ", runs `on_error` hook". In `authoring.md`: drop the "Runs the ... hook." sentences at 657 and 658, and on 660 change "for `on_error` hooks and debugging surfaces" to "for debugging surfaces".

CHANGELOG.md
```diff
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+### Removed
+
+- **Dead workflow lifecycle hooks (`on_start` / `on_complete` / `on_error`)**
+  (#476). The `hooks:` block was parsed and its templates rendered, but the
+  result was discarded — never emitted as an event, logged, shown, stored, or
+  used for any side effect — so the feature was entirely unobservable,
+  including when a hook failed. `HooksConfig`, `WorkflowDef.hooks`, the engine's
+  `_execute_hook` / `LifecycleHookResult` and all call sites, and the Hooks
+  section of `docs/workflow-syntax.md` have been removed. A workflow that still
+  declares `hooks:` now fails validation with a clear error rather than
```
jrob5756, line 19: **RECOMMENDED**: "Fails validation with a clear error" oversells what actually shows up. The block genuinely isn't ignored anymore — `conductor validate` exits 1 and names the path — but what a user actually sees is raw Pydantic:

```
1 validation error for WorkflowConfig
workflow.hooks
  Extra inputs are not permitted [type=extra_forbidden, ...]
💡 Suggestion: Check the workflow configuration against the schema.
   Ensure all required fields are present and have valid values.
```

No mention of removal, no version, no pointer to #476. I grepped `src/conductor/config/` for `extra_forbidden`, `Extra inputs`, and any removed-field mapping — nothing. Someone upgrading a working `hooks:` block gets a message indistinguishable from a typo, and the generic "Suggestion" actively points them the wrong way — they'll go looking for the right `hooks:` spelling in the docs and find the section already deleted.

Two related test gaps, since fixing them depends on which way you go here:

1. `TestHooksRemoved::test_hooks_field_is_rejected` calls the `WorkflowDef` constructor directly with an unknown kwarg, which no user actually does. It wouldn't catch a regression at the loader or CLI layer. The class docstring frames this in YAML terms, but the test doesn't exercise YAML at all.
2. `assert "extra_forbidden" in message` is coupled to a Pydantic v2 internal error-type slug. It adds nothing over `"hooks" in message`, and it'll actively break if you add the friendlier validator below — i.e., it fails on a change that makes the behavior better.

**Suggested fix:** Either make the sentence true with a removed-field check:

```python
_REMOVED_FIELDS = {
    "hooks": (
        "`workflow.hooks:` (on_start/on_complete/on_error) was removed (#476) — "
        "the templates were rendered and discarded, so the block never had any "
        "effect. Delete it. For completion/failure notification, subscribe to the "
        "`workflow_completed` / `workflow_failed` events in the JSONL event log."
    ),
}

@model_validator(mode="before")
@classmethod
def _reject_removed_fields(cls, data: Any) -> Any:
    if isinstance(data, dict):
        for key, guidance in _REMOVED_FIELDS.items():
            if key in data:
                raise ValueError(guidance)
    return data
```

or soften the CHANGELOG line to "now fails validation with a Pydantic `extra_forbidden` error on `workflow.hooks` rather than silently ignoring the block."

Either way, drop the `extra_forbidden` assertion for the user-visible text, and add a CLI-level test:

```python
def test_validate_rejects_removed_hooks_block(self, fixtures_dir: Path) -> None:
    result = runner.invoke(app, ["validate", str(fixtures_dir / "invalid_hooks.yaml")])
    assert result.exit_code == 1
    assert "hooks" in result.output
```

CHANGELOG.md
```diff
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  silently ignoring the block. Lifecycle hooks may return later, but the right
+  syntax will be designed against a concrete requirement (emitting an event,
+  invoking a `type: script` step, or calling a webhook) rather than rendering a
+  template and throwing it away.
```
jrob5756, line 23: **RECOMMENDED**: Two small formatting slips from this insertion (confirmed with `cat -A`):

1. No blank line between the end of the Removed bullet ("template and throwing it away.") and the `### Added` heading — they run into each other. Every other `###` heading in the file has a blank line before it.
2. Lines 9-10 now have a double blank line between `## [Unreleased]` and `### Removed`, where every other release heading in the file has exactly one.

Separately, and much more weakly: `### Removed` comes before `### Added`, which departs from the ordering the file's own header cites (Keep a Changelog). I checked past entries and it's genuinely inconsistent already (0.1.28 goes Added → Removed → Fixed, 0.1.29 goes Security → Fixed → Added), so this is a judgment call rather than a violation — just flagging it so the placement is a choice rather than an accident.

```suggestion
template and throwing it away.

### Added
```

RounakKumarAgarwal:
Thanks for the thorough review — all five items addressed in 73e3b9d.

Blocking:

B1 (garbled except BaseException comment) — restored the comment using your suggested text, including the is_base_exception paragraph documenting what the flag is for. The duplicated lines and cut-off clause are gone; no executable lines were touched.
B2 (bundled skill still teaches hooks:) — removed the hooks: block from all three skill files: the table row in SKILL.md, the YAML block in references/authoring.md, and the block in references/yaml-schema.md. An agent following the skill will no longer generate a workflow that hard-fails.

Recommended:

R1 (stale "on_error hooks" prose) — reworded exceptions.py (the SubworkflowTerminatedError docstring), AGENTS.md:216, and authoring.md:657-660 to match the wording already used in docs/workflow-syntax.md. I also found and fixed one more the sweep missed: a docstring in tests/test_engine/test_subworkflow.py.
R2 (error message oversells) — went with the validator approach: added a _reject_removed_fields model validator on WorkflowDef that raises a targeted message naming #476 and pointing at the workflow_completed/workflow_failed events. Reworked the tests accordingly — dropped the brittle extra_forbidden assertion, and now assert the removal message at three layers: schema constructor (test_schema.py), YAML loader (test_loader.py), and CLI conductor validate (test_validate.py, with a new invalid_hooks.yaml fixture).
R3 (CHANGELOG spacing) — fixed the double blank line after ## [Unreleased] and the missing blank line before ### Added. Left the ordering as-is since, as you noted, the file is already inconsistent.

On the process gap: agreed a grep -rn 'hooks:' plugins/ docs/ examples/ would've caught B2. I ran that sweep this round and it's clean apart from the intended references (the validator message and the CHANGELOG entry).

RounakKumarAgarwal:
CI note: the one failing check (Test (Python 3.13, ubuntu-latest)) is a flaky Fleet TUI test, not a regression from this PR.

It failed on tests/test_fleet/test_tui_runs.py::TestRunsScreenPausesWhileNotOnTop::test_the_animation_timer_is_paused_and_resumed — a timing assertion on the splash animation timer. An earlier run of this same branch failed instead on a different set of Fleet TUI tests (test_tui_theme.py::TestShortenHome and test_tui_actions.py::TestDirectoryPickerModal, on the Windows job). Different platform, different tests, all under tests/test_fleet/ — the inconsistency across runs points to pre-existing flakiness/platform-fragility in the Fleet TUI suite rather than anything this change introduced.

For reference, this PR's changes (git diff --name-only upstream/main...HEAD) don't touch fleet/ or tui/ at all — they're limited to config/schema, engine, exceptions, the bundled conductor skill docs, and the corresponding tests. Every test this PR adds or modifies passes on all jobs. A re-run of the failed job should likely go green; happy to open a separate issue for the flaky Fleet tests if that's useful.

jrob5756 (approved):

RounakKumarAgarwal:
> LGTM, thanks for contributing!

Thank you! I’m really interested in contributing more to the project. How can I connect with you to collaborate on issues and work on such repositories? I’d love to learn from the community and contribute more. Looking forward to hearing from you!

PR #482, opened 2026-08-24
feat(telemetry): add OpenTelemetry tracing for workflow execution

### LINKED ISSUE #481: Add opt-in OpenTelemetry tracing for workflow execution
## Summary

Add opt-in OpenTelemetry tracing to Conductor so workflow authors can inspect orchestration behavior, state transitions, step execution, failures, latency, token usage, cost, and tool calls without reconstructing a run from CLI output.

With `runtime.telemetry.enabled: true` and a standard `OTEL_EXPORTER_OTLP_ENDPOINT`, an OTLP-compatible backend such as Jaeger, Tempo, or Grafana should receive an orchestration span tree shaped like:

```text
workflow
├── agent / validator / script / set / wait / gate
├── parallel / for_each / subworkflow
│   └── member or item spans
└── tool calls
```

For the Pydantic AI-backed `claude` and `openai` providers, also enable native Pydantic AI instrumentation. Those native LLM traces should use the same `TracerProvider` and be correlated with the orchestration trace through `gen_ai.conversation.id = run_id`.

## Motivation

The current CLI and event-log surfaces are useful but require manual reconstruction and do not expose every timing and state transition in a form suitable for distributed-tracing tools. OpenTelemetry would provide:

- detailed debugging of workflow control flow and failure paths;
- performance analysis across agents, groups, subworkflows, and tools;
- identification of latency and cost bottlenecks;
- data for subsequent prompt and workflow optimization;
- a standards-based integration surface rather than a backend-specific telemetry implementation.

## Proposed configuration

Telemetry is disabled by default and enabled per workflow:

```yaml
workflow:
  runtime:
    telemetry:
      enabled: true
      service_name: conductor   # optional
```

Exporter and SDK behavior should use standard OpenTelemetry environment variables, including:

```bash
OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
OTEL_EXPORTER_OTLP_PROTOCOL=grpc
OTEL_SERVICE_NAME=conductor
```

The SDK/exporters should ship as an optional `telemetry` extra. A workflow with telemetry enabled but the extra unavailable should warn with the correct install command and continue without spans.

### Event-driven orchestration tracing

Build orchestration spans in a dedicated `TelemetrySubscriber` over the existing `WorkflowEventEmitter` stream rather than adding tracing calls throughout the engine hot path.

The subscriber should:

- create detached spans with explicit parent contexts because event callbacks do not own the async task context between paired events;
- preserve event timestamps, converting Unix seconds to the nanoseconds expected by the OTel SDK;
- track open spans by stable identities for workflows, nested contexts, agents, parallel members, for-each item indexes, validators, and tool calls;
- close and remove every span on success, failure, cancellation, workflow failure, or subscriber shutdown;
- handle duplicate `workflow_started` events during resume without creating duplicate roots;
- receive the directly synthesized root event used by the resume `--web` path, where the engine root emission is suppressed;
- close gate presentation and resolution as separate short spans rather than holding a span open while waiting for a human;
- flush on shutdown with a bounded `force_flush` wait and explicit OTLP exporter timeout.

Use OpenTelemetry GenAI semantic conventions where applicable:

- `gen_ai.operation.name`: `invoke_workflow`, `invoke_agent`, `execute_tool`;
- `gen_ai.conversation.id`;
- `gen_ai.agent.name`, `gen_ai.tool.name`;
- `gen_ai.provider.name`, `gen_ai.request.model`;
- `gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens`;
- standard `error.type` on failures.

Conductor-specific data should remain under `conductor.*`, for example cost, step type, group name, item key, iteration, resumed, and superseded state.

### Pydantic AI native instrumentation

For `claude` and `openai`, enable Pydantic AI instrumentation only while Conductor telemetry is active. Pass the Conductor `run_id` explicitly as `conversation_id` to Pydantic AI execution calls and provide the active `TracerProvider` explicitly.

The v1 contract is intentionally **two correlated trace trees**, not one parented tree:

1. the Conductor orchestration trace;
2. native Pydantic AI traces for model and tool internals.

They share `gen_ai.conversation.id = run_id` and the same provider, but cross-layer context parenting is out of scope for v1. Documentation must state this clearly.

Validator agents should receive the same native instrumentation as primary agents because their latency and cost are part of the run.

### Privacy

Prompt and response content must not be captured by default. Content capture is controlled only by the standard `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` environment variable, with no Conductor-specific YAML switch.

Pydantic AI instrumentation must be configured with an explicit `include_content` policy. In particular, do not use a bare global instrumentation switch whose upstream default captures content.

### Included

- OTLP export over gRPC or HTTP/protobuf;
- optional telemetry dependency extra;
- workflow-level `runtime.telemetry` configuration;
- orchestration spans for workflow, agent, group, item, step, validator, gate, and tool lifecycle events;
- native Pydantic AI instrumentation for `claude` and `openai`;
- run/resume parity, including dashboard resume event seeding;
- validation warning when the optional SDK is unavailable;
- an example workflow and documentation for Jaeger/Tempo/Grafana setup, environment variables, privacy, and experimental GenAI semantic conventions.

### Out of scope for v1

- a new telemetry CLI command;
- `~/.conductor/config.toml` settings;
- dashboard integration;
- cross-layer parenting that merges Conductor and Pydantic AI spans into one trace;
- native provider telemetry for Copilot, Claude Agent SDK, Hermes, or ACA beyond orchestration-level attributes already available from Conductor events;
- prompt-content capture unless explicitly enabled through the standard OTel environment variable;
- vendor-specific exporters other than OTLP.

## Acceptance criteria

- [ ] Telemetry remains a silent no-op by default: no spans, warnings, or behavior changes when `runtime.telemetry` is absent or disabled.
- [ ] `runtime.telemetry.enabled: true` plus an OTLP endpoint exports a workflow-to-agent/group/item/tool orchestration tree.
- [ ] All success, failure, cancellation, fail-fast, and resume paths leave no open spans.
- [ ] Resume with and without `--web` produces the correct root span and preserves the run correlation identifier.
- [ ] For-each items remain distinct even when authored item keys collide; item index is available in every relevant per-item event payload.
- [ ] Repeated same-name tool calls produce distinct spans and retain the correct agent/item parent.
- [ ] `claude` and `openai` emit native Pydantic AI traces through the same `TracerProvider`, with `gen_ai.conversation.id` equal to the Conductor `run_id`.
- [ ] Documentation describes the native traces as a second correlated tree, not a single end-to-end trace.
- [ ] Prompt and response content is absent from spans by default and appears only when explicitly enabled via the standard OTel environment variable.
- [ ] Missing telemetry dependencies, `OTEL_SDK_DISABLED=true`, exporter initialization failure, and an unreachable endpoint never crash or block workflow completion beyond the documented flush/export timeout.
- [ ] Documentation includes an OTLP quickstart, privacy guidance, semantic-convention stability notice, configuration reference, and a validated example workflow.
- [ ] Unit and integration tests use an in-memory span exporter and cover normal execution, agent failure, resume, fail-fast cancellation, for-each key collision, tool spans, disabled SDK, missing extra, content policy, and exporter degradation.

hertznsk:
Closes #481

> **Note:** This PR was substantially reworked after its first iteration, and the branch history was rewritten. The commit list is therefore not a useful guide to the evolution of the implementation; this description reflects the final state.

## What this adds

This PR adds opt-in OpenTelemetry tracing for Conductor workflow execution. Tracing is activated entirely through the standard `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable; no workflow YAML setting is required.

When enabled, Conductor exports one unified trace tree:

```text
invoke_workflow
├── invoke_agent (model, tokens, cost, duration)
│   ├── chat <model>            # native Pydantic AI span
│   ├── execute_tool            # native provider span where available
│   └── MCP send tools/call     # mcp-python-sdk span
├── parallel / for_each groups and items
├── script / set / wait steps
└── human-gate presentation and resolution markers
```

The implementation includes:

- A new `conductor.telemetry` package with environment-driven tracer setup, OpenTelemetry GenAI semantic-convention attributes, and an event-driven `TelemetrySubscriber`. The subscriber builds spans from Conductor's existing workflow event stream rather than adding tracing calls throughout the engine hot path.
- A process-lifetime `_DelegatingTracerProvider` that routes spans from instrumented libraries to the active run's provider. This supports multiple sequential runs in one process without sending later spans to a provider that an earlier run shut down, while preserving any global provider already installed by a host application.
- A task-local context bridge that attaches orchestration spans in the owning asyncio task. Native Pydantic AI and MCP spans therefore nest below Conductor's `invoke_agent` spans instead of appearing as unrelated root traces. Context tokens are detached only by their owning task, including fail-fast and cancellation paths.
- Native Pydantic AI instrumentation for the `claude` and `openai` providers. Conductor supplies the active tracer provider explicitly and passes the workflow `run_id` as `conversation_id`, so native spans also carry `gen_ai.conversation.id`.
- Native Copilot CLI spans for spawned Copilot runtimes when the OTLP protocol is `http/protobuf` or `http/json`. With the default gRPC protocol, native Copilot spans remain disabled, Conductor keeps its own tool spans so no telemetry is lost, and one warning per run explains how to enable the HTTP path. Externally hosted Copilot runtimes are not configured by Conductor.
- Per-provider tool-span deduplication through `ProviderCapabilities.native_otel_spans` and the per-execution `native_otel_spans_active` event field. Conductor suppresses its synthetic tool span only when the provider's native span is actually active for that execution.
- Run-latched OTLP endpoint and protocol state shared by tracer setup, engine event emission, and Copilot client construction. A process embedding Conductor cannot accidentally desynchronize export and deduplication decisions by changing environment variables during a run.
- Complete orchestration coverage for workflows, nested workflows, agents, validators, parallel members, for-each items, scripts, set/wait steps, tools, questions, and human gates. Gate waiting time is deliberately excluded from the active agent span.
- Token breakdowns on parallel-agent completion events and stable `index` identity on every for-each item event, preventing colliding authored item keys from mixing spans.
- `opentelemetry-api` as a core dependency and the SDK plus OTLP exporters in the optional `telemetry` extra. `conductor validate` warns when an endpoint is configured but the extra is unavailable.
- User documentation in `docs/telemetry.md`, a Jaeger quickstart, `examples/telemetry.yaml`, provider-comparison and environment-variable documentation, bundled-skill references, and changelog updates.

## Deliberate deviations from #481

The final implementation differs from the original proposal in several significant ways:

- **Environment-only activation instead of `runtime.telemetry`.** Setting `OTEL_EXPORTER_OTLP_ENDPOINT` enables tracing for a run. The unreleased YAML form is explicitly rejected with guidance to use the environment variable, preventing a silently ignored stale configuration.
- **One trace tree instead of two correlated trees.** The original issue proposed separate orchestration and native Pydantic AI trees for v1. Testing exposed noisy orphan native and MCP roots, so the final design uses a task-local context bridge to parent native spans directly below Conductor's orchestration spans. `gen_ai.conversation.id = run_id` remains available for correlation and querying.
- **Copilot native telemetry is now included.** Although it was originally out of scope, the Copilot SDK already supports tracing in its spawned CLI process. Conductor now enables that integration for OTLP HTTP while retaining a lossless Conductor-side fallback for gRPC.
- **MCP tracing uses MCP Python SDK v2.** Native MCP spans only exist in v2, so the dependency was raised to `mcp>=2,<3` and the obsolete v1 field-name compatibility shim was removed.

The privacy posture remains unchanged: prompt and response content is not captured by default. Content capture is controlled only by `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`.

## Additional fixes

- Agent start events now expose the resolved provider; parallel completion/failure events retain input and output token counts.
- For-each lifecycle and callback events now include the numeric item index for unambiguous correlation.
- The bundled skill's provider capability table now uses the correct `openai` provider name.

## Verification

- `make check`, `make test`, `make test-frontend`, and `make validate-examples` pass.
- The new telemetry suite covers activation and SDK-degradation paths, exporter failures, run-latched configuration, task ownership and LIFO context cleanup, questions and gate lifecycles, native/synthetic tool-span deduplication, sequential runs in one process, resume with and without `--web`, fail-fast cancellation, for-each key collisions, and Copilot client telemetry configuration.
- Tests use an in-memory exporter and include failure-path assertions; the unreachable-exporter case uses loopback only.
- The implementation has also been exercised with real workflows and a local collector to inspect trace shape, nesting, failures, and the disabled path. Native Copilot CLI spans have not yet been visually verified in a real collector.


hertznsk:
I mark it with a draft to test it.

hertznsk:
I finished testing this in real workflows, made a number of fixes based on what that testing surfaced, and pushed the updated implementation. The branch was force-pushed with rewritten history, and the PR description now reflects the final state rather than the original iteration.

Full disclosure: I'm still not entirely happy with how parts of this are implemented. However, I don't think keeping it to myself any longer is likely to make it better. Given how thorough your reviews are known to be, I hope the remaining weaknesses will be found, and that the community can then help refine this into something excellent.

There is one thing I specifically did not have an opportunity to verify: what the traces actually look like when using the `copilot` provider. Native Copilot CLI telemetry, its HTTP-protocol gating, the gRPC fallback, and tool-span deduplication are covered by tests against the SDK contract, but I have not seen those spans rendered in a real collector. If you get a chance to inspect a Copilot-backed workflow in Jaeger, Tempo, or another collector, I would be very grateful.

jrob5756 (commented):
This adds OpenTelemetry tracing across workflow execution, and the mechanism mostly works, but there are real correctness bugs in the span hierarchy and a couple of gaps that undercut the privacy story the docs promise. CI is also red on the reviewed commit, so this can't merge as-is.

**Blocking, by location:**
- `src/conductor/telemetry/setup.py:133` — HTTP exporter posts to the wrong path
- `docs/telemetry.md:118` — privacy guarantee doesn't hold for exception messages
- `src/conductor/telemetry/subscriber.py:61` — flush timeout doesn't actually bound shutdown
- `tests/test_telemetry/test_provider_instrumentation.py:76` — new tests fail in the CI environments as configured
- `src/conductor/telemetry/subscriber_state.py:173` — nested workflow spans get skipped
- `src/conductor/telemetry/subscriber_execution.py:191` — concurrent for-each validators can close each other's spans
- `src/conductor/telemetry/subscriber_workflows.py:142` — for-each child workflow span closes early, before item metadata lands
- `src/conductor/engine/workflow.py:4469` — native-span detection reads the wrong workflow's settings, breaks under an inherited external runtime
- CI run [33962179167](https://github.com/microsoft/conductor/actions/runs/33962179167) is failing (attempt 2, overall)
- The Type Check job in that same run is failing: https://github.com/microsoft/conductor/actions/runs/33962179167/job/102472908882

The recommended list covers real gaps too — inaccurate `native_otel_spans_active` reporting, cancellations exported as clean completions, telemetry cleanup skipped on dashboard shutdown errors, and a couple of tests that assert less than they claim to. Worth fixing before this settles, but none of it should hold up the merge once the blocking items are addressed.

src/conductor/telemetry/setup.py
```diff
@@ -0,0 +1,135 @@
...
+def _create_otlp_exporter(protocol: str, endpoint: str) -> SpanExporter:
+    """Create the OTLP exporter selected by the captured protocol and endpoint."""
+    if protocol == "grpc":
+        module_name = "opentelemetry.exporter.otlp.proto.grpc.trace_exporter"
+    else:
+        module_name = "opentelemetry.exporter.otlp.proto.http.trace_exporter"
+
+    exporter_module = import_module(module_name)
+    return exporter_module.OTLPSpanExporter(
+        endpoint=endpoint,
```
jrob5756, line 221: **BLOCKING**: Passing `OTEL_EXPORTER_OTLP_ENDPOINT` straight through as the HTTP exporter's explicit endpoint skips its automatic `/v1/traces` suffix. With the documented `http://localhost:4318` config, spans get posted to `/` instead of `/v1/traces`, and any standard collector rejects that — which breaks both the HTTP example and the orchestration half of unified Copilot traces.

Derive the HTTP traces endpoint from the captured base URL using the normal OTLP path-appending rule, and leave the gRPC endpoint handling as-is. Keep the base URL you're passing to Copilot intact.

docs/telemetry.md
```diff
@@ -0,0 +1,128 @@
...
+
+Conductor emits its own spans for MCP tool calls from the workflow's tool events. The `mcp` Python SDK can additionally export native spans starting with version `2.0.0`; Conductor currently depends on `mcp>=1.28.1,<2`, so those native spans become available only once that upgrade lands.
+
+### Human-Gate Spans
+
+When a workflow hits a human-in-the-loop gate, the active `invoke_agent` span for that gate agent ends immediately at presentation. Conductor then emits zero-duration event spans for `gate_presented` and `gate_resolved` to track the state transitions. This prevents the agent span from remaining active and showing a misleading duration while waiting for human input.
+
+## Privacy and Content Capture
+
+By default, Conductor does not record prompt text, system messages, or model responses in OTel spans. This protects sensitive data and keeps secrets out of your tracing backend.
```
jrob5756: **BLOCKING**: Turning off message-content capture doesn't actually keep response content out of traces. `SpanState.end()` exports failure messages as-is, both as `error.message` and in the span status description, and output-validation exceptions can include the rejected value itself. So a sensitive response can end up in agent and workflow spans with content capture off. As written, this guarantee could talk someone into pointing tracing at a collector that isn't cleared for that data.

Suggested replacement: "Prompt and response capture is disabled by default, but exception messages can contain input or response values. Treat traces as potentially sensitive even with content capture disabled." The same claim shows up in the configuration and example docs and needs the same fix.

src/conductor/telemetry/subscriber.py
```diff
@@ -0,0 +1,128 @@
...
+            return
+        self._closed = True
+        if self._state is not None:
+            closed_event = WorkflowEvent(type="telemetry_closed", timestamp=time.time())
+            self._state.finish_all(closed_event, failed=False)
+            self._state.detach_close_tokens()
+            self._state.clear_indexes()
+        if self._tracer_provider is not None:
+            with suppress(Exception):
+                self._tracer_provider.force_flush(timeout_millis=5_000)
```
jrob5756: **BLOCKING**: Both CLI execution helpers call `close()` synchronously on the asyncio thread, and on the pinned OpenTelemetry SDK (1.44.0) the batch processor ignores `force_flush()`'s timeout and drains synchronously. A slow or unreachable collector can trigger repeated ten-second export attempts, blocking workflow teardown and cancellation well past the deadline this code implies.

Keep span closure and context detachment on the owning task, but move the actual exporter I/O off the event loop and enforce a single overall shutdown deadline. Bound the draining/retry behavior, report an incomplete export instead of hanging on it, and make sure executor or interpreter shutdown can't end up waiting on leftover export work.

tests/test_telemetry/test_provider_instrumentation.py
```diff
@@ -0,0 +1,226 @@
...
+    )
+
+
+def test_active_telemetry_configures_pydantic_ai_without_message_content(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Requirement: native tracing uses Conductor's provider and hides content by default."""
+    monkeypatch.setenv("OTEL_EXPORTER_OTLP_ENDPOINT", "http://localhost:4317")
+    provider = init_tracer_provider(run_id="run-telemetry")
+    assert provider is not None
```
jrob5756, line 77: **BLOCKING**: Five cases here call `init_tracer_provider()` and assume it returns a real SDK provider, with no guard for the optional dependency. CI installs dev deps plus `claude-agent-sdk`, and release testing installs only dev deps — neither installs telemetry, so these five fail outright while the rest of the suite quietly skips. Net effect: the feature isn't actually exercised by either workflow.

Add telemetry-enabled coverage to CI and release validation, and guard the SDK-dependent cases so they don't fail on a base install. Keep a separate SDK-absent path that checks normal uninstrumented execution still works — just skipping these five would leave tracing itself untested.

src/conductor/telemetry/subscriber_state.py
```diff
@@ -0,0 +1,334 @@
...
+                    self._discard_indexes(key)
+            elif owner_task is not None and owner_task.done():
+                self._attach_tokens.pop(key, None)
+                logger.debug("Discarded OpenTelemetry context token from completed worker task")
+
+    def parent_for_path(self, path: PathKey) -> SpanKey | None:
+        """Return the nearest open workflow or remembered subworkflow parent."""
+        remembered_parent = self.subworkflow_parents.get(path)
+        if remembered_parent in self.open_spans:
+            return remembered_parent
```
jrob5756: **BLOCKING**: `parent_for_path()` returns the remembered outer invocation before it checks whether there's an open workflow span at the exact child path. So once a child workflow starts, its agents and groups attach to the outer invocation instead of the child workflow span — the exported tree reads `delegate → inner-agent` when it should read `delegate → inner-workflow → inner-agent`.

Prefer an open workflow span for the exact path; only fall back to the remembered invocation parent before that span exists, and to ancestor workflows after. Add a test that actually nests engine execution rather than just asserting the lookup in isolation.

src/conductor/telemetry/subscriber_execution.py
```diff
@@ -0,0 +1,317 @@
...
+    state.agent_keys.setdefault((path, validator), deque()).append(key)
+
+
+def validator_completed(state: SpanState, event: WorkflowEvent) -> None:
+    """Finish a validator span, marking execution errors as failures."""
+    path = event_path(event, "subworkflow_path")
+    agent = event_text(event, "agent_name")
+    if agent:
+        state.end(
+            state.latest_agent(path, f"{agent} (validator)"),
```
jrob5756: **BLOCKING**: For-each validator callbacks all share the group's `agent_name` and are told apart by `index`/`item_key`, but the validator lookup here ignores both and just grabs the most recently started validator for the group. Two overlapping validators means item 0 can close item 1's span and inherit its token count, model, and error metadata.

Key validator spans by the full execution identity — workflow path, group, and item index for for-each work — and resolve both parent and completion off that key instead of the shared display name.

src/conductor/telemetry/subscriber_workflows.py
```diff
@@ -0,0 +1,143 @@
...
+    _finish_subworkflow_parent(state, event, failed=True)
+
+
+def _finish_subworkflow_parent(state: SpanState, event: WorkflowEvent, *, failed: bool) -> None:
+    parent_path = event_path(event, "parent_path")
+    slot_key = event_text(event, "slot_key")
+    if slot_key is None:
+        return
+    parent = state.subworkflow_parents.pop((*parent_path, slot_key), None)
+    if parent and parent[0].startswith(f"{INVOKE_AGENT} "):
```
jrob5756: **BLOCKING**: Sequential agent spans and for-each item spans both have names starting with `invoke_agent`, so this prefix check closes a for-each item's span on `subworkflow_completed` — before `for_each_item_completed` has a chance to attach its terminal metadata. Once the span is closed, that later item event can't update it, so things like the child's aggregated `cost_usd` are lost.

Track whether each remembered subworkflow parent is a sequential agent or a for-each item, and only close sequential parents here. Let `for_each_item_completed`/`for_each_item_failed` own item termination.

src/conductor/engine/workflow.py
```diff
@@ -4458,6 +4464,11 @@ async def _execute_loop(self, current_agent_name: str) -> dict[str, Any]:
                         }
                         if is_llm_agent:
                             started_payload["working_dir"] = resolved_agent.working_dir
+                            started_payload["native_otel_spans_active"] = native_otel_spans_active(
+                                event_provider,
+                                self.config.workflow.runtime.provider,
```
jrob5756: **BLOCKING**: Child engines inherit the parent's provider registry, but this activation check reads the child workflow's own runtime settings rather than the provider actually running the agent. So a parent configured with an external Copilot `runtime_url` plus a child on default settings still reports native spans as active under HTTP — even though the inherited external runtime was never configured for Conductor tracing. The subscriber then suppresses its fallback tool spans, and tools go untraced against an uninstrumented external runtime.

Derive provider identity and native-span activation from the provider/registry actually executing the agent, inherited connection settings included, and apply the same resolution consistently across sequential, parallel, and for-each start events.

src/conductor/providers/capabilities.py
```diff
@@ -519,6 +525,46 @@ def uses_native_skills(provider_type: str) -> bool | None:
...
+    """Return whether the resolved provider can emit native OTEL spans for this run."""
+    try:
+        native_otel_spans = get_capabilities(provider_name).native_otel_spans
+    except (AttributeError, ImportError, KeyError):
+        return False
+
+    if not native_otel_spans:
+        return False
+    if provider_name != "copilot":
+        return True
```
jrob5756, line 579: **RECOMMENDED**: For Claude and OpenAI, `native_otel_spans_active()` returns true off static capability support alone — it doesn't check `guards.is_telemetry_active()`, which is what the actual instrumentation depends on. A run with no endpoint configured, tracing disabled, or a failed init still reports `native_otel_spans_active=true` on its events, despite producing no native tracing.

Require an active-run telemetry check before evaluating provider-specific availability, and keep the static support answer in `has_native_otel_spans()` where it belongs.

src/conductor/telemetry/subscriber.py
```diff
@@ -0,0 +1,128 @@
...
+        _dispatch(self._state, event)
+
+    def close(self) -> None:
+        """Finish open spans, flush exporters, and reset process-local guards."""
+        if self._closed:
+            return
+        self._closed = True
+        if self._state is not None:
+            closed_event = WorkflowEvent(type="telemetry_closed", timestamp=time.time())
+            self._state.finish_all(closed_event, failed=False)
```
jrob5756: **RECOMMENDED**: `KeyboardInterrupt` or an external asyncio cancellation can exit the engine without ever emitting `workflow_failed`. `close()` then finishes every remaining span with `failed=False`, so interrupted spans end up looking identical to successful ones — there's no way to tell a completed run from an interrupted one in the trace.

Pass the terminal outcome from both CLI helpers into telemetry cleanup, and mark unfinished spans as cancelled/interrupted/incomplete rather than successful — without touching spans a genuine terminal event already closed.

src/conductor/cli/run.py
```diff
@@ -2543,6 +2554,9 @@ async def run_workflow_async(
         if dashboard is not None:
             await dashboard.stop()
 
+        if telemetry_subscriber is not None:
```
jrob5756, line 2606: **RECOMMENDED**: `dashboard.stop()` can raise from its server task, and because telemetry cleanup runs after that `await`, an exception there skips exporter shutdown and telemetry-state reset entirely. The resume helper has the identical ordering at lines 3378–3379, and its CLI-attached root context can be left attached in the same failure case.

Wrap dashboard shutdown in a `finally` in both helpers so telemetry cleanup always runs, and keep that cleanup fail-open so it can't mask the original dashboard or workflow exception.

src/conductor/telemetry/setup.py
```diff
@@ -0,0 +1,135 @@
...
+    if not guards.OTEL_SDK_AVAILABLE:
+        guards.reset_telemetry_context()
+        _warn_sdk_unavailable_once()
+        return None
+
+    protocol = _resolve_otlp_protocol()
+    try:
+        provider = _build_tracer_provider(run_id, endpoint, protocol)
+        _install_delegating_global_provider()
+    except Exception:  # noqa: BLE001 -- optional tracing must never stop a workflow.
```
jrob5756, line 70: **RECOMMENDED**: If the provider constructs fine but global-provider discovery raises afterward, initialization logs the error and returns `None` without shutting down the provider, batch worker, or exporter it already allocated. A misconfigured `OTEL_PYTHON_TRACER_PROVIDER` can hit this repeatedly in a long-lived process, and partial construction also means a `TracerProvider` gets allocated before the exporter is even created.

Make initialization transactional: hold onto allocated resources until setup fully succeeds, and roll back on any failure including inside `_build_tracer_provider()`. Log rollback errors separately, and never shut down a provider the host owns rather than Conductor.

src/conductor/telemetry/subscriber.py
```diff
@@ -0,0 +1,128 @@
...
+            self._state.detach_close_tokens()
+            self._state.clear_indexes()
+        if self._tracer_provider is not None:
+            with suppress(Exception):
```
jrob5756: **RECOMMENDED**: `close()` ignores a `False` return from `force_flush()` and swallows exceptions from both flush and shutdown. That's a reasonable fail-open default, but treating a failed cleanup identically to a successful one means spans can go unexported, or resources unclosed, with zero diagnostic — missing traces become very hard to track down.

Check the flush result and log flush/shutdown failures with the run ID and a note about possible trace loss. Attempt shutdown independently of flush, and reset telemetry guards in an outer `finally`.

docs/telemetry.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Unified Traces
+
+For `copilot`, `claude`, and `openai` providers, Conductor automatically registers native OpenTelemetry instrumentation. Conductor unifies high-level orchestration spans and low-level LLM call spans into a single unified trace tree, instead of generating separate, disconnected trace trees.
+
+### Parenting and Trace Structure
+
+- **Orchestration Spans as Parents:** The high-level Conductor orchestration spans, such as `invoke_agent` and `execute_tool`, act as parent spans.
+- **Provider Spans as Children:** Native provider spans, including LLM API calls and token counts, are nested directly under the corresponding Conductor spans. They share the same `trace_id` and trace context.
+- **Correlation Identifier:** Conductor propagates the workflow `run_id` as the `gen_ai.conversation.id` attribute across all spans in the tree.
```
jrob5756: **RECOMMENDED**: The claim that every span carries the workflow `run_id` as `gen_ai.conversation.id` doesn't hold for Copilot. Conductor stamps its own spans and passes `conversation_id` to Pydantic AI, but the Copilot configuration path never forwards that identifier — and W3C trace-context propagation carries trace identity, not span attributes, so filtering native Copilot spans by workflow conversation ID isn't reliable.

Document that Conductor and Pydantic AI spans use the workflow run ID, while native Copilot spans should be correlated by shared trace ID and parent relationship instead. The execution skill reference makes the same claim and needs the same correction.

examples/README.md
```diff
@@ -152,6 +152,24 @@ conductor run examples/context-tier.yaml \
...
+## OpenTelemetry Tracing
+
+Enable tracing by setting an OTLP endpoint after installing the optional extra. The trace
+contains provider-independent workflow, agent, and tool spans for every
+provider. Native spans from `copilot`, `claude`, and `openai` are nested
+within a single unified trace tree.
+
+```bash
+uv sync --extra telemetry
+export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
```
jrob5756: **RECOMMENDED**: This recipe runs `simple-qa.yaml` with Copilot selected, but leaves the protocol at its gRPC default, and `_build_client_telemetry()` supplies no native tracing config under gRPC. So the recipe can't actually demonstrate the native Copilot spans it advertises right above it.

Once the HTTP endpoint handling is fixed, switch this to port 4318 with `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf`, or just label this recipe explicitly as orchestration-only for Copilot.

tests/test_telemetry/test_integration.py
```diff
@@ -0,0 +1,1131 @@
...
+            custom_agents: list[dict[str, Any]] | None = None,
+            extra_mcp_servers: dict[str, Any] | None = None,
+        ) -> AgentOutput:
+            pydantic_agent = Agent(
+                TestModel(call_tools=["echo"]),
+                tools=[Tool(lambda value: value, name="echo")],
+                name=agent.name,
+                retries=0,
+            )
+            pydantic_agent.instrument = None
```
jrob5756: **RECOMMENDED**: This deduplication test builds its own Pydantic agent, explicitly disables instrumentation, and synthesizes tool callbacks after the fact. Its assertions only check for a workflow root and a connected tree, so it passes even when native tool spans are completely absent — which is exactly the failure mode that suppressing Conductor's fallback spans is supposed to guard against.

Exercise the real Claude/OpenAI builder and runner with only the model swapped for `TestModel`, and assert that native invocation, model, and tool spans actually reach the run-local exporter with correct ancestry and conversation identity — with exactly one tool span, no Conductor duplicate.

tests/test_telemetry/test_integration.py
```diff
@@ -0,0 +1,1131 @@
...
+        patch("conductor.fleet.retention.maybe_prune_event_logs"),
+        patch("sys.stdin") as mock_stdin,
+        pytest.raises(ExecutionError),
+    ):
+        mock_stdin.isatty.return_value = False
+        await run_workflow_async(wf_path, {})
+
+    # Cross-task detach ownership means no detach token for worker spans was
+    # reset from the main task during fail-fast cleanup.
+    assert detached_in_main == []
```
jrob5756: **RECOMMENDED**: `spy_detach` is defined but never installed, so `detached_in_main` stays empty no matter what production code does — the assertion can't actually catch a worker's context token being detached by another task during fail-fast cleanup. Installing the current spy alone won't fix this either, since it relies on `loop._task` rather than tracking real token ownership.

Install attach/detach recording around the fail-fast execution, record each token's owning task via `asyncio.current_task()`, and assert worker tokens are observed and never detached by another task — plus that cleanup restores the caller's original context.

src/conductor/telemetry/setup.py
hertznsk: Fixed. Conductor now resolves general vs trace-specific OTLP endpoint/protocol precedence from one copied environment snapshot. HTTP general endpoints get `/v1/traces`; signal-specific URLs and gRPC endpoints remain verbatim; the unsuffixed general base URL remains the only endpoint passed to Copilot. I added coverage for trailing/custom paths, per-signal precedence, signal-only activation, and the snapshot race. Commits: `3228b58`, `d09b479`, `106ad79`, `cb397c3`.

docs/telemetry.md
hertznsk: Agreed and fixed. The telemetry and configuration docs now explicitly say that exception messages may contain input/response values and that traces must be treated as potentially sensitive even when prompt/response capture is disabled. Commit: `f7d1848`.

src/conductor/telemetry/subscriber.py
hertznsk: Fixed. Span/context closure still runs on the owning task, while exporter flush/shutdown runs on a daemon thread with one five-second overall deadline. `shutdown_on_exit=False` prevents the SDK's atexit handler from reintroducing an interpreter-exit stall, and timeout/failure paths report possible trace loss. Thread start/join failures are contained and tested. Commits: `ee294de`, `d09b479`, `106ad79`.

tests/test_telemetry/test_provider_instrumentation.py
hertznsk: Fixed. SDK-dependent instrumentation cases now use `importorskip`; a separate SDK-absent path verifies normal uninstrumented execution. CI and release test/typecheck jobs install the telemetry extra, so the feature is exercised rather than silently skipped. Signal-specific endpoint diagnostics are covered too. Commits: `1bf3cc0`, `106ad79`, `cb397c3`.

src/conductor/telemetry/subscriber_state.py
hertznsk: Fixed. `parent_for_path()` now prefers an open workflow span at the exact child path, using the remembered invocation only before that child span exists. A real nested-engine regression test proves `delegate -> inner-workflow -> inner-agent` and fails with the old ordering. Commits: `ba04565`, `1bf3cc0`.

src/conductor/telemetry/subscriber_execution.py
hertznsk: Fixed. For-each validator spans are keyed and resolved by `(workflow path, group, item index)`, with the item span as parent. The overlapping-validator test verifies independent completion metadata/status. Commit: `ba04565`.

src/conductor/telemetry/subscriber_workflows.py
hertznsk: Fixed. Remembered subworkflow parents now carry whether they are sequential agents or for-each items. `subworkflow_completed/failed` closes only sequential parents; item terminal envelopes retain ownership so cost/token metadata lands before closure. Commit: `ba04565`.

src/conductor/engine/workflow.py
hertznsk: Fixed. Provider identity is resolved through `ProviderRegistry.provider_type_for()`, the same path used by actual execution, and settings come from that inherited registry. This is applied to sequential, parallel, and for-each events. A real nested-engine test covers root external Copilot + child OpenAI default and verifies the event still reports Copilot, native spans stay inactive, and the fallback tool span remains. Commits: `e4cf652`, `d09b479`.

src/conductor/providers/capabilities.py
hertznsk: Fixed. `native_otel_spans_active()` now requires an active telemetry run before checking provider capability; Copilot additionally requires an HTTP protocol and a general/base endpoint actually supplied to its runtime. Static support remains in `has_native_otel_spans()`. Commits: `e4cf652`, `d09b479`.

src/conductor/telemetry/subscriber.py
hertznsk: Fixed. Both CLI helpers pass terminal status/type/message into telemetry cleanup. Only spans still open at close are marked failed/incomplete, so spans already closed by genuine terminal events are untouched. Commit: `ee294de`.

src/conductor/cli/run.py
hertznsk: Fixed. Dashboard shutdown, telemetry close, event-log close, and file-logging close are independent fail-open teardown operations in both run and resume. Dashboard cleanup failures are logged, cannot replace the primary workflow/resume exception, and cannot skip later resources; parity regressions cover both paths. Commits: `ee294de`, `d09b479`.

src/conductor/telemetry/setup.py
hertznsk: Fixed. Initialization now owns and rolls back provider/exporter/processor resources transactionally across exporter construction, processor construction, processor attachment, and global-provider installation failures. Host-owned providers are never closed. Commits: `3228b58`, `d09b479`.

src/conductor/telemetry/subscriber.py
hertznsk: Fixed. `force_flush()` false returns, flush exceptions, shutdown exceptions, and the overall deadline now log the run ID and possible trace loss. Shutdown remains independent of flush, and telemetry guards reset in an outer `finally`. Commit: `ee294de`.

docs/telemetry.md
hertznsk: Fixed. Docs now distinguish Conductor/Pydantic AI `gen_ai.conversation.id` from Copilot correlation by shared trace ID/parent relationship, including the execution skill reference. The external-runtime section no longer promises automatic propagation; it states that correlation depends on runtime-connection W3C support. Commits: `f7d1848`, `d09b479`.

examples/README.md
hertznsk: Fixed. The recipe now uses port 4318 with `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf`, and explicitly explains that gRPC exports orchestration spans only for Copilot. Commit: `3228b58`.

tests/test_telemetry/test_integration.py
hertznsk: Fixed. The dedup regression now runs the real OpenAI builder and runner, swapping only the model for `TestModel`. It asserts native invocation/chat/tool spans, ancestry, conversation identity, and exactly one tool span so a Conductor duplicate fails the test. Commit: `1bf3cc0`.

tests/test_telemetry/test_integration.py
hertznsk: Fixed. The test installs attach/detach recording around the fail-fast execution, records real owner tasks by token identity, asserts worker tokens were observed and never detached cross-task, and verifies the caller context is restored. Commit: `1bf3cc0`.

hertznsk:
Added OpenTelemetry tracing support for the new deterministic MCP workflow step introduced in #518. Direct `type: mcp` executions are now exported as `execute_tool` spans with correct workflow, parallel-group, and for-each parentage, including collision-safe item identity, routed tool-error status, execution failures, and interrupted attempts. The spans intentionally exclude argument values, result contents, and spill paths. Targeted telemetry and MCP coverage passes (166 tests).

jrob5756 (approved):

Issue #397, opened 2026-08-10
Harden the web dashboard request surface: origin/host validation and consistent auth across HTTP + WebSocket


jrob5756:
## Summary

The web dashboard (`src/conductor/web/server.py`) serves an HTTP API and a WebSocket endpoint for the local browser UI. The request path currently has no origin or host validation, and the optional auth token covers only part of the surface. This issue tracks hardening that surface to standard practice for a locally-bound developer server.

This is defense-in-depth work, not a behavior change: the dashboard should continue to work exactly as it does today for the intended local-browser use case.

## Current gaps

1. **No origin/host validation.** `WebDashboard._create_app` builds the FastAPI app with no middleware, and no route validates `Origin`, `Referer`, or `Host`. A repo-wide search for `add_middleware` / `CORSMiddleware` / `TrustedHostMiddleware` returns nothing in the request path.

2. **Content type is not constrained on mutating routes.** Mutating endpoints parse their body regardless of the declared `Content-Type`.

3. **Auth coverage is inconsistent.** `_gate_token_ok` is applied to gate resolution only. Other mutating routes and the other WebSocket message types (`dialog_message` / `dialog_decline`, `iteration_limit_response`) are not covered by any equivalent check. Notably, `iteration_limit_response` resolves the `max_iterations` safety gate.

4. **The token can't be used from the browser UI.** `CONDUCTOR_GATE_TOKEN` is read from an `Authorization` header, but the dashboard sends gate responses over the WebSocket (`stores/workflow-store.ts`), and browsers cannot set request headers on a WebSocket handshake. So enabling the token today breaks the dashboard's own approve/reject buttons — there is no configuration that is both enabled and usable from the UI.

## Proposed work

- [ ] Add an origin/host allowlist enforced on **all** routes, including the `/ws` handshake — the WebSocket handshake is not covered by the browser's same-origin policy, so it needs the check explicitly rather than inheriting one. Reject requests whose `Origin` is not the dashboard's own and whose `Host` is not the bound loopback address and port.
- [ ] Require `application/json` on mutating routes.
- [ ] Apply consistent auth coverage across every mutating route and every inbound WebSocket message type, rather than gate resolution alone.
- [ ] Make the token usable in-browser: generate a per-run token automatically, deliver it to the page, and accept it on the WebSocket via a query parameter or subprotocol value (since headers aren't available there). Goal is for the protected configuration to be the default and remain fully functional.

[...]

PR #503, opened 2026-09-02
feat(providers): always-on client-side context compaction for pydantic-ai providers (claude, openai)

hertznsk:
## Summary

Agents running on the `claude` or `openai` providers no longer crash when a long conversation outgrows the model's context window. Instead, the oldest parts of the conversation are automatically condensed — a zero-LLM cleanup of old tool results first, an AI-written summary only when that is not enough, and a deterministic sliding-window fallback — and the run continues. Every compaction is surfaced in the console, the JSONL event log, and the web dashboard.

The implementation is built on the official `pydantic-ai-harness` `TieredCompaction` capability (client-side only, so behavior is identical behind API proxies) and is **always on** with no new YAML surface.

## What changed

- **Tiered compaction** for both pydantic-ai providers via a conductor-owned, fail-open wrapper: `ClearToolResults` → `SummarizingCompaction` (inherits the agent's model) → `SlidingWindowCompaction`. A compaction failure never aborts a run; an unrecovered failure disables compaction for the rest of that agent execution.
- **Reserve-based trigger**: `trigger = max(1, window − output_limit − tool_buffer)`, where `output_limit` is the minimum of the effective `max_tokens` actually sent to the API (user setting or the new unified default of **16384**, after Claude thinking coercion) and the provider-advertised model cap, and `tool_buffer` derives from `runtime.tool_output.max_chars`. The post-compaction target is a clamped 55% hysteresis ceiling.
- **Provider-advertised model limits**: the Claude provider now fully paginates the Anthropic models listing, and the OpenAI provider reads vendor-advertised token limits (`max_input_tokens` / `context_length` / `max_model_len` / `max_context_length`; `max_output_tokens` / `top_provider.max_completion_tokens`) from OpenAI-compatible endpoints (OmniRoute, LiteLLM, vLLM, OpenRouter, LM Studio) via a shared parser — no more conservative 128k guess behind a proxy. Endpoints that advertise nothing keep the registry/fallback cascade.
- **Unified `max_tokens` default of 16384** for both providers; the OpenAI provider now always sends the parameter (strict endpoints that reject it will fail fast). Claude thinking runs at `low`/`medium` effort without an explicit `max_tokens` now send 16384 (was 8192/12288). Dialog turns keep their 4096 cap.
- **Observability parity**: `agent_compaction_config` (once per execution, with the resolved window/limit/trigger/target and their sources), `agent_compaction_start` / `agent_compaction_complete` flow through the same event callback as every other pipeline event — console verbose log, JSONL event log, and the dashboard activity stream.

## Notes

- **No compaction for other providers** (copilot, claude-agent-sdk, hermes, aca); no new YAML options; no server-side vendor compaction; checkpoint/resume, `max_agent_iterations`, and the dashboard context-bar semantics are unchanged (the bar uses provider-only limits and may disagree with the compaction window — documented).
- A summarizing compaction consumes one request slot from the agent's `max_agent_iterations` budget (harness behavior, documented).
- Verified end-to-end against a live OpenAI-compatible gateway: a proxy-advertised 200k window resolves with source `provider`, `runtime.max_tokens` flips the output-limit source to `settings`, and unreachable endpoints fail as ordinary provider errors.

## Apology for the commit history

Sorry that the commit history on this branch is not as clean as it should be — it reflects the path we took while exploring the right approach to the problem. We expect the maintainer to **squash on merge**, as usual, so we left the history as-is rather than rewriting it.

## Test plan

- [x] `uv run pytest tests/test_providers -q -m "not performance"` — 1297 passed
- [x] Full suite `uv run pytest tests/ -q -m "not performance"` — green except two pre-existing environment-dependent unreadable-directory tests and one flaky timing-sensitive performance test (all fail identically on the base commit)
- [x] `ruff check` / `ruff format --check` clean; `ty check src` — no new diagnostics
- [x] `make validate-examples` passes (incl. new `examples/compaction.yaml`)
- [x] Live smoke against an OpenAI-compatible proxy (OmniRoute) for both the OpenAI and Anthropic-shaped paths


src/conductor/providers/_pydantic_ai/compaction.py
```diff
@@ -0,0 +1,379 @@
...
+                self._on_before(estimate=before_estimate, messages_before=len(before_messages))
+                start = time.monotonic()
+
+            result = await self._inner.before_model_request(ctx, request_context)
+
+            if not will_compact:
+                return result
+
+            after_messages = list(result.messages)
+            after_estimate = await _estimate_context_tokens(
```
jrob5756: **BLOCKING**

`_on_after` measures the post-compaction size with `_estimate_context_tokens(after_messages, ...)`, but the harness's estimator isn't a heuristic — it anchors on the most recent `ModelResponse` with provider-reported `usage.input_tokens` and only estimates the text after that anchor. Its own docstring says as much: a history rewritten after the anchor's request went out is overestimated unless the caller compensates, which is exactly what `TieredCompaction._escalate` does internally by subtracting a heuristic reclaim from its baseline. This wrapper never does that subtraction.

All three tiers keep the recent tail (`keep_pairs=3`, `keep_messages=20`), so the anchoring `ModelResponse` always survives compaction and the after-estimate equals the before-estimate. I checked this against the installed harness: a 60-message history compacted down to its last 20 (dropping two-thirds of the content, 45000 -> 15000 by the anchor-free heuristic) still reports `tokens_before = 100500`, `tokens_after = 100500`, `tokens_saved = 0`.

That's the headline observability feature of this PR, and it's broken for every real compaction. Console output, the JSONL log, and the dashboard will all show `150000 -> 150000` for a run that actually shrank its history — indistinguishable from a compaction that did nothing. The test at `tests/test_providers/test_pydantic_ai_compaction.py:767` only asserts `tokens_before >= tokens_after`, which equality satisfies, and its fixture has no `ModelResponse` usage anchor, so the estimator falls back to the pure heuristic there and never exercises this path.

Measure the reduction with the anchor-free heuristic, the same way the harness compensates internally:

```python
from pydantic_ai_harness.compaction import estimate_token_count

reclaimed = estimate_token_count(before_messages, None) - estimate_token_count(after_messages, None)
after_estimate = max(before_estimate - reclaimed, 0)
```

Add a regression test whose fixture includes a `ModelResponse(usage=RequestUsage(input_tokens=...))` inside the retained tail, and assert `tokens_saved > 0`.

src/conductor/providers/_pydantic_ai/compaction_window.py
```diff
@@ -0,0 +1,318 @@
...
+        window: Resolved context-window size in tokens.
+        output_limit: Resolved output-token cap in tokens.
+        tool_buffer: Tool-result reserve in tokens.
+
+    Returns:
+        Token threshold at which compaction should fire.
+    """
+    if output_limit + tool_buffer >= window - 1:
+        _warn_trigger_degenerate()
+    return max(1, window - output_limit - tool_buffer)
```
jrob5756: **BLOCKING**

`trigger_tokens` floors at `max(1, window - output_limit - tool_buffer)`. With this PR's defaults — `output_limit = DEFAULT_MAX_TOKENS = 16_384` and `tool_buffer = 40_000` — the reserve is 56,384 tokens. Any model whose resolved window is at or below that yields `trigger = 1`, and `target_tokens` collapses to 1 right alongside it. I checked this numerically: 8192, 16384, 32768, and 56384 all produce `trigger=1, target=1`.

`trigger = 1` isn't a low threshold, it's an always-true predicate. `_FailOpenCompactionWrapper.before_model_request` gates on `before_estimate > trigger`, so compaction fires from the first user turn. Since the target of 1 is unreachable, `TieredCompaction._escalate` never breaks early — it escalates through all three tiers on every single request, including `SummarizingCompaction`, which makes a real nested LLM call that (per this PR's own AGENTS.md note) burns a non-refunded slot from `max_agent_iterations`. The agent spends its iteration budget and API cost on summarizations that can never converge, while its history gets hard-trimmed to 20 messages every turn regardless.

This isn't a corner case. Any vLLM/LM Studio/proxy endpoint advertising `max_model_len: 8192` or `context_length: 32768` — the exact endpoints `model_listing.py` was added in this PR to read — lands here on the very first run. Large windows aren't safe either: `runtime.max_tokens: 100000` (the schema allows up to 200000) against the 128k fallback gives `128000 - 100000 - 40000 = -12000`, and `runtime.tool_output.max_chars` has `ge=1000` with no ceiling, so it can push `tool_buffer` past the window on its own. Compaction is always-on with no opt-out, and the only signal for any of this is a single one-shot `logger.warning` whose text ("Hysteresis is lost") badly undersells the consequence.

Treat "the reserve doesn't fit in the window" as *don't arm compaction*, not as `trigger = 1`. Have the resolution layer return `None` below a viable floor and have both providers pass `compaction=None` (already a valid value for the parameter):

```python
MIN_VIABLE_TRIGGER = 4_096

def trigger_tokens(*, window: int, output_limit: int, tool_buffer: int) -> int | None:
    trigger = window - output_limit - tool_buffer
    if trigger < MIN_VIABLE_TRIGGER:
        _warn_reserve_exceeds_window(window, output_limit, tool_buffer)
        return None
    return trigger
```

Also clamp `tool_buffer` to a fraction of the window so an unbounded `max_chars` can't exceed it, and have `emit_compaction_config` report the disabled state so it's visible to the operator.

src/conductor/providers/_pydantic_ai/compaction_window.py
```diff
@@ -0,0 +1,318 @@
...
+    re-compaction.
+
+    Args:
+        window: Resolved context-window size in tokens.
+        trigger: Trigger threshold from :func:`trigger_tokens`.
+
+    Returns:
+        Escalation-ceiling target in tokens.
+    """
+    return max(1, min(int(window * TARGET_FRACTION), trigger - 1))
```
jrob5756, line 432: **BLOCKING**

`target_tokens` returns `max(1, min(int(window * TARGET_FRACTION), trigger - 1))`. The `trigger - 1` clamp wins whenever `0.55 * window >= trigger`, i.e. whenever `0.45 * window <= output_limit + tool_buffer`. With the default 56,384-token reserve, that's every window below roughly 125,300 tokens — the target collapses to `trigger - 1`, and the post-compaction history sits one token below the threshold that just fired.

I measured the gap across realistic windows: 60000 -> 1 token, 64000 -> 1, 80000 -> 1, 100000 -> 1, 120000 -> 1. There's no hysteresis anywhere in that band — the next request re-crosses the trigger immediately and compacts again, escalating through the tiers and issuing another `SummarizingCompaction` call each time. Even `FALLBACK_CONTEXT_WINDOW = 128_000`, the value used whenever provider metadata and the registry both come up empty (the common path behind a proxy), yields a gap of only 1216 tokens — under 1% of the window. A single tool result blows past that.

The PR summary, docs/workflow-syntax.md, and the docstring at lines 306-309 all describe the 55% target as providing hysteresis against immediate re-compaction. In practice that only holds above ~125k, so the stated benefit is absent for most real configurations. The docstring's claim that the target is "clamped strictly below the trigger" is also literally false when `trigger == 1`, where `max(1, 0)` sets target equal to trigger.

Derive the target from a hysteresis margin that scales with the window instead of `trigger - 1`, e.g. `target = min(int(window * TARGET_FRACTION), trigger - max(1, int(window * 0.05)))`, and guarantee it stays positive. Combined with returning `None` for an unviable budget (see the `trigger_tokens` finding), `0 < target < trigger` becomes an invariant worth asserting directly in the function. Add a test whose reserve is large enough to actually exercise the clamp branch, and assert the resulting gap rather than just `target < trigger`.

src/conductor/providers/openai.py
```diff
@@ -206,15 +213,18 @@ def __init__(
...
         if temperature is not None:
             self._validate_temperature(temperature)
         self._default_temperature = temperature
 
+        # Validate and store max_tokens (enforce schema bounds at instantiation)
+        self._max_tokens_user_configured = max_tokens is not None
         if max_tokens is not None:
             self._validate_max_tokens(max_tokens)
-        self._default_max_tokens = max_tokens
+        self._default_max_tokens = max_tokens or DEFAULT_MAX_TOKENS
```
jrob5756: **BLOCKING**

`self._default_max_tokens = max_tokens or DEFAULT_MAX_TOKENS` removes the "leave it unset" path the constructor docstring used to describe. `_build_openai_model_settings` sets `max_tokens` whenever it isn't `None`, so every OpenAI request now carries it, and pydantic-ai maps that onto `max_completion_tokens`.

For OpenAI reasoning models — including Conductor's own default `gpt-5-mini` — `max_completion_tokens` covers reasoning tokens too. At `reasoning.effort: high`/`xhigh`/`max`, 16384 is routinely eaten by reasoning alone, producing `finish_reason: "length"` with empty visible content, which then shows up as a structured-output parse failure rather than a recognizable truncation. Before this PR the parameter was omitted and the server applied the model's full output allowance, so this is a silent regression for every existing OpenAI workflow that doesn't set `runtime.max_tokens`. It also now sends a parameter to arbitrary OpenAI-compatible `base_url` endpoints that never received one before — the PR body notes strict endpoints will "fail fast" from this, but doesn't mention the reasoning-token budget problem at all.

The compaction reserve doesn't need the parameter to actually go out on the wire — it just needs a number to reserve against, and `resolve_output_limit` already falls back to `DEFAULT_MAX_TOKENS` with source `"default"` when handed `None`.

Keep the wire behavior unchanged and supply the default only to the reserve calculation:

```python
self._default_max_tokens = max_tokens  # None when unconfigured; stays off the wire
```

`resolve_output_limit(effective_max_tokens=agent_max_tokens or self._default_max_tokens, ...)` will resolve `None` to `DEFAULT_MAX_TOKENS` internally. If always sending the parameter is genuinely intended, make that its own documented, opt-out-able change rather than a side effect of compaction, and pick a default well above typical reasoning budgets.

src/conductor/providers/claude.py
```diff
@@ -903,14 +1026,74 @@ async def execute(
...
+        effective_max_tokens = resolve_anthropic_effective_max_tokens(
+            agent,
+            default_max_tokens=self._default_max_tokens,
+            default_reasoning_effort=self._default_reasoning_effort,
+            default_model=self._default_model,
+        )
+        window = await resolve_compaction_window(
+            provider=self,
+            model=effective_model,
+            has_custom_base_url=self._base_url is not None,
```
jrob5756, line 1110: **RECOMMENDED**

`has_custom_base_url=self._base_url is not None` exists to keep the genai-prices registry from being consulted for a model id that describes someone else's deployment. On the Claude provider it doesn't work when the base URL comes from the environment.

`ClaudeProvider.__init__` stores `self._base_url = base_url`, and `_initialize_client` only passes `base_url` to `AsyncAnthropic` when it's not `None` — the `ANTHROPIC_BASE_URL` fallback the constructor docstring documents is handled by the Anthropic SDK itself, not by Conductor. So when someone points Conductor at a Databricks gateway or LiteLLM proxy through that env var, `self._base_url` stays `None`, `has_custom_base_url` reads `False`, and `resolve_compaction_window` queries the registry for the proxy's model id anyway — the exact case the flag was supposed to prevent.

It's also a parity gap: `openai.py` writes the environment value back at line 245 (`self._base_url = os.environ.get("OPENAI_BASE_URL")`), so the same deployment shape behaves correctly there and incorrectly here.

Mirror the OpenAI provider and fold the environment fallback into the instance attribute in `__init__`, so one attribute is the source of truth for both client construction and the registry gate:

```python
self._base_url = base_url if base_url is not None else os.environ.get("ANTHROPIC_BASE_URL")
```

Add a test that sets `ANTHROPIC_BASE_URL` with no explicit `base_url` and asserts `resolve_compaction_window` is called with `has_custom_base_url=True`.

jrob5756 (commented):
This adds always-on context compaction to the claude/openai providers, and the core metric it exists to report is broken: tokens_saved is structurally zero on every real compaction (b1), and the trigger/target math around it degenerates for a wide, realistic band of context windows, including the 128k fallback (b2, b3). There's also a silent behavior change in the OpenAI provider that can truncate reasoning-model output (b4). None of that is exotic to hit — the 56k-reserve default and the 128k fallback are both defaults, not edge cases.

Blocking, by location:
- `src/conductor/providers/_pydantic_ai/compaction.py:294` — tokens_after/tokens_saved always report zero reduction
- `src/conductor/providers/_pydantic_ai/compaction_window.py:296` — trigger collapses to 1 (compacts on every request) for windows ≤56,384
- `src/conductor/providers/_pydantic_ai/compaction_window.py:318` — the 55% hysteresis ceiling doesn't apply below ~125k, including the 128k fallback
- `src/conductor/providers/openai.py:227` — max_tokens=16384 is now always sent, truncating reasoning-model responses that used to get the server default

The other 19 are recommended: a base-URL env-var gap that defeats has_custom_base_url on Claude, unbounded model-listing pagination, a couple of unpinned/undeclared dependencies, dead code in the threshold gate, two places where compaction failure gets reported as success, process-global warning latches, a dropped partial model listing, a single-page bug in OpenAI's model listing, a missing get_max_output_tokens on Copilot, missing frontend styling for the new activity types, a docstring with the trigger formula backwards, and a handful of doc/example/changelog corrections (including one example that can't actually trigger the feature it's demonstrating) plus two test gaps a mutation check caught. All are inline.

**AGENTS.md:156**

**RECOMMENDED**

AGENTS.md is the repo's authoritative architecture doc, and line 156 enumerates the `_pydantic_ai` package member by member — `converters`, `events`, `mcp_toolset`, `agent_builder`, `interrupt`, `retry`, `structured_output`, `usage`. This PR adds three siblings totaling roughly 840 lines (`compaction.py`, `compaction_window.py`, `model_listing.py`) without touching that list.

The `#### openai.py parity notes` section also gets no new bullet, even though the OpenAI provider changed materially here: `get_max_prompt_tokens` went from a hardcoded `return None` to a real models-listing lookup, `max_tokens` went from "unset, server decides" to "always sent, defaults to 16384," and a vendor-field parser was added. The parity event list also doesn't record that three new event types now exist on only two of six providers — exactly the kind of declared exception that section already tracks, alongside the existing Hermes `max` and OpenAI `xhigh` carve-outs.

The one paragraph that was added is accurate as far as it goes, but it doesn't tell a maintainer which module owns what, that the capability stack is four layers deep and why each layer exists, that the disable latch is per agent execution and fires only on outer failures while tier failures currently get reported as successes, or that `execute_dialog_turn` is deliberately excluded from compaction (it keeps a fixed 4096 cap, so it's outside the "unified 16384 default" too).

Add the three modules to line 156, add an `openai.py` parity bullet covering the listing-derived limits and the always-sent `max_tokens`, note in the event-callback list that `agent_compaction_*` is claude/openai-only and why that's not a parity violation, and expand the compaction entry to match the density of the neighboring #397 and MCP entries.

src/conductor/providers/claude.py
```diff
@@ -459,35 +486,122 @@ def _connection_probe_verdict(self, exc: Exception) -> bool:
...
+        """
+        assert self._client is not None
+        result = self._client.models.list(limit=100)
+        page: Any = await result if inspect.isawaitable(result) else result
+        if AsyncPage is None or not isinstance(page, AsyncPage):
+            return self._single_page_model_infos(page)
+
+        models: list[Any] = []
+        try:
+            async for model in page:
```
jrob5756, line 584: **RECOMMENDED**

`async for model in page` drives the Anthropic SDK's auto-paginator with no page or item cap. `AsyncPage.has_next_page()` returns `True` whenever `has_more` isn't explicitly `False` and the page has items, and the cursor comes from `last_id`. An Anthropic-compatible gateway that ignores `after_id`, or that always reports `has_more: true`, loops forever here, appending to `models` without bound.

Before this PR only one page was fetched, so this is a new hang, and it's reachable from three entry points that each have only a per-request timeout, not an aggregate one: `validate_connection()` at workflow startup, `list_models()` from `conductor doctor`, and `_ensure_model_cache()` on the first agent execution. Proxies are precisely the deployment this PR adds `has_custom_base_url` handling for, so the risky configuration and the new code path show up together.

Bound the drain by item or page count and log once when the bound is hit:

```python
MAX_MODEL_LISTING_ITEMS = 2_000

async for model in page:
    models.append(model)
    if len(models) >= MAX_MODEL_LISTING_ITEMS:
        logger.warning(
            "models.list() exceeded %d entries; truncating the listing.",
            MAX_MODEL_LISTING_ITEMS,
        )
        break
```

pyproject.toml
```diff
@@ -39,7 +39,8 @@ dependencies = [
     "simpleeval>=1.0.0",
     "github-copilot-sdk>=1.0.9",
     "anthropic>=0.77.0,<1.0.0",
-    "pydantic-ai-slim[anthropic,openai]>=1.44.0",
+    "pydantic-ai-slim[anthropic,openai]>=2.28.0",
+    "pydantic-ai-harness>=0.22.0",
```
jrob5756: **RECOMMENDED**

`pydantic-ai-harness>=0.22.0` has no upper bound, and a 0.x package carries no semver guarantee — a 0.25 release can rename or restructure anything. The line above it also bumps `pydantic-ai-slim` across a major version (`>=1.44.0` -> `>=2.28.0`), also unbounded.

This repo already has an explicit convention against exactly this, visible in the same dependency block: `anthropic>=0.77.0,<1.0.0`, `openai>=2.48.0,<3.0.0`, and `mcp>=1.28.1,<2` with a comment spelling out that a major bump renames attributes and turns calls into runtime `AttributeError`s with no import-time signal.

The exposure here is bigger than a normal dependency because the new code reasons about the harness's private behavior. `_TierWrapper`'s docstring justifies its whole existence with "`TieredCompaction._escalate` calls each tier's `compact` directly and does not catch exceptions"; `build_tiered_compaction` passes `max_messages=1` as "the harness placeholder used when the tier is driven by an outer gate"; and the token measurement depends on the anchoring semantics of `estimate_context_tokens`. If upstream adds tier-level error handling or changes `_escalate`, this wrapper goes silently redundant or silently wrong.

Add upper bounds consistent with the rest of the file — `"pydantic-ai-harness>=0.22.0,<0.25"` and `"pydantic-ai-slim[anthropic,openai]>=2.28.0,<3"` — with a short comment naming the private behaviors relied on, mirroring the existing `mcp` comment. A test asserting the "`_escalate` doesn't catch" invariant would make a future bump fail loudly instead of quietly.

src/conductor/providers/_pydantic_ai/compaction_window.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+def _resolve_window_from_registry(model: str) -> int | None:
+    """Look up ``model`` in the genai-prices registry.
+
+    Returns the model's ``context_window`` when found, otherwise ``None``.
+    Any lookup failure is logged at debug level and returns ``None``.
+    """
+    try:
+        from genai_prices.data_snapshot import get_snapshot
```
jrob5756, line 142: **RECOMMENDED**

`from genai_prices.data_snapshot import get_snapshot` is the first runtime import of `genai_prices` anywhere in `src/` (`engine/pricing.py` only mentions it in comments), but `genai-prices` doesn't appear in `pyproject.toml` at all — it's present only as a transitive dependency of `pydantic-ai-harness` in `uv.lock`.

So Tier 2 of the window-resolution cascade rests on a package Conductor never actually asked for. If `pydantic-ai-harness` drops or renames that transitive dependency in a future release, the registry tier stops working invisibly: the `ImportError` at line 125 is swallowed and logged only at debug, and every model silently resolves to the 128k fallback labeled `source="fallback"`, indistinguishable from an endpoint that just doesn't advertise anything.

Declare it explicitly alongside the other direct dependencies, e.g. `"genai-prices>=0.1.4"`. Also promote the import failure from `logger.debug` to a warn-once latch so a vanished dependency reads differently from a model the registry simply doesn't know about.

src/conductor/providers/_pydantic_ai/compaction.py
```diff
@@ -0,0 +1,379 @@
...
+        self._trigger_tokens = trigger_tokens
+
+    async def before_model_request(
+        self,
+        ctx: RunContext[Any],
+        request_context: ModelRequestContext,
+    ) -> ModelRequestContext:
+        messages = list(request_context.messages)
+
+        anchor = _last_response_input_tokens(messages)
```
jrob5756: **RECOMMENDED**

`_ThresholdGatedCompaction.before_model_request` computes `anchor = _last_response_input_tokens(messages)` and branches on it, but the `if` and `else` bodies are character-for-character identical: both call `_estimate_context_tokens(...)` and both return early when `estimate <= trigger`. `anchor` gets computed and compared but never affects control flow, which makes `_last_response_input_tokens` (line 97) dead code in production. Both docstrings describe a pre-check that avoids running the full estimator; there is no such saving here.

On top of that, the gate is entirely redundant with its caller. `_FailOpenCompactionWrapper.before_model_request` already computes `before_estimate` over the same messages with the same `model_request_parameters`, evaluates the same predicate against the same `config.trigger_tokens`, and then delegates unconditionally. `estimate_context_tokens` is pure, so the gate's verdict provably always matches the caller's. The cost is a second full estimator pass on every model request of every agent run, and a third when compaction fires — right when the history is largest and the estimator is most expensive.

The two tests that look like they cover the anchor don't: their fixtures already exceed the trigger on the estimator alone, so deleting the anchor block leaves the whole test file green.

Delete `_ThresholdGatedCompaction` and `_last_response_input_tokens`, and gate inside the fail-open wrapper on the estimate it already has:

```python
before_estimate = await _estimate_context_tokens(
    before_messages, request_context.model_request_parameters
)
if before_estimate <= self._config.trigger_tokens:
    return request_context

self._on_before(estimate=before_estimate, messages_before=len(before_messages))
start = time.monotonic()
result = await self._inner.before_model_request(ctx, request_context)
```

`build_tiered_compaction` then wraps `TieredCompaction` directly. This is behavior-preserving, removes a class and a helper, halves the estimator work, and leaves one place where the trigger comparison lives.

src/conductor/providers/_pydantic_ai/compaction.py
```diff
@@ -0,0 +1,379 @@
...
+        self._tier_name = tier_name
+
+    async def compact(
+        self,
+        messages: list[ModelMessage],
+        ctx: RunContext[Any],
+    ) -> list[ModelMessage]:
+        try:
+            return await self._inner.compact(messages, ctx)
+        except Exception as exc:  # noqa: BLE001 - tier failure is recoverable
```
jrob5756, line 145: **RECOMMENDED**

`_TierWrapper.compact` swallows every tier exception and returns `messages` unchanged. `TieredCompaction._escalate` then measures no reclaim and moves to the next tier; if all three wrappers swallow a failure, it returns the original list and no exception ever reaches `_FailOpenCompactionWrapper`. Since `will_compact` was `True`, line 298 unconditionally emits `emit_compaction_complete(..., errored=False)`.

What the user sees for a total compaction failure is a dim `NOTE: context compacted for 'researcher': 190000 -> 190000 tokens` on the console, a `🧹 compacted` entry on the dashboard, and `"errored": false` in the JSONL log. The next model call then dies with a raw provider 400 for exceeding the context window — the exact failure this feature exists to prevent — while the structured log insists compaction succeeded. Anyone debugging this later reads `errored: false` and rules compaction out.

The same false-success shape happens whenever compaction runs but can't get below the trigger, and tier failures themselves only ever reach `logger.warning`, which this repo has already flagged elsewhere (the `_context_window_anomaly_warned` discussion) as not reliably reaching an operator.

Track degradation and report it. Have `_TierWrapper` record that it failed, and add `degraded_tiers` and `still_over_trigger` to the `agent_compaction_complete` payload:

```python
reduced = after_estimate < before_estimate
still_over = after_estimate > self._config.trigger_tokens
```

In `cli/run.py`, render `still_over_trigger or degraded_tiers` with `style="yellow"` and a `WARNING:` prefix instead of the neutral `NOTE`. A compaction that reclaimed nothing shouldn't render as an ordinary success.

src/conductor/providers/_pydantic_ai/compaction.py
```diff
@@ -0,0 +1,379 @@
...
+            )
+            self._on_after(
+                before_messages=before_messages,
+                after_messages=after_messages,
+                before_estimate=before_estimate,
+                after_estimate=after_estimate,
+                elapsed_seconds=time.monotonic() - start,
+            )
+            return result
+        except Exception as exc:  # noqa: BLE001 - compaction must never fail the run
```
jrob5756: **RECOMMENDED**

The single `try` starting at line 273 spans three unrelated concerns: the before-estimate, the inner strategy, and the after-estimate plus event emission at lines 293-305. The harness mutates the request context in place, so if `_estimate_context_tokens(after_messages, ...)` at line 294 raises, the history has already been compacted successfully — yet the handler logs "Compaction failed ... Continuing without compaction" (false; the compacted context is what gets returned), emits `errored=True` with the estimator's exception type (pointing the operator at the wrong component), and sets `self._disabled = True`.

That last part is the expensive one. One bad measurement on turn 3 disables compaction for turns 4..N with no further notice, and the run eventually dies on context overflow much later with no proximate signal.

Making this worse, the errored branch in `cli/run.py` only goes through `verbose_log`, so in a normal (non-verbose) run, a compaction failure — which puts the run at real risk of the exact overflow this feature exists to prevent — produces no console output at all. The console text also never mentions the most important fact: that compaction is now off for the rest of the run.

Scope the guards so telemetry can't latch off the feature:

```python
try:
    result = await self._inner.before_model_request(ctx, request_context)
except Exception as exc:          # genuine compaction failure -> latch + errored event
    ...
try:                              # measurement/telemetry only
    after_estimate = await _estimate_context_tokens(...)
    self._on_after(...)
except Exception:
    logger.warning("Compaction succeeded but post-measurement failed", exc_info=True)
return result
```

Also surface the errored branch outside verbose mode, and state plainly that compaction is disabled for the remainder of the run.

src/conductor/providers/_pydantic_ai/compaction_window.py
```diff
@@ -0,0 +1,318 @@
...
+#: pydantic-ai may run more than two parallel tool calls per segment, so this
+#: is a sizing heuristic, not a hard bound.
+TOOL_RESULT_COUNT_HEURISTIC = 2
+
+#: Extra token slack for the compaction estimator, tool schemas, and other
+#: non-result overhead. This is a heuristic, not a guarantee.
+ESTIMATOR_SLOP = 15_000
+
+#: One-shot warnings per process.
+_custom_base_url_warned = False
```
jrob5756, line 66: **RECOMMENDED**

`_custom_base_url_warned`, `_tool_output_disabled_warned`, and `_trigger_degenerate_warned` are module-level globals, set once and never reset. In any process running more than one workflow — `conductor mcp serve`, the fleet runner, the Fleet TUI, a test session — every run after the first silently loses these diagnostics, including the degenerate-trigger warning that's currently the only signal for the `trigger = 1` pathology.

AGENTS.md documents the established pattern here as a warn-once-per-run latch held on engine instance state (`_pricing_hook_failed_warned`, `_budget_unpriced_warned`, `_context_window_anomaly_warned`). Process-global latches also turn this into order-dependent state in the test suite: `tests/test_providers/test_compaction_window.py:294` already has to reach in and reset `cw._trigger_degenerate_warned = False` manually, while the two sibling tests asserting on the other two warnings don't reset theirs — so any earlier test that trips those latches will produce confusing order-dependent failures down the line.

Separately, `_warn_registry_lookup_failed` at line 90 is named like its three neighbors and documented as logging "a one-time debug message," but it has no latch at all and logs at debug — it's neither one-time nor a warning, and the visual symmetry with the others makes that easy to miss.

Move the latches onto per-execution state so they reset per run, and — since a bare `logger.warning` here effectively reaches `logging.lastResort` as unattributed stderr — emit the degenerate-trigger condition as a field on the `agent_compaction_config` event so it actually reaches the JSONL log and dashboard. Either give `_warn_registry_lookup_failed` a latch or rename it `_log_registry_lookup_failed` and drop "one-time" from its docstring.

src/conductor/providers/claude.py
```diff
@@ -459,35 +486,122 @@ def _connection_probe_verdict(self, exc: Exception) -> bool:
...
+
+        if self._client is None:
+            return
+
+        async with self._model_cache_lock:
+            if self._max_input_cache is not None or self._max_output_cache is not None:
+                return
+            try:
+                models = await self._fetch_all_model_infos()
+            except Exception as e:  # noqa: BLE001 - best-effort; don't poison the cache
```
jrob5756: **RECOMMENDED**

`_fetch_all_model_infos` deliberately raises `_PartialModelListingError` carrying the models retrieved before a later paginator page failed, and `validate_connection` handles that correctly — warns, then calls `_report_available_models(e.models)`. `_ensure_model_cache` throws all of it away instead: its broad `except Exception` catches `_PartialModelListingError` along with everything else, logs at debug, and returns without installing anything. The models it did paginate successfully get dropped, `get_max_prompt_tokens`/`get_max_output_tokens` return `None`, and compaction quietly falls back to the 128k guess. `list_models` has the identical problem and returns `None`.

This is the path taken whenever `validate_connection` wasn't run first — programmatic or embedded use, or a re-initialized provider after `close()` clears the caches. The same broad handler also swallows `AttributeError`/`TypeError` from `_single_page_model_infos` — a real schema-shape bug in Conductor's own parsing — as an indistinguishable debug line.

Related: `_PartialModelListingError` derives from `OSError` (line 72), which only meant something under the `except (TimeoutError, AnthropicError, OSError)` clause this PR deleted. Nothing catches `OSError` in either provider now, so the base class is a leftover that makes an internal control-flow signal look like an I/O error.

Handle the partial case explicitly, before the broad handler, in both `_ensure_model_cache` and `list_models`:

```python
except _PartialModelListingError as e:
    logger.warning(
        "Anthropic model listing incomplete (%s); using %d model(s) from completed pages",
        e.cause, len(e.models),
    )
    self._install_model_cache(e.models)
    return
```

Narrow the remaining handler to transport errors so a parser bug isn't mistaken for a network blip, and change the exception's base class to `Exception`.

src/conductor/providers/openai.py
```diff
@@ -393,24 +414,170 @@ def _connection_probe_verdict(self, exc: Exception) -> bool:
...
+
+        The OpenAI SDK returns an awaitable paginator. Awaiting it yields an
+        ``AsyncPage`` that is itself async-iterable, so we drain the page to
+        collect every model entry. Legacy single-page mocks that expose a
+        ``data`` attribute are still supported.
+        """
+        assert self._client is not None
+        result = self._client.models.list()
+        page: Any = await result if inspect.isawaitable(result) else result
+        data: Any = getattr(page, "data", None)
```
jrob5756: **RECOMMENDED**

`data: Any = getattr(page, "data", None)` is checked first and, when non-None, returns `list(data)` immediately. `openai.pagination.AsyncPage` always declares `data: List[_T]`, so the real SDK takes this branch on every call, and the `async for model in page` drain below it never runs in production. The docstring claims the opposite — that the code drains every page and only falls back to `.data` for "legacy single-page mocks" — which inverts which path is actually live.

This is also a parity gap introduced by this PR: `claude.py` checks `isinstance(page, AsyncPage)` first and genuinely drains, so the two pydantic-ai providers now differ in listing completeness. Against an OpenRouter/LiteLLM/vLLM gateway that paginates `/v1/models` — exactly the deployments `model_listing.py` was added to read vendor fields from — every model past page one silently doesn't exist for context-window resolution or for `conductor doctor`, and the resulting `None` reports as `source="fallback"`, indistinguishable from an endpoint that advertises nothing at all.

Also worth tightening: an empty or garbage listing installs `_model_limits_cache = {}`, which is `is not None`, so `_ensure_model_limits_cache` treats emptiness as a resolved answer for the process lifetime and never retries. A proxy returning `{"data": []}` while warming up pins the run to the fallback permanently.

Mirror the Claude implementation: check for a real paginator (`has_next_page` / `isinstance(page, AsyncPage)`) and drain it, keeping the `.data` shortcut only for the legacy-mock branch, and fix the docstring either way. Treat an empty listing as unresolved rather than cached — don't install it, and log once that the endpoint returned no models.

src/conductor/providers/base.py
```diff
@@ -545,6 +545,32 @@ async def get_max_prompt_tokens(self, model: str) -> int | None:
         """
         return None
 
+    async def get_max_output_tokens(self, model: str) -> int | None:
```
jrob5756, line 548: **RECOMMENDED**

This PR adds `AgentProvider.get_max_output_tokens()` alongside the existing `get_max_prompt_tokens()`. The Copilot provider already reads `capabilities.limits.max_output_tokens` from its SDK and surfaces it through `get_model_capabilities`, but it doesn't override the new hook, so it returns the base `None`.

AGENTS.md makes provider parity a hard rule — any change to one provider's contract applies to all providers where applicable. Unlike compaction itself, which is legitimately pydantic-ai-only, this hook is provider-agnostic model metadata, and Copilot is sitting on the exact value it needs. Leaving it unimplemented means the provider with the best metadata of the bunch silently returns `None` to any future consumer of this hook.

Implement `get_max_output_tokens` on `copilot.py` by delegating to the same `capabilities.limits` lookup `get_model_capabilities` already does, and record in AGENTS.md which providers legitimately return `None` (hermes, aca, claude-agent-sdk) so the gap is declared instead of accidental.

src/conductor/web/frontend/src/stores/workflow-store.ts
```diff
@@ -283,7 +283,7 @@ export interface HighlightedEdge {
 
 export type LogLevel = 'info' | 'success' | 'error' | 'warning' | 'debug';
 
-export type ActivityLogType = 'reasoning' | 'tool-start' | 'tool-complete' | 'turn' | 'message' | 'prompt' | 'parse-recovery';
+export type ActivityLogType = 'reasoning' | 'tool-start' | 'tool-complete' | 'turn' | 'message' | 'prompt' | 'parse-recovery' | 'compaction-config' | 'compaction-start' | 'compaction-complete' | 'compaction-error';
```
jrob5756, line 286: **RECOMMENDED**

`ActivityLogType` gains `'compaction-config' | 'compaction-start' | 'compaction-complete' | 'compaction-error'`, and the store emits all four, but neither consumer style map was updated. `ACTIVITY_TYPE_STYLES` in `components/layout/OutputPane.tsx` is looked up as `ACTIVITY_TYPE_STYLES[entry.type] || ACTIVITY_TYPE_STYLES.message`, and `typeColors` in `components/detail/ActivityStream.tsx` falls through to `text-[var(--text)]`.

The result is that `compaction-error` — the one entry actually signaling a failure — renders in the activity firehose with the neutral `MSG` label and default text color, visually identical to ordinary agent chatter. The `⚠️` in the message text is the only differentiator, and it's neither filterable nor color-scannable, unlike `parse-recovery`, which this codebase deliberately styles yellow with a `RETRY` label.

Separately, `tokens_saved` is computed by the backend and declared in `events.ts`, but nothing on the frontend actually reads it — arguably the single most operator-relevant number in the whole event.

Add entries to both maps, e.g. `'compaction-error': { color: 'text-yellow-400', label: 'COMPACT', labelColor: 'text-yellow-600' }` plus dim/neutral entries for the other three so they stop reading as `MSG`, and mirror `'compaction-error': 'text-yellow-400'` into `typeColors`. Either render `tokens_saved` somewhere or drop it from both the emitter and the type.

src/conductor/providers/_pydantic_ai/compaction.py
```diff
@@ -0,0 +1,379 @@
+"""Tiered context-window compaction for Pydantic AI agents.
+
+This module assembles the harness's :class:`TieredCompaction` into a conductor
+wrapper that:
+
+1. Gates compaction on a reserve-based token trigger (window minus max(output
```
jrob5756: **RECOMMENDED**

Line 6 describes the gate as "a reserve-based token trigger (window minus max(output limit, buffer))", but `compaction_window.trigger_tokens` actually computes `max(1, window - output_limit - tool_buffer)` — the two reserves are summed, not maxed. Every other artifact in this PR gets it right (AGENTS.md, docs/workflow-syntax.md, docs/providers/claude.md, docs/providers/openai.md, the CHANGELOG, and `trigger_tokens`'s own docstring), which makes this the one place a reader could reasonably "fix" the code toward the wrong formula.

At the default 200k window with a 16384 output limit and a 40000 buffer, that's the difference between a 143,616 trigger and a 160,000 trigger — not a rounding error. This is the header of the feature's core module, so it's the first thing anyone touching this code reads.

Change it to "(window minus the output limit minus the tool buffer)" so the module docstring matches `trigger_tokens`.

docs/configuration.md
```diff
@@ -433,7 +433,7 @@ workflow:
 
 **Limits**:
 - Haiku: 4096 max
-- Sonnet/Opus: 8192 max
+- Sonnet/Opus: capped at runtime by `ModelInfo.max_tokens` from the provider, supporting up to 16384 tokens
```
jrob5756: **RECOMMENDED**

The line reads "Sonnet/Opus: capped at runtime by `ModelInfo.max_tokens` from the provider, supporting up to 16384 tokens." Both halves are wrong.

`get_max_output_tokens()` has exactly one consumer — `resolve_output_limit` in `compaction_window.py` — and it feeds the compaction reserve arithmetic only. The value actually sent to the API comes from `_build_anthropic_model_settings` via `resolve_anthropic_effective_max_tokens`, which never looks at provider metadata at all. So `runtime.max_tokens: 60000` on Haiku still goes out verbatim and still returns a 400. And 16384 is Conductor's default when the field is omitted, not a supported maximum — Sonnet 4.5's real cap is 64000, so this understates the model by 4x and reads like a hard ceiling that doesn't exist.

The same mistake appears in docs/providers/claude.md, including a table cell that presents Conductor's default as if it were the model's capability.

Replace with something like: "Conductor defaults `max_tokens` to 16384 when you don't set it. Conductor doesn't clamp your configured value to the model's advertised cap — a value above the model limit is rejected by the API. The provider-advertised `ModelInfo.max_tokens` is used only to size the compaction output reserve." Restore the real per-model caps in the claude.md table, and fix the same misstatement in docs/providers/migration.md, which still calls `max_tokens` "Required."

docs/workflow-syntax.md
```diff
@@ -2338,6 +2338,100 @@ agent, without validating anything else.
...
+Compaction operates in a fail-open manner. If an error occurs during compaction, Conductor logs a warning, disables compaction for the rest of that agent's execution, and continues with the uncompacted history.
+
+Conductor emits three event types to track compaction:
+*   `agent_compaction_config`: Emitted once at the start of agent execution to log resolved window and limit values.
+*   `agent_compaction_start`: Emitted when context size exceeds the trigger threshold and compaction begins.
+*   `agent_compaction_complete`: Emitted when compaction completes, detailing token savings or errors.
+
+### Dashboard Caveat
+
+The web dashboard's context remaining bar estimates context size using only provider-supplied model limits. It might disagree with the actual compaction window, especially under proxy configurations. The bar updates to show post-compaction usage after a compaction event completes.
```
jrob5756: **RECOMMENDED**

The last sentence of the Dashboard Caveat reads "The bar updates to show post-compaction usage after a compaction event completes." In `workflow-store.ts`, `context_pct` is written in exactly three places: reset on `agent_started`, and computed from `context_window_used`/`context_window_max` on `agent_completed` and `parallel_agent_completed`. The three new handlers this PR adds only append activity-log entries — none of them touches `context_pct`, `context_window_used`, or `context_window_max`.

So the bar doesn't move at compaction time. It moves once, when the agent step finishes. The first two sentences of this paragraph hold up; this one doesn't, and it's the sentence a user would actually rely on to interpret what they're seeing mid-run.

Delete the sentence, or replace it with: "The bar is refreshed only when the agent step completes; a mid-execution compaction shows up in the activity log, not in the bar."

examples/compaction.yaml
```diff
@@ -0,0 +1,69 @@
+# Context Compaction Workflow Example
+#
+# This workflow demonstrates automatic client-side context compaction.
+# It uses a loop-back agent pattern to accumulate conversation history.
+# As history grows and crosses the trigger threshold, Conductor will
+# automatically compact older tool results and messages.
+#
+# There is no user-facing context window override. Resolution uses provider
+# metadata, the registry, or the fallback. For testing, compaction triggers
```
jrob5756: **RECOMMENDED**

Two problems in the file users will actually copy from.

First, line 9 says compaction "triggers early when configured with a small `runtime.max_tokens` value." Since `trigger = window - output_limit - tool_buffer`, a smaller `max_tokens` produces a larger trigger, so compaction fires later, not earlier. This PR's own docs/workflow-syntax.md has the direction right — this comment doesn't. Anyone following it to reproduce compaction gets the opposite of what they're trying to do.

Second, the example's premise doesn't hold up at all. It uses `context.mode: accumulate` plus a self-routing agent to "accumulate conversation history," but compaction operates on the pydantic-ai message history inside a single `provider.execute()` call, and Conductor never passes `message_history` on the initial run. Each loop iteration starts a fresh agent execution with a two-message history; `context.mode` only controls which prior outputs Jinja templates can see. The prompt only references the previous iteration's output, which gets overwritten every loop, so prompt size stays roughly constant. There are also no tools in the example, so tier 1 (`ClearToolResults`) has nothing to clear. This PR's own docs/workflow-syntax.md states that provider history doesn't accumulate across loop-back iterations — so the example contradicts a doc added in the same PR, and the same misconception shows up again in plugins/conductor/skills/conductor/references/authoring.md.

Fix the direction of the sentence ("a large `runtime.max_tokens` lowers the trigger and makes compaction fire sooner"), and rebuild the example around a single agent with MCP tools returning large payloads across a multi-turn tool loop — that's what actually grows the pydantic-ai message history — dropping the loop-back/`mode: accumulate` framing entirely. Fix the same passage in authoring.md for the same reason.

tests/test_providers/test_compaction_window.py
```diff
@@ -0,0 +1,411 @@
...
+            result = trigger_tokens(128_000, 16_384, 265_000)
+
+        assert result == 1
+        assert "degenerated" in caplog.text.lower()
+
+
+class TestTargetFormula:
+    """Tests for target_tokens math."""
+
+    def test_small_window_clamps_below_trigger(self) -> None:
```
jrob5756, line 441: **RECOMMENDED**

Two of this PR's headline behaviors have no effective coverage — confirmed by mutating the source and re-running.

The `trigger - 1` term in `target_tokens` is the entire hysteresis guarantee, but none of the three `TestTargetFormula` cases reach it: `test_small_window_clamps_below_trigger` computes `min(70_400, 71_615)`, so the 55% fraction wins and the clamp never fires (the test's own name asserts a clamp that doesn't happen), `test_large_window_uses_fraction` is the same story, and `test_floor` only exercises the `max(1, ...)` floor. Removing `- 1` from the source and running `tests/test_providers/` still gives 1301 passed, 13 skipped.

Same problem with the anchor short-circuit: replacing the entire anchor/if/else block in `_ThresholdGatedCompaction` with a single unconditional estimate leaves all 24 tests in `test_pydantic_ai_compaction.py` green. `test_anchor_short_circuit_fires_trigger` passes for an unrelated reason — its fixture estimates to 511 tokens against a trigger of 100, so the estimator alone decides the outcome regardless of the anchor.

Also worth tightening: `test_pydantic_ai_compaction.py:768` asserts `tokens_saved == max(0, tokens_before - tokens_after)`, which just recomputes the emitter's own expression from the same payload. It holds for any internally consistent implementation, including one that reports zero saved on every compaction — which is what currently happens.

Add a case where the reserve is large enough for the clamp to win, and assert the exact value plus the strict inequality:

```python
trigger = trigger_tokens(128_000, 64_000, 40_000)   # 24_000
assert target_tokens(128_000, trigger) == 23_999    # clamp beats the 70_400 ceiling
```

For the anchor, patch `_estimate_context_tokens` with an `AsyncMock` and assert `assert_not_awaited()` on the short-circuit path, so removing the anchor actually fails a test. Assert `tokens_saved > 0` on a compacting fixture instead of recomputing the formula. It's also worth adding a test that two concurrently built compaction stacks don't share the `_disabled` latch or each other's event callbacks, and a parity test confirming copilot/claude-agent-sdk/hermes/aca emit no `agent_compaction_*` events.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/microsoft/conductor/compare/v0.1.35...HEAD)
 
+### Added
+
+- **Automatic client-side context compaction for pydantic-ai providers** (`claude` and `openai`). Conductor automatically condenses conversation history when it crosses a calculated trigger threshold. This threshold is computed using an additive reserve formula, keeping headroom for the output limit and a tool-output-derived buffer. Additionally, this update unifies the default `max_tokens` to 16384 for both providers and removes the old 64k fallback. The `openai` provider now always sends the `max_tokens` parameter to the API and honors vendor-advertised token limits from models listing when available. For Claude, the model listing from the Anthropic SDK is now fully paginated. For Claude thinking agents, `low` or `medium` effort levels without an explicit `max_tokens` limit now send 16384 tokens instead of the former 8192 or 12288 tokens. To condense history, the compaction process runs a three-stage strategy: clearing old tool results first, summarizing older messages with a nested model call, and sliding the window as a deterministic fallback.
```
jrob5756: **RECOMMENDED**

This is a single `### Added` bullet with no `### Changed` section, and that undersells three things.

Claude's default `max_tokens` moves from 8192 to 16384, doubling the worst-case output cost per call for every existing user who never set it explicitly. That belongs under `Changed`, not tucked inside `Added`.

The OpenAI change is more consequential and reads like a neutral fact ("now always sends the `max_tokens` parameter"). Previously the parameter was left unset and the server applied its own, usually much larger, default. An existing workflow that used to produce a 40k-token response now silently truncates at 16384, and on reasoning models that cap gets eaten by reasoning tokens before it reaches the answer. There's no warning about any of that and no remedy named.

Last, "removes the old 64k fallback" describes something users never actually had — `git grep` on `origin/main` finds no 64k fallback in `src/conductor/providers` outside the extended-thinking output caps, which this PR keeps unchanged. That line looks like a leftover from the branch's own development history.

Split this into `### Added` (compaction) and `### Changed` (the defaults). State the Claude 8192->16384 and OpenAI unset->16384 changes explicitly, with the truncation/cost implications and the remedy (`runtime.max_tokens`), and delete or correct the 64k-fallback line.

src/conductor/providers/_pydantic_ai/compaction.py
hertznsk: Confirmed and fixed. The post-compaction estimate now compensates for the usage anchor the same way `TieredCompaction._escalate` does internally: `_estimate_after_compaction_tokens` computes the heuristic reclaim (`estimate_token_count(before) - estimate_token_count(after)`) and subtracts it from the anchored before-estimate, so `tokens_after`/`tokens_saved` reflect the real reduction. A regression test with a usage-carrying `ModelResponse` in the retained tail asserts `tokens_saved > 0`. (a7551d0)

src/conductor/providers/_pydantic_ai/compaction_window.py
hertznsk: Confirmed and fixed. The `max(1, ...)` floor is gone from the live path: `resolve_compaction_plan()` treats a reserve that doesn't fit the window as *don't arm compaction* — `trigger < MIN_VIABLE_TRIGGER (4096)` (or a target below 1) returns a disabled plan with `disabled_reason: "insufficient_headroom"`, both providers pass `compaction=None`, and `emit_compaction_config` reports `enabled: false` with the reason so the state is visible per run. The tool buffer is also clamped to 25% of the window so an unbounded `max_chars` can't exceed it. (a7551d0)

src/conductor/providers/_pydantic_ai/compaction_window.py
hertznsk: Confirmed and fixed. The `trigger - 1` clamp is replaced by a window-scaled hysteresis margin: `target = min(int(window * TARGET_FRACTION), trigger - max(1, int(window * 0.05)))`, and `0 < target < trigger` is now an asserted invariant of an enabled plan (plans that can't satisfy it are disabled per the trigger_tokens finding). Tests exercise the clamp branch directly and assert the gap. (a7551d0)

src/conductor/providers/openai.py
hertznsk: Confirmed and fixed. `self._default_max_tokens = max_tokens` again — the parameter stays off the wire when unset (verified: `_build_openai_model_settings` only sends it when not None), while `resolve_output_limit` still falls back to the 16384 default internally, so the reserve math is unchanged. The constructor docstring and the affected tests were updated. The CHANGELOG no longer claims the always-send behavior. (a7551d0)

src/conductor/providers/claude.py
hertznsk: Fixed. `ClaudeProvider.__init__` now folds the env fallback into the attribute: `self._base_url = base_url if base_url is not None else os.environ.get("ANTHROPIC_BASE_URL")`, mirroring openai.py, so one attribute is the source of truth for both client construction and the registry gate. A test sets `ANTHROPIC_BASE_URL` with no explicit base_url and asserts `resolve_compaction_window` is called with `has_custom_base_url=True`. (a7551d0)

src/conductor/providers/claude.py
hertznsk: Fixed. The drain is bounded by `MAX_MODEL_LISTING_ITEMS = 2_000` (module constant documenting the proxy-hang scenario); hitting the bound logs one warning and truncates. A test with an endless paginator double (has_more always true) asserts truncation at the cap. (a7551d0)

pyproject.toml
hertznsk: Fixed. `pydantic-ai-harness>=0.22.0,<0.25` and `pydantic-ai-slim[anthropic,openai]>=2.28.0,<3`, with a comment naming the private behaviors relied on (`_escalate` not catching tier exceptions, the `max_messages=1` placeholder, `estimate_context_tokens` anchoring), mirroring the existing `mcp` comment. `uv.lock` refreshed. (a7551d0)

src/conductor/providers/_pydantic_ai/compaction_window.py
hertznsk: Fixed. `genai-prices>=0.1.4` is now a direct dependency in pyproject.toml with a comment noting it was previously transitive-only. On the warn-once promotion: the import failure log was renamed to `_log_registry_lookup_failed` (dropping the misleading "one-time" docstring) rather than latched to warning, since the registry tier is explicitly best-effort and per-model lookup failures share the path — the dependency declaration is the real fix for the invisible-degradation concern. (a7551d0)

src/conductor/providers/_pydantic_ai/compaction.py
hertznsk: Confirmed and fixed. `_ThresholdGatedCompaction` and `_last_response_input_tokens` are deleted; `build_tiered_compaction` wraps `TieredCompaction` directly in `_FailOpenCompactionWrapper`, which now gates on its own `before_estimate <= trigger` before delegating — one estimator pass per request. The gating tests were rewritten against the wrapper (patching `_estimate_context_tokens` so removing the gate fails a test). (a7551d0)

src/conductor/providers/_pydantic_ai/compaction.py
hertznsk: Fixed. `_TierWrapper` now records its failure on a `failed` flag; the wrapper collects degraded tier names and adds `degraded_tiers` and `still_over_trigger` to the `agent_compaction_complete` payload. In `cli/run.py`, a complete event with either set renders a yellow `WARNING:` line naming the tiers / the above-trigger state instead of the neutral NOTE, and the dashboard styles them as `compaction-error` entries. (a7551d0)

src/conductor/providers/_pydantic_ai/compaction.py
hertznsk: Confirmed and fixed. The wrapper now has three failure zones: gate-measurement failure warns and skips compaction for that request only (no latch, no event); inner-strategy failure keeps the old behavior (errored event + per-execution latch); after-telemetry failure warns, returns the *compacted* result, and leaves the latch off. The errored branch in cli/run.py now states plainly that compaction is disabled for the remainder of the run. Tests cover the telemetry-failure path including that a subsequent above-trigger request still compacts. (a7551d0)

src/conductor/providers/_pydantic_ai/compaction_window.py
hertznsk: Partially taken. The degenerate-trigger latch is gone entirely — that condition is now a per-run `enabled: false` + `disabled_reason` field on `agent_compaction_config`, which reaches the JSONL log and dashboard. `_warn_registry_lookup_failed` was renamed `_log_registry_lookup_failed` with the "one-time" claim dropped. The remaining two latches stay module-scoped deliberately: their call sites (`resolve_compaction_window`, `tool_buffer_tokens`) are pure functions with no execution state to hang a flag on, and the warnings describe process-wide configuration. Order-dependence in tests is solved uniformly via a `_reset_warning_latches()` helper used by an autouse fixture — the sibling tests that didn't reset now do. (a7551d0)

src/conductor/providers/claude.py
hertznsk: Fixed. `_ensure_model_cache` and `list_models` now handle `_PartialModelListingError` explicitly before the broad handler (warn + install the completed pages' models), the remaining catch is narrowed to `(APIConnectionError, APIStatusError, TimeoutError)` so parser bugs surface, and the exception now derives from `Exception` with a docstring explaining why it is not an OSError. An empty successful listing is also no longer cached as a resolved answer. (a7551d0)

src/conductor/providers/openai.py
hertznsk: Fixed. `_fetch_model_infos` now checks `isinstance(page, AsyncPage)` first and genuinely drains the paginator (bounded by `MAX_MODEL_LISTING_ITEMS = 2_000` with a warning); the `.data` shortcut only serves legacy single-page mocks, and the docstring matches reality. An empty listing is not installed — the cache stays None so the next call retries, and the same guard was applied to the `validate_connection` probe path. A multi-page test fails on the old code and passes on the new. (a7551d0)

src/conductor/providers/base.py
hertznsk: Fixed. `CopilotProvider.get_max_output_tokens` delegates to the same `get_model_capabilities` limits lookup, placed beside the existing `get_max_prompt_tokens` override. AGENTS.md's parity section now records that `claude`/`openai`/`copilot` implement both hooks while `hermes`, `aca`, and `claude-agent-sdk` legitimately return `None` from the default. (a7551d0)

src/conductor/web/frontend/src/stores/workflow-store.ts
hertznsk: Fixed. `ACTIVITY_TYPE_STYLES` (OutputPane) gains neutral COMPACT entries for compaction-config/start/complete and a yellow one for compaction-error; `typeColors` (ActivityStream) mirrors the yellow. `tokens_saved` is now rendered in the compaction-complete text (`saved N tokens`), and the store handles the new payload fields (disabled config, degraded tiers / still-over-trigger render with the warning shape). Store tests extended; static assets rebuilt. (a7551d0)

src/conductor/providers/_pydantic_ai/compaction.py
hertznsk: Fixed — the module docstring now reads "(window minus the output limit minus the tool buffer)", matching `trigger_tokens`. (a7551d0)

docs/configuration.md
hertznsk: Fixed. docs/configuration.md now states Conductor defaults `max_tokens` to 16384 when unset, does not clamp the configured value to the model's advertised cap (a value above the model limit is rejected by the API), and that `ModelInfo.max_tokens` is used only to size the compaction output reserve. The claude.md table and the same misstatement in its config/troubleshooting sections were corrected, and migration.md no longer calls `max_tokens` "Required." (a7551d0)

docs/workflow-syntax.md
hertznsk: Fixed — replaced with: "The bar is refreshed only when the agent step completes; a mid-execution compaction shows up in the activity log, not in the bar." (a7551d0)

examples/compaction.yaml
hertznsk: Fixed both. The comment direction is corrected (a *large* `runtime.max_tokens` lowers the trigger and fires compaction sooner), and the example is rebuilt around a single researcher agent with an open-websearch MCP server running a multi-turn tool loop — the thing that actually grows the pydantic-ai message history within one `execute()`. The loop-back/`mode: accumulate` framing is dropped, the header now states explicitly that loop-back iterations do not accumulate provider history, and the same passage in authoring.md is corrected. `conductor validate` passes. (a7551d0)

tests/test_providers/test_compaction_window.py
hertznsk: Fixed. `TestResolveCompactionPlan` exercises the clamp branch with a large reserve and asserts the exact target plus the strict inequality; the anchor-block tests are gone with the dead gate (the wrapper gate is now tested by patching `_estimate_context_tokens` so its removal fails); the complete-event test asserts `tokens_saved > 0` on a compacting fixture with a usage anchor in the retained tail; and a test confirms two concurrently built stacks share neither the `_disabled` latch nor event callbacks. The `test_sdk_error_returns_none` test in test_list_models.py was also aligned with the intentional narrowing (transport errors return None; unexpected errors raise). (a7551d0)

CHANGELOG.md
hertznsk: Fixed. The entry is split into `### Added` (compaction, provider-advertised token-limit metadata incl. pagination, the copilot hook) and `### Changed` (Claude default 8192→16384 with the cost implication and the `runtime.max_tokens` remedy; thinking low/medium 8192/12288→16384). The "removes the old 64k fallback" phrase is deleted, and the "now always sends max_tokens" clause is gone since that behavior was reverted. (a7551d0)

jrob5756 (approved):

PR #508, opened 2026-09-07
fix(providers): openai backend uses PromptedOutput instead of ToolOutput for structured output

josephrobertlopez:
## Summary

- Several widely-used local models served behind an OpenAI-compatible endpoint (e.g. Ollama) do not reliably emit a real function/tool call for structured output, even when explicitly requested. Under `ToolOutput`, any agent with a non-empty `output:` schema against such a model exhausted its parse-recovery budget and raised `UnexpectedModelBehavior: Exceeded maximum output retries`, regardless of schema size or retry configuration.
- Switches `_build_output_type` to return `PromptedOutput` instead of `ToolOutput` when `backend == "openai"`. The `claude`/Anthropic backend is untouched (still `ToolOutput`, since real tool-calling is reliable there).
- Matches known upstream reports: [ollama/ollama#8095](https://github.com/ollama/ollama/issues/8095) ("structured output with tools always produces empty tool_calls array"), [ollama/ollama#8063](https://github.com/ollama/ollama/issues/8063), [pydantic/pydantic-ai#877](https://github.com/pydantic/pydantic-ai/issues/877) ("Ollama can't return structured output").

## Known limitation (documented, not fixed here)

`ToolOutput`'s rendered tool schema is sanitized (`$ref`/`$defs` inlined, pydantic-internal keys stripped) before being sent to the model. `PromptedOutput`'s current public API (`outputs`/`name`/`description`/`template`) has no equivalent hook, so it renders the raw `model_json_schema()` for a **nested** output schema — `$ref`/`$defs` included. This can make a nested `output:` schema harder for a weak local model to satisfy than a flat one, partially working against this fix's own goal. A regression test pins this behavior as a concrete target for a follow-up rather than silently shipping an overclaim.

**Workaround, until a follow-up fix lands:** keep each `openai`-backend agent step's `output:` schema flat, and reassemble any nested structure in a downstream `type: script` (or `type: set`) step instead — nesting is pure data shaping and needs no LLM call. This is also the same decomposition pattern that improves parse-recovery reliability on weak local models generally, so it's a good default shape for a local-model-backed workflow regardless of this specific limitation.

## Test plan

- [x] `uv run pytest tests/test_providers/test_pydantic_ai_agent_builder.py -q` — 49 passed (includes new tests for both the flat-schema fix and the pinned nested-schema limitation)
- [x] `uv run pytest tests/test_providers/ -q` — 1567 passed, 1 skipped, 1 pre-existing unrelated failure (`test_copilot.py::TestLogRecoveryAttempt::test_omits_tag_when_unnamed`, confirmed failing identically on a clean `origin/main` checkout before this change)
- [x] `uv run ruff check` / `uv run ruff format --check` — clean
- [x] `uv run ty check src/conductor/providers/_pydantic_ai/agent_builder.py` — clean
- [x] Real end-to-end verification: ran an actual conductor workflow (a flat output schema) against `qwen2.5-coder:14b-8k` over a real Ollama endpoint — failed consistently with `ToolOutput` (`UnexpectedModelBehavior: Exceeded maximum output retries`) before this change, passed cleanly after
- [x] Adversarially reviewed for other call sites assuming `ToolOutput` on the openai backend, retry-semantics changes, and unintended effects on `hermes.py`/`aca.py` — none found; both confirmed to not import `agent_builder`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,10 +42,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   regardless of schema size or retry configuration. `PromptedOutput` asks
   the model to emit the schema as JSON in its normal text response instead,
   which these backends handle correctly; verified via a real end-to-end
-  workflow run against `qwen2.5-coder:14b-8k` over Ollama, which failed
-  consistently before this change and passed cleanly after. Scoped to the
-  `openai` backend only — the `claude`/Anthropic backend keeps `ToolOutput`,
-  since real tool-calling is reliable there.
+  workflow run against `qwen2.5-coder:14b-8k` over Ollama with a **flat**
+  output schema, which failed consistently before this change and passed
+  cleanly after. Scoped to the `openai` backend only — the `claude`/
+  Anthropic backend keeps `ToolOutput`, since real tool-calling is reliable
+  there.
+
+  **Known limitation, not fixed by this change:** `ToolOutput`'s rendered
+  tool schema is sanitized (`$ref`/`$defs` inlined, pydantic-internal keys
+  stripped) before being sent to the model; `PromptedOutput` has no
+  equivalent hook in its current public API and renders the raw
+  `model_json_schema()` — `$ref`/`$defs` included — for a **nested** output
+  schema. This can make a nested `output:` schema harder for a weak local
+  model to satisfy than a flat one, partially working against this fix's
+  own goal. `tests/test_providers/test_pydantic_ai_agent_builder.py::TestOpenAIBackend::test_openai_nested_output_schema_is_not_ref_inlined_known_limitation`
+  pins the current behavior as a concrete target for a follow-up fix.
 
 ## [0.1.36](https://github.com/microsoft/conductor/compare/v0.1.35...v0.1.36) - 2026-09-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   entirely. See
   [`examples/claude-agent-sdk-setting-sources.yaml`](examples/claude-agent-sdk-setting-sources.yaml).
 
+### Fixed
+
+- **`openai` provider structured output uses `PromptedOutput` instead of
+  `ToolOutput`** — several widely-used local models served behind an
+  OpenAI-compatible endpoint (e.g. Ollama) do not reliably emit a real
+  function/tool call for structured output, even when explicitly requested
+  (see [ollama/ollama#8095](https://github.com/ollama/ollama/issues/8095),
+  [ollama/ollama#8063](https://github.com/ollama/ollama/issues/8063),
+  [pydantic/pydantic-ai#877](https://github.com/pydantic/pydantic-ai/issues/877)).
+  Under the previous `ToolOutput` wrapping, any agent with a non-empty
+  `output:` schema against such a model exhausted its parse-recovery budget
+  and raised `UnexpectedModelBehavior: Exceeded maximum output retries`,
+  regardless of schema size or retry configuration. `PromptedOutput` asks
+  the model to emit the schema as JSON in its normal text response instead,
+  which these backends handle correctly; verified via a real end-to-end
+  workflow run against `qwen2.5-coder:14b-8k` over Ollama, which failed
+  consistently before this change and passed cleanly after. Scoped to the
+  `openai` backend only — the `claude`/Anthropic backend keeps `ToolOutput`,
+  since real tool-calling is reliable there.
+
 ## [0.1.36](https://github.com/microsoft/conductor/compare/v0.1.35...v0.1.36) - 2026-09-02
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   own goal. `tests/test_providers/test_pydantic_ai_agent_builder.py::TestOpenAIBackend::test_openai_nested_output_schema_is_not_ref_inlined_known_limitation`
   pins the current behavior as a concrete target for a follow-up fix.
 
+  **Workaround, until that follow-up lands:** avoid a nested `output:`
+  schema on an `openai`-backend agent step entirely. Have each agent step
+  return a flat schema, then reassemble the nested structure in a
+  downstream `type: script` (or `type: set`) step — nesting is pure data
+  shaping and needs no LLM call, so it never hits this limitation. This is
+  also the same decomposition pattern that improves parse-recovery
+  reliability on weak local models generally (smaller, flatter one-shot
+  generations retry more predictably than one large structured call), so
+  it is a good default shape for a local-model-backed workflow regardless
+  of this specific limitation. See `examples/` (a worked example is a good
+  follow-up addition here).
+
 ## [0.1.36](https://github.com/microsoft/conductor/compare/v0.1.35...v0.1.36) - 2026-09-02
 
 ### Added
--- a/src/conductor/providers/_pydantic_ai/agent_builder.py
+++ b/src/conductor/providers/_pydantic_ai/agent_builder.py
@@ -20,7 +20,7 @@ from openai import AsyncOpenAI
 from pydantic_ai import Agent, AgentRetries
 from pydantic_ai.models.anthropic import AnthropicModel, AnthropicModelSettings
 from pydantic_ai.models.openai import OpenAIChatModel, OpenAIChatModelSettings
-from pydantic_ai.output import ToolOutput
+from pydantic_ai.output import PromptedOutput, ToolOutput
 from pydantic_ai.providers.anthropic import AnthropicProvider
 from pydantic_ai.providers.openai import OpenAIProvider
 
@@ -215,19 +215,34 @@ def _resolve_anthropic_model(
 
 def _build_output_type(
     agent: AgentDef,
-) -> type[Any] | ToolOutput[Any] | None:
+    backend: str = "anthropic",
+) -> type[Any] | ToolOutput[Any] | PromptedOutput[Any] | None:
     """Translate the agent's ``output`` schema into a Pydantic AI output spec.
 
     An empty or missing schema means plain text output (``None`` is returned
     so callers fall back to the default ``str`` output). Otherwise, the schema
-    is converted to a dynamic Pydantic model and wrapped in ``ToolOutput`` to
-    request tool-based structured output as required by the provider plan.
+    is converted to a dynamic Pydantic model.
+
+    For the ``openai`` backend (typically an OpenAI-compatible endpoint such
+    as Ollama, vLLM, or LM Studio), the output is wrapped in ``PromptedOutput``
+    rather than ``ToolOutput``. Several widely-used local models do not
+    reliably emit a real function/tool call for structured output over an
+    OpenAI-compatible API, even when explicitly requested (see
+    ollama/ollama#8095, ollama/ollama#8063, pydantic/pydantic-ai#877) --
+    ``ToolOutput`` on these backends silently fails after exhausting the
+    parse-recovery budget (``UnexpectedModelBehavior: Exceeded maximum output
+    retries``). ``PromptedOutput`` instead asks the model to emit the schema
+    as JSON in its normal text response, which these backends handle
+    correctly. The Anthropic backend is left on ``ToolOutput`` since real
+    tool-calling is reliable there.
 
     Args:
         agent: The Conductor agent definition.
+        backend: Which LLM backend this agent will run against.
 
     Returns:
-        A ``ToolOutput`` wrapping the dynamic model, or ``None`` for text output.
+        A ``PromptedOutput`` (openai backend) or ``ToolOutput`` (anthropic
+        backend) wrapping the dynamic model, or ``None`` for text output.
     """
     dynamic_model = output_schema_to_pydantic_model(
         f"{agent.name}Output",
@@ -235,6 +250,8 @@ def _build_output_type(
     )
     if dynamic_model is None:
         return None
+    if backend == "openai":
+        return PromptedOutput(dynamic_model)
     return ToolOutput(dynamic_model)
 
 
@@ -595,7 +612,7 @@ def build_agent(
             timeout=timeout,
         )
 
-    output_type = _build_output_type(agent)
+    output_type = _build_output_type(agent, backend)
     if output_type is None:
         output_type = str
 
--- a/src/conductor/providers/_pydantic_ai/agent_builder.py
+++ b/src/conductor/providers/_pydantic_ai/agent_builder.py
@@ -233,8 +233,17 @@ def _build_output_type(
     parse-recovery budget (``UnexpectedModelBehavior: Exceeded maximum output
     retries``). ``PromptedOutput`` instead asks the model to emit the schema
     as JSON in its normal text response, which these backends handle
-    correctly. The Anthropic backend is left on ``ToolOutput`` since real
-    tool-calling is reliable there.
+    correctly for flat/scalar output schemas. The Anthropic backend is left
+    on ``ToolOutput`` since real tool-calling is reliable there.
+
+    KNOWN LIMITATION: ``ToolOutput``'s rendered tool schema is sanitized
+    (``$ref``/``$defs`` inlined, pydantic-internal keys stripped) before
+    being sent to the model; ``PromptedOutput`` has no equivalent hook in
+    its current public API and renders the raw ``model_json_schema()`` for
+    a nested output schema, ``$ref``/``$defs`` included. This can make a
+    nested ``output:`` schema harder for a weak local model to satisfy than
+    a flat one. See the pinned regression test
+    ``TestOpenAIBackend::test_openai_nested_output_schema_is_not_ref_inlined_known_limitation``.
 
     Args:
         agent: The Conductor agent definition.
--- a/src/conductor/providers/_pydantic_ai/agent_builder.py
+++ b/src/conductor/providers/_pydantic_ai/agent_builder.py
@@ -245,6 +245,11 @@ def _build_output_type(
     a flat one. See the pinned regression test
     ``TestOpenAIBackend::test_openai_nested_output_schema_is_not_ref_inlined_known_limitation``.
 
+    Workaround until a follow-up fix lands: keep each ``openai``-backend
+    agent step's ``output:`` schema flat, and reassemble any nested
+    structure in a downstream ``type: script``/``type: set`` step instead
+    -- nesting is pure data shaping and needs no LLM call.
+
     Args:
         agent: The Conductor agent definition.
         backend: Which LLM backend this agent will run against.
--- a/tests/test_providers/test_pydantic_ai_agent_builder.py
+++ b/tests/test_providers/test_pydantic_ai_agent_builder.py
@@ -175,6 +175,46 @@ class TestOpenAIBackend:
         assert isinstance(pydantic_agent.output_type, PromptedOutput)
         assert not isinstance(pydantic_agent.output_type, ToolOutput)
 
+    def test_openai_nested_output_schema_is_not_ref_inlined_known_limitation(self) -> None:
+        """KNOWN LIMITATION, tracked as a follow-up, not fixed by this test.
+
+        ToolOutput's rendered tool schema is sanitized via
+        ``_sanitize_json_schema`` (inlining ``$ref``/``$defs``, stripping
+        pydantic-internal keys) before being sent to the model. PromptedOutput
+        has no equivalent hook in its public API (``outputs``/``name``/
+        ``description``/``template`` only) -- it renders the dynamic model's
+        raw ``model_json_schema()`` into the prompt, ``$ref``/``$defs`` and
+        all. For a *nested* output schema this means the openai/PromptedOutput
+        path hands a weak local model a schema shape it must dereference
+        itself, which is exactly the class of prompt the models this fix
+        targets are weakest at satisfying -- a real, open gap this fix does
+        not close. This test pins the current (undesirable but honest)
+        behavior so a future fix has a concrete regression target, and so a
+        silent behavior change doesn't go unnoticed either way.
+        """
+        agent_def = AgentDef(
+            name="nested_formatter",
+            output={
+                "nested": OutputField(
+                    type="object",
+                    properties={"inner": OutputField(type="string")},
+                )
+            },
+        )
+
+        pydantic_agent = build_agent(
+            agent_def, system_prompt="", rendered_prompt="", backend="openai"
+        )
+
+        assert isinstance(pydantic_agent.output_type, PromptedOutput)
+        rendered_schema = pydantic_agent._output_schema.object_def.json_schema
+        assert "$defs" in rendered_schema, (
+            "If this now fails, PromptedOutput's rendered schema is no longer "
+            "$ref-shaped for nested output -- the known limitation documented "
+            "above may have been resolved; update this test and the CHANGELOG "
+            "entry to reflect nested-schema coverage."
+        )
+
     def test_openai_sampling_settings(self) -> None:
         """OpenAI model settings must carry temperature, max_tokens and timeout."""
         agent_def = AgentDef(name="sampler")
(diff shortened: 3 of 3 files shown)


hertznsk:
A few architectural concerns regarding the unconditional switch to `PromptedOutput` for the `openai` backend:

### 1. Regression for OpenAI models
Applying `PromptedOutput` unconditionally across all `backend == "openai"` configurations degrades workflows running against actual OpenAI models (e.g. GPT-4o, o3-mini). For standard OpenAI endpoints, native function calling (`ToolOutput`) is reliable and ensures higher adherence than raw prompt-based JSON extraction.

### 2. Breakdown of schema sanitization (nested structures)
As noted in the description, `ToolOutput` applies `_sanitize_json_schema` (inlining `$defs`/`$ref` and stripping internal keys). With `PromptedOutput`, raw schemas containing `$defs` and `$ref` are passed directly. Requiring users to flatten schemas and reassemble them via manual `type: script` steps is a regression in workflow ergonomics.

### 3. Provider divergence and session continuation
Claude and OpenAI backends were intentionally unified under the Pydantic AI runner. Splitting them across different output mechanisms produces divergent message formats in `message_history`, complicating validator continuation loops and replay.

---

### Alternative solutions:

1. **System prompt injection for `final_result`:**
   Local models often miss calling `final_result` simply due to the lack of explicit instruction in the prompt. In practice, running local models (e.g. Qwen 3.8 / 27B FP8) with `ToolOutput` works reliably when injecting an explicit directive into the system prompt:
   > *"Before finishing the session, you must call the final_result tool. Without this tool call, your response will be ignored."*

   This resolves tool-calling reliability on local setups without abandoning `ToolOutput` or breaking schema sanitization.

2. **Opt-in / Configurable output mode:**
   If `PromptedOutput` is needed for certain constrained backends, it should be an explicit opt-in setting (e.g., `output_mode: tool | prompt`) rather than an unconditional default that degrades standard OpenAI and local tool-calling workflows.

jrob5756 (commented):
Five blocking findings here, so this shouldn't merge as-is. The core problem: switching from `ToolOutput` to `PromptedOutput` for the openai backend fixes the local-model tool-calling failure it targets, but takes several other things down with it — the issue #343 error contract, schema sanitization, parse-recovery telemetry, and the accuracy of the provider's own capability descriptor. None of those are edge cases; they're the direct, reproducible consequences of the same three-line branch.

- `src/conductor/providers/_pydantic_ai/agent_builder.py:267` — resurfaces the exact `Exceeded maximum output retries` error this PR exists to remove, with a now-false "output tool not called" suggestion (see inline).
- `src/conductor/providers/_pydantic_ai/agent_builder.py:268` — schema sanitization is skipped for every openai schema, not just nested ones, so the documented "keep it flat" workaround doesn't hold (see inline).
- `src/conductor/providers/_pydantic_ai/agent_builder.py:268` — `agent_parse_recovery` never fires on this backend, so every retry is invisible to the console, dashboard, and JSONL log (see inline).
- `src/conductor/providers/_pydantic_ai/agent_builder.py:267` — the branch keys on `backend == "openai"`, which also covers api.openai.com and Azure OpenAI, degrading structured output on endpoints with reliable tool calling that never needed this fix (see inline).
- No single line — `CAPABILITIES.structured_output` is still declared `"native"` for the openai provider (`src/conductor/providers/openai.py:138-139`), which is no longer true and is pinned by a test named `test_capabilities_are_stable`.

### b4 — BLOCKING: `CAPABILITIES.structured_output="native"` is now false, and a test pins the stale value

`openai.py:138-139` still reads:
```python
# Tool-based structured output: the schema is enforced via a forced tool call.
structured_output="native",
```
Both the comment and the value are wrong now. With `ToolOutput`, `allow_text_output=False` forces `tool_choice: "required"` and the model is API-constrained. With `PromptedOutput`, it resolves to `"auto"` — there's no API-level enforcement, the schema is just prompt text. `capabilities.py` defines `native` as SDK-level enforcement and `prompt_injection` as "schema appended to the prompt, model asked to comply" — that's exactly the mechanism this PR now uses, and `aca.py`/`claude_agent_sdk.py` already declare `prompt_injection` for the same thing. AGENTS.md is direct about this: "the descriptor is a contract; behavior must match what the provider declares."

This also strips users of the one runtime signal that would help them interpret a downstream parse failure — `describe_limitations()` renders `prompt_injection` as "structured output via prompt injection," but only if the capability says so.

Two things need to move with it: `tests/test_providers/test_openai.py:184` asserts `caps.structured_output == "native"` in a test whose docstring claims it reflects "the reviewed OpenAI contract," and `openai.py:113`'s class docstring says "structured output via tool use."

Worth flagging separately: `prompt_injection` is listed in AGENTS.md as an experimental-tier carve-out, and `openai` is `tier="stable"`. That's a real question this PR doesn't answer, which is also an argument for scoping the change to non-default endpoints (see the b5 thread).

**Suggested fix:** change the value to `"prompt_injection"`, rewrite the line-138 comment, update the docstring at `:113`, and update the pinned test. The validator's prompt-injection warning is gated on `caps.is_experimental`, so this introduces no new warnings for stable-tier workflows.

### r1 — RECOMMENDED: structured JSON output now streams to the console, dashboard, and JSONL log as an `agent_message`

Under `ToolOutput` the structured payload traveled in a `ToolCallPart` and never became an `agent_message`. Under `PromptedOutput` it arrives as a `TextPart`, and `events.py:157` maps `TextPart` to `agent_message` unconditionally (`:165` does the same for `TextPartDelta`).

I confirmed the part shape with a `FunctionModel`:
```
openai    -> TextPart '{"answer": "42"}'
anthropic -> ToolCallPart final_result, ToolReturnPart
```
So every openai-backend agent with an `output:` schema now emits its raw JSON as assistant chat text everywhere events get surfaced — and since `agent_parse_recovery` is also missing on this backend, rejected intermediate JSON attempts stream too, as ordinary assistant messages. AGENTS.md lists "same event types at the same semantic points" first under Provider Parity, and every other declared divergence in that document has an explicit carve-out. This one doesn't.

**Suggested fix:** either suppress the final-result text part from `agent_message` emission on the prompted path, or document the divergence in the `#### openai.py parity notes` section of AGENTS.md. Either way, pin it with a test asserting the observed `agent_message` sequence for an openai structured run.

### r5 — RECOMMENDED: user-facing docs still describe the old behavior

This changes user-visible structured-output behavior for the openai provider, but nothing outside the CHANGELOG was updated:

- `docs/providers/comparison.md:23` — the Structured Output row still says `Native` for OpenAI.
- `docs/providers/comparison.md:80` — "native schema enforcement... forced tool call structured output" is now the opposite of what happens, and it's given as a reason to pick this provider.
- `src/conductor/providers/_pydantic_ai/structured_output.py:3` — the module docstring says pydantic-ai is configured for `ToolOutput`, but this is the shared post-processing path for both backends. Its recovery suggestions at `:68`/`:105` ("ensure the model calls the output tool") are wrong advice on the openai path too.
- AGENTS.md's `#### openai.py parity notes` section, which exists specifically to record this kind of divergence, has no entry for it — the convention there is that an undocumented divergence tends to get re-litigated or accidentally reverted.

Separately, the workaround narrative ("keep the schema flat") is filed in the docstring of a private Python helper. A user hitting `Exceeded maximum output retries` won't find it there — `docs/providers/openai.md`'s Troubleshooting section currently only covers a missing API key.

**Suggested fix:** update both comparison.md rows and the `structured_output.py` docstring/suggestions, add a parity-notes bullet to AGENTS.md, and add a "Structured output on local models" subsection to `docs/providers/openai.md` with a troubleshooting entry keyed on the exact exception string.

src/conductor/providers/_pydantic_ai/agent_builder.py
```diff
@@ -215,26 +215,57 @@ def _resolve_anthropic_model(
...
+        A ``PromptedOutput`` (openai backend) or ``ToolOutput`` (anthropic
+        backend) wrapping the dynamic model, or ``None`` for text output.
     """
     dynamic_model = output_schema_to_pydantic_model(
         f"{agent.name}Output",
         agent.output,
     )
     if dynamic_model is None:
         return None
+    if backend == "openai":
```
jrob5756, line 267: **BLOCKING**: this switch breaks the issue #343 error contract and brings back the exact failure this PR is meant to fix.

AGENTS.md's recovery rule: "On budget exhaustion, re-raise the original ValidationError... reserve ProviderError for syntax failures." `retry.py:512` implements that with a single-level check — `getattr(last_error, "__cause__", None)` then `isinstance(cause, PydanticValidationError)`. That only holds for `ToolOutput`. Under `PromptedOutput`, pydantic-ai raises through a `ToolRetryError`, which adds a level to the chain.

I reproduced this with an identical agent (`output: {answer: string}`) and an identical malformed payload, through `execute_with_retry`:

```
openai    -> ProviderError: ...Exceeded maximum output retries (2)
             suggestion: "...(output tool not called); retry later or simplify the output schema"
             chain: UnexpectedModelBehavior <- ToolRetryError <- ValidationError

anthropic -> ValidationError: 1 validation error for fOutput — answer: Field required
             chain: ValidationError
```

So openai users get the literal `Exceeded maximum output retries` message the CHANGELOG says this PR eliminates, for the failure class it knowingly leaves open. The attached suggestion ("output tool not called") is also just wrong now — there's no output tool. And the field-level detail is still one level down in the chain, discarded. The synthetic fixture at `tests/test_providers/test_pydantic_ai_retry.py:925` (`umb.__cause__ = e`) can't catch any of this — it hardcodes a one-level chain that no longer matches what pydantic-ai actually produces here.

Suggested fix — walk the chain instead of peeking one level. The bounded, cycle-guarded pattern already exists in this file (`_describe_root_cause`, `_get_retry_after`):
```python
cause: BaseException | None = last_error
seen: set[int] = set()
for _ in range(5):
    cause = getattr(cause, "__cause__", None)
    if cause is None or id(cause) in seen:
        break
    seen.add(id(cause))
    if isinstance(cause, PydanticValidationError):
        raise ValidationError(
            f"Structured output validation failed: {cause}",
            suggestion=(
                "Fix the model's output to match the declared schema; "
                "retrying will not resolve a shape mismatch."
            ),
        ) from cause
```
Also fix or drop the "(output tool not called)" text at `retry.py:522`, and add a two-backend test asserting the same `ValidationError` shape from both.

src/conductor/providers/_pydantic_ai/agent_builder.py
```diff
@@ -215,26 +215,57 @@ def _resolve_anthropic_model(
...
+        return PromptedOutput(dynamic_model)
```
jrob5756, line 268: **BLOCKING**: schema sanitization is skipped for every openai schema now, including flat ones, so the documented "keep it flat" workaround doesn't actually work.

The sanitization loop at `agent_builder.py:675` is gated on `isinstance(pydantic_agent.output_type, ToolOutput)`, so it's skipped entirely on this path now — not just for nested schemas. The docstring at line 236, the KNOWN LIMITATION note, and the CHANGELOG all scope this gap to nested schemas and prescribe a flat-schema workaround as sufficient. It isn't. `_sanitize_json_schema` does four things (inlines `$ref`/`$defs`, strips `default`, strips pydantic-internal `ge`/`le`, flattens nullable `anyOf`), and the openai path now skips all four even on a flat schema.

I dumped the rendered schema for `{answer: string, score: number(0-10), note: string(optional), maybe: string|null}` on both backends:
```
openai (raw):     "score": {anyOf:[int,number], ge:0, le:10, maximum:10, minimum:0}
                  "note":  {default: null, type: string}
                  "maybe": {anyOf:[{type:string},{type:null}]}

anthropic (san.): "score": {anyOf:[int,number], maximum:10, minimum:0}
                  "note":  {type: string}
                  "maybe": {type: [string, null]}
```
`note` is the worst case — `{default: null, type: string}` tells a compliant model to emit `null`, and the generated validator then rejects it ("Input should be a valid string"), burning parse-recovery budget and reproducing the exact `Exceeded maximum output retries` failure this PR targets. This is a regression against `origin/main`, and it lands hardest on the weak local models the change is meant to help.

The stated reason for skipping this — that `PromptedOutput` "has no equivalent hook" — doesn't hold up: the `ToolOutput` sanitization four lines above is itself private-attribute access (`_output_schema.toolset._tool_defs`), and this PR's own test at line 210 already reads the exact dict that needs sanitizing. I verified the symmetric fix end to end: after sanitizing, `$ref`/`$defs` are inlined, `default`/`ge`/`le` are stripped, and a full `FunctionModel` run still parses correctly. Validation is unaffected since `ObjectOutputProcessor.validate` uses the `TypeAdapter`, never `object_def.json_schema` (that's render-only).

Suggested fix — extend the guard at `agent_builder.py:675` to cover both wrappers, which closes the gap entirely and lets the KNOWN LIMITATION paragraphs come out rather than ship:
```python
if isinstance(pydantic_agent.output_type, ToolOutput):
    toolset = pydantic_agent._output_schema.toolset
    if toolset is not None:
        for tool_def in toolset._tool_defs:
            _sanitize_json_schema(tool_def.parameters_json_schema)
elif isinstance(pydantic_agent.output_type, PromptedOutput):
    object_def = getattr(pydantic_agent._output_schema, "object_def", None)
    if object_def is not None:
        _sanitize_json_schema(object_def.json_schema)
```
The pinned test can then be inverted into a positive assertion. If this is deferred instead, the docstring and CHANGELOG need to stop claiming flat schemas are unaffected.

jrob5756, line 268: **BLOCKING**: `agent_parse_recovery` never fires on the openai backend, so every retry is invisible.

`events.py:49` gates recovery telemetry on `isinstance(event, OutputToolResultEvent)`. Checking against the installed pydantic-ai, that event is produced from exactly one place — `_tool_execution.py:138`, inside `_emit_output_tool_events`, which takes a `ToolCallPart` and only runs on the output-tool dispatch path. `PromptedOutput` never dispatches an output tool; its correction loop runs through `_agent_graph.py` via a `ModelRequestNode` and emits no such event.

Reproduced with an identical agent and a malformed-then-valid payload sequence:
```
openai     parse_recovery: []          <- nothing emitted
anthropic  parse_recovery: [{attempt: 1, max_attempts: 3, reason: 'schema', ...}]
```
AGENTS.md states this directly: "Each attempt emits agent_parse_recovery," and Provider Parity leads with "same event types emitted at the same semantic points."

The user cost is real. `cli/run.py` renders these as a `WARNING: retrying '<agent>' output... attempt N/M`; that warning, the JSONL record, and the dashboard row all vanish for openai. Someone on Ollama watching a step take 3x as long gets no signal it's being re-prompted, and a workflow spending 3x tokens on a structured step is invisible in the usage summary. `interrupt.py:225`'s `recovery_attempt` counter is now permanently stuck at zero on this backend.

Suggested fix — extend `emit_output_recovery_event` to also detect the prompted-path retry. It's observable at the node transition: the returned `ModelRequestNode` carries a `RetryPromptPart` with `tool_name is None`. Whatever mechanism you land on, add a test asserting both backends emit `agent_parse_recovery` for the same malformed response — the missing symmetry test is why this got through CI.

src/conductor/providers/_pydantic_ai/agent_builder.py
jrob5756, line 267: **BLOCKING**: this degrades api.openai.com and every other endpoint with reliable tool calling.

The branch keys only on `backend == "openai"`, but that backend serves api.openai.com, Azure OpenAI, OpenRouter, corporate gateways, vLLM, LM Studio, and Ollama. Every justification in the docstring (lines 226-236) and the PR description is about weak local models — none of it applies to GPT-4o or GPT-5, where tool calling is reliable and tool-enforced structured output is strictly better than prompt compliance.

For those endpoints this trades server-side constrained decoding (`tool_choice: "required"`) for pure prompt compliance, and stacks the unsanitized schema, degraded error reporting, and lost recovery telemetry raised elsewhere in this review on top. There's an added cost for tool-using agents specifically: under `ToolOutput` the model must emit a tool call each turn, but under prompted mode any prose turn without one gets consumed as a final-output parse attempt and burns retry budget.

The discriminator is too coarse for what's actually being tested. The real condition is "does this endpoint reliably tool-call," not "which SDK client is this." Someone on GPT-5 with an `output:` schema gets downgraded because someone else's Ollama can't tool-call.

`build_agent` already receives `base_url` (line 561), and it's `None` exactly when the run targets the official endpoint — the provider deliberately never writes it back on the ambient path, per the openai.py parity notes in AGENTS.md.

Suggested fix — gate the prompted path on a custom endpoint rather than the backend family:
```python
if backend == "openai" and base_url is not None:
```
or add an explicit opt-in like `runtime.provider.structured_output_mode: tool | prompted | auto`. Keep `ToolOutput` as the default for the official endpoint. `pydantic_ai.output.NativeOutput` (`response_format: json_schema`) is worth a look too, for endpoints whose profile advertises `supports_json_schema_output` — Ollama supports `format`/`response_format` as well.

src/conductor/providers/_pydantic_ai/agent_builder.py
```diff
@@ -215,26 +215,57 @@ def _resolve_anthropic_model(
 
 def _build_output_type(
     agent: AgentDef,
-) -> type[Any] | ToolOutput[Any] | None:
+    backend: str = "anthropic",
```
jrob5756, line 218: **RECOMMENDED**: `backend` is typed as bare `str` with a default, so a typo silently picks the wrong output mode.

`build_agent` types the same concept as `Literal["anthropic", "openai"]` (line 565); this helper widens it to `str` and adds a default. Two problems follow. First, the widened type means `_build_output_type(agent, "openAI")` or `"open-ai"` type-checks under `ty` and silently falls through to `ToolOutput` — reproducing the exact bug this PR fixes, with no error and no log. The codebase already defines named `Literal` aliases for closed vocabularies throughout `providers/` (`ProviderTier`, `StructuredOutputMode`, `ReasoningEffortLevel`); a bare `str` here is off-convention, and it's a second, wider spelling of a vocabulary already spelled precisely 350 lines below. Second, the default is unreachable — the only caller (line 629) always passes `backend` explicitly — so it buys nothing while creating a second source of truth that can drift from `build_agent`'s own dispatch.

Separately, the return union at line 219 still lists `type[Any]`, which the function never returns (only `None`, `PromptedOutput`, or `ToolOutput`). That was already inaccurate before this PR touched the line; now the union is four members wide for no reason.

```suggestion
def _build_output_type(
    agent: AgentDef,
    *,
    backend: Literal["anthropic", "openai"],
) -> ToolOutput[Any] | PromptedOutput[Any] | None:
```
`Literal` is already imported. Worth hoisting a shared `PydanticAIBackend = Literal["anthropic", "openai"]` alias used at both this line and line 565, so the vocabulary has one spelling.

tests/test_providers/test_pydantic_ai_agent_builder.py
```diff
@@ -149,8 +150,18 @@ def test_openai_default_constant_used_when_no_model_anywhere(self) -> None:
 class TestOpenAIBackend:
     """Tests specific to the openai backend branch."""
 
-    def test_openai_output_schema_becomes_tool_output(self) -> None:
-        """OpenAI backend must wrap a non-empty output schema in ToolOutput."""
+    def test_openai_output_schema_becomes_prompted_output(self) -> None:
```
jrob5756, line 153: **RECOMMENDED**: no test exercises the actual behavior this PR fixes, and the seam to write one already exists.

The PR's whole justification is a runtime failure — a model answering with JSON in a text part instead of a tool call. Both new tests here assert only the type of `pydantic_agent.output_type`, which just restates the implementation. They'd pass unchanged even if `PromptedOutput` were completely unable to parse a text-JSON response — i.e., if the fix didn't fix anything. The full provider suite (1322 passed) noticed nothing about the behavior change either.

`FunctionModel` is already imported at line 22, and the monkeypatch idiom is already used elsewhere in this file. I checked both halves in the worktree: with `_resolve_openai_model` patched to a `FunctionModel` returning `TextPart('{"answer": "42"}')`, `build_agent(..., backend="openai")` parses correctly; the same text-only model against the `ToolOutput` agent raises `UnexpectedModelBehavior: Exceeded maximum output retries (1)` — the exact error the CHANGELOG cites.

A test that fails on `main` and passes on this branch is about fifteen lines. Without it, nothing catches a regression where the text-JSON path stops parsing — a pydantic-ai upgrade, a stray `output_type` override in the runner, or a revert. `TestOutputRecovery` (lines 636-712) already does this shape of test for the anthropic tool path; the openai text path deserves the same.

Two smaller gaps alongside it: no test passes `backend="anthropic"` explicitly (only covered via `build_agent`'s default), and the empty-output-schema fallback is only tested on anthropic, even though "local model, no output schema" is the most common openai configuration.

Suggested fix: add an end-to-end test on the openai backend driving a `FunctionModel` that returns the structured payload as a `TextPart` and asserts the parsed output, mirroring `TestOutputRecovery`, plus a direct unit test over `_build_output_type` covering both backends explicitly and an empty-schema case on openai.

tests/test_providers/test_pydantic_ai_agent_builder.py
```diff
@@ -161,7 +172,48 @@ def test_openai_output_schema_becomes_tool_output(self) -> None:
...
+                )
+            },
+        )
+
+        pydantic_agent = build_agent(
+            agent_def, system_prompt="", rendered_prompt="", backend="openai"
+        )
+
+        assert isinstance(pydantic_agent.output_type, PromptedOutput)
+        rendered_schema = pydantic_agent._output_schema.object_def.json_schema
```
jrob5756, line 210: **RECOMMENDED**: this test asserts on two layers of pydantic-ai private attributes and can't report its own failure message.

`pydantic_agent._output_schema.object_def.json_schema` reaches through two independently private hops under a `pydantic-ai-slim>=2.28.0,<3` pin. If a minor upgrade renames either, the test raises `AttributeError` — a test error, not an assertion failure — so the failure message at lines 211-216 (the whole mechanism by which a future maintainer learns what to do) never prints. The likely reaction to a loud, unrelated breakage like that is to just delete the test.

It's also a weaker proxy than it looks for "a weak model receives a harder-to-satisfy schema." `object_def.json_schema` gets populated whether or not pydantic-ai actually injects the schema into the request, so a future regression that dropped schema instructions entirely — catastrophic for exactly the models this PR targets — would still pass this assertion.

There's a public seam closer to what actually matters: inside a `FunctionModel` function, `info.instructions` is the literal text the model receives. I captured it for the nested case and it contains the rendered schema verbatim, `$defs`/`$ref` included.

(If the sanitization fix from the other thread lands, this becomes a positive assertion and most of this concern goes away on its own.)

Suggested fix: assert against the model-facing contract instead — capture `info.instructions` inside a `FunctionModel`. If you keep the private access, at least localize it behind a named helper next to `_extract_output_tool_schema`, so an upgrade breaks one clearly named place instead of an `AttributeError` inside a test body.

CHANGELOG.md
```diff
@@ -27,6 +27,49 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  **Workaround, until that follow-up lands:** avoid a nested `output:`
+  schema on an `openai`-backend agent step entirely. Have each agent step
+  return a flat schema, then reassemble the nested structure in a
+  downstream `type: script` (or `type: set`) step — nesting is pure data
+  shaping and needs no LLM call, so it never hits this limitation. This is
+  also the same decomposition pattern that improves parse-recovery
+  reliability on weak local models generally (smaller, flatter one-shot
+  generations retry more predictably than one large structured call), so
+  it is a good default shape for a local-model-backed workflow regardless
+  of this specific limitation. See `examples/` (a worked example is a good
```
jrob5756, line 70: **RECOMMENDED**: this entry ships a note-to-self and a dead pointer, with no tracking issue for the promised follow-up.

The entry ends with a parenthetical addressed to maintainers ("a worked example is a good follow-up addition here") that ships verbatim in the release notes and stays there permanently — including after the example gets added, at which point it's just wrong. "See `examples/`" also points at 50+ files with no indication of which one, and none of them currently demonstrate the flatten-and-reassemble pattern. The entry directly above this one links a specific file, which is the convention here.

The docstring, the CHANGELOG, and the test all reference a "follow-up fix" with no issue number anywhere, while the neighboring entry is anchored to `(#501)` and the file carries 60 other `(#NNN)` references. Six months out, nobody can check whether the follow-up landed — which matters because the workaround is a structural instruction, and workflows restructured on this advice stay restructured after the limitation is fixed. The entry also embeds a full pytest node ID (line 58), which rots the first time that test gets renamed.

Small accuracy note: `qwen2.5-coder:14b-8k` isn't an upstream Ollama tag, so the verification isn't reproducible as written.

```suggestion
```
Suggested fix: drop the parenthetical, or add the worked example in this PR and link it directly. File a tracking issue and reference it from the docstring and CHANGELOG instead of "until a follow-up fix lands." Keep the pytest node ID in the docstring only, and describe the model as "a 14B qwen2.5-coder build with an 8K context window." If the sanitization fix from the other thread is applied, most of this section can just come out.

jrob5756:
Hi @josephrobertlopez - Thank you so much for contributing to conductor! We decided to take #524 in place of this one. I'll go ahead and close this PR for now. But feel free to continue to contribute to the repository and accept the Contributor License Agreement :) Cheers!

PR #511, opened 2026-09-07
Preserve agent conversation across validator retries

hertznsk:
# Preserve agent conversation across validator retries

## Summary

- preserve the completed conversation when a semantic validator rejects output, for providers with an in-memory continuation surface (`claude`, `openai`, `hermes`)
- send validation feedback as the next user turn instead of rebuilding the original prompt from scratch
- keep existing stateless validator retry behavior for providers that do not expose continuation state (`copilot`, `claude-agent-sdk`, `aca`)

## Why

Validator retries previously called the primary agent again with a reconstructed prompt. That discarded the model response and any tool-call/tool-result exchanges from the first attempt. The correction now continues the completed conversation — Pydantic AI `message_history` for `claude`/`openai`, the run's own `messages` list for `hermes` — so the model can revise its existing work with the validator feedback in context.

## Provider contract

- a provider that can resume a completed run declares `supports_continuation=True` and populates `AgentOutput.continuation_state` on completed outputs; the executor refuses state from a provider that does not declare it (the rendered prompt is never discarded on a coincidence)
- the value is provider-opaque (`object` at the boundary), never handed to a different provider, and never serialized to checkpoints or event logs
- `hermes` withholds the state when a parse-recovery call produced the output, since that output lives in the recovery conversation the outer result never carries

## Testing

- `make check`
- full suite: 7839 passed (+ 790 fleet), the only failures are four pre-existing environment ones (two POSIX unreadable-path assertions, two Claude performance thresholds)
- inbound forwarding pinned by mutation check: deleting the `message_history=` forwarding fails the new tests
- frontend: 159 vitest tests pass; static assets rebuilt via `make build-frontend`

## Notes

- `agent_prompt_rendered` is tagged `continuation: true` on retries; the dashboard appends the feedback turn to the node's prompt instead of replacing it
- a failed re-run now reports its cause (`error`, plus a `continued` flag) on the `agent_validation_failed` event, in the console log, and in the dashboard activity entry
- the context-growth trade-off (the retry replays the whole first conversation) is documented in `docs/workflow-syntax.md`
- rebased onto `main` @ `544a2bf`


src/conductor/providers/_pydantic_ai/runner.py
```diff
@@ -172,6 +175,7 @@ def intercepting_callback(event_type: str, data: dict[str, Any]) -> None:
             usage_limits=UsageLimits(request_limit=max_agent_iterations),
             max_session_seconds=max_session_seconds,
             max_parse_recovery_attempts=retry_cfg.max_parse_recovery_attempts,
+            message_history=message_history,
```
jrob5756, line 178: **BLOCKING**: The feature's central wire — forwarding `message_history` into the provider — has no test.

Deleting all three `message_history=` forwardings (`runner.py:178`, `openai.py:1053`, `claude.py:1220`) from an isolated copy of `src` leaves the suite green: 2737 tests pass. Nothing in the repo asserts that a `continuation_state` handed to `provider.execute()` actually reaches `agent.iter()`.

The two new provider assertions only check the outbound half — `output.continuation_state is not None` after a fresh run (`test_openai.py:248`, `test_pydantic_ai_provider.py:78`). The inbound half, which is the entire point of this PR, is unpinned. When it breaks, it won't break loudly: every validator retry silently degrades back to a stateless re-run that sends only the feedback text, since `AgentExecutor` already skipped prompt rendering on that path. The model gets asked to correct output for a task it was never shown, and answers plausibly anyway.

**Suggested fix:** Add an inbound assertion to `tests/test_providers/test_openai.py::TestExecute` (and the Claude equivalent), capturing `run_with_interrupt`'s kwargs:

```python
history = (await provider.execute(agent, {}, "first")).continuation_state
await provider.execute(agent, {}, "feedback only", continuation_state=history)
assert captured["message_history"] is history
assert captured["user_prompt"] == "feedback only"
```

A cheaper equivalent is a direct unit test on `run_agent_pipeline(message_history=sentinel, ...)` asserting the sentinel reaches a patched `run_with_interrupt`.

src/conductor/executor/agent.py
```diff
@@ -424,17 +426,16 @@ async def execute(
...
-        prefix = self._build_prompt_prefix(agent, event_callback)
-        if prefix:
-            rendered_prompt = prefix + rendered_prompt
-
-        # Append user guidance section if provided
-        if guidance_section:
-            rendered_prompt = rendered_prompt + guidance_section
+        # Render a new task or add one user turn to provider-owned state.
+        if continuation_state is not None:
+            rendered_prompt = guidance_section or ""
```
jrob5756, line 464: **BLOCKING**: Nothing verifies the provider can honor `continuation_state` before the task prompt gets discarded.

`rendered_prompt = guidance_section or ""` throws away the rendered task, the workspace-instructions preamble, and the eagerly injected skills. The only thing making that safe is the coincidence that just two of six providers populate `AgentOutput.continuation_state` today. The executor itself performs no check.

`continuation_state` is a public, `Any`-typed, mutable field on `AgentOutput` (`base.py:184`). All four non-supporting providers (`copilot.py:834`, `hermes.py:209`, `aca.py:1074`, `claude_agent_sdk.py:798`) now accept the kwarg and silently discard it — no docstring entry, no `del` — so "deliberately ignored" is indistinguishable from "not wired up yet." Mutation testing shows the consequence directly: making `CopilotProvider` return `continuation_state="session-abc"`, a plausible future change since it already has `session_continuity`, causes every validator retry to send a bare `## Validation feedback` turn with no task and no history, and all 16 validator integration tests still pass. The retry succeeds and its answer gets accepted as final.

The repo already has the idiom for this: `supports_native_skills`, `supports_native_plugins`, and `session_continuity` all declare provider-differentiated behavior and get enforced (`_reject_unsupported_session_key`). This is the first behavior like it added with nothing enforcing it. One reviewer argued against a `ProviderCapabilities` field because every existing entry drives a `conductor validate` decision and this one has no YAML surface — a plain `supports_continuation` property on `AgentProvider`, alongside the two existing `supports_native_*` properties, sidesteps that objection.

**Suggested fix:** Add a `supports_continuation` property to `AgentProvider` (default `False`, `True` on `ClaudeProvider`/`OpenAIProvider`) and gate the branch:

```python
if continuation_state is not None and not getattr(self.provider, "supports_continuation", False):
    raise ExecutionError(
        f"Agent '{agent.name}': provider '{type(self.provider).__name__}' returned "
        f"continuation state it cannot resume. Continuing would send the model only "
        f"the follow-up turn, without the original task prompt.",
        agent_name=agent.name,
    )
```

Pair it with two tests: assert the four non-supporting providers leave `continuation_state is None`, and tighten `test_fail_then_rerun_succeeds` (line 194) with `assert primary_prompts[1].startswith(primary_prompts[0])` so the stateless path can no longer pass with a feedback-only prompt.

src/conductor/executor/agent.py
```diff
@@ -424,17 +426,16 @@ async def execute(
...
-        # Prepend prompt prefix (workspace instructions + optional skills)
-        prefix = self._build_prompt_prefix(agent, event_callback)
-        if prefix:
-            rendered_prompt = prefix + rendered_prompt
-
-        # Append user guidance section if provided
-        if guidance_section:
-            rendered_prompt = rendered_prompt + guidance_section
+        # Render a new task or add one user turn to provider-owned state.
+        if continuation_state is not None:
```
jrob5756, line 444: **RECOMMENDED**: Validator retries overwrite the agent's prompt in the dashboard and event log with a feedback-only stub.

On the continuation path `rendered_prompt` is just `"## Validation feedback\n…"`, and that's what the `agent_prompt_rendered` event carries a few lines below. The frontend store applies it unconditionally (`nd.prompt = data.rendered_prompt` at `web/frontend/src/stores/workflow-store.ts:1677`), so after a validator retry on `claude` or `openai`, the node's Prompt panel shows only the feedback fragment. The agent's actual task prompt is gone. The same stub lands in the JSONL event log's last prompt record too.

Before this PR the retry prompt was a strict superset of the original, so nothing was lost. A validator retry is exactly the moment a user opens that node to understand what happened, and now what they find is indistinguishable from the genuine "task prompt was dropped" bug in `b2`.

**Suggested fix:** Tag the event so the consumer can tell the two apart, and have the store append rather than replace when the flag is set:

```python
event_callback(
    "agent_prompt_rendered",
    {
        "rendered_prompt": rendered_prompt,
        "context_keys": list(context.keys()) if isinstance(context, dict) else [],
        "continuation": continuation_state is not None,
    },
)
```

Requires a `make build-frontend` for the store change.

src/conductor/executor/agent.py
```diff
@@ -524,18 +525,33 @@ async def execute(
...
-            tools=resolved_tools,
-            interrupt_signal=interrupt_signal,
-            event_callback=event_callback,
-            skill_directories=skill_dirs,
-            custom_agents=custom_agents,
-            extra_mcp_servers=extra_mcp_servers,
-        )
+        # Execute via provider. Keep the existing call shape unless this is a
+        # provider-owned continuation, so external providers remain compatible.
+        if continuation_state is None:
```
jrob5756: **RECOMMENDED**: The duplicated `provider.execute` call guards against nothing, while the real compatibility break went in unguarded.

The comment says the branch exists "so external providers remain compatible," but it can't protect anything: the executor only forwards `continuation_state` when it's non-`None`, and it's non-`None` only because a provider returned it — so that provider already accepts the kwarg. There's also no external-provider extension point: `ProviderName` is a closed `Literal` of six names, and `create_provider` is a `match` over those six with no entry-point or registration hook.

Meanwhile the actual compatibility break shipped unguarded. `AgentProvider.execute` has no keyword-only marker, and `continuation_state` was inserted between `custom_agents` and `extra_mcp_servers` (`base.py:423`), silently changing the meaning of positional argument #9 for every caller and subclass. Precedent cuts the same way: PR #394 added `custom_agents` and `extra_mcp_servers` to this exact call site unconditionally, no compatibility branch.

The cost is nine kwargs duplicated across two arms. Whoever adds the next provider kwarg has to edit both, and forgetting the second produces a bug that only shows up on validator retries with Claude/OpenAI — the least-exercised path in the file.

**Suggested fix:** Move `continuation_state` to the end of the signature (after `extra_mcp_servers`) in `base.py` and all six providers, add `*` after `rendered_prompt` so future insertions are provably non-breaking, and collapse lines 530-554 into a single unconditional call passing `continuation_state=continuation_state`. Do the same for `AgentExecutor.execute`, where the parameter was inserted ahead of `interrupt_signal`.

src/conductor/providers/base.py
```diff
@@ -181,6 +181,9 @@ class AgentOutput:
     """Sandbox wall-clock time reported by a remote-runtime provider (issue #284,
     FR7). ``None`` for providers with no distinct sandbox time to report."""
 
+    continuation_state: Any = None
```
jrob5756: **RECOMMENDED**: `Any` disables type checking at the two boundaries where the value is actually unsafe.

`run_with_interrupt` and `run_agent_pipeline` are both correctly typed `Sequence[ModelMessage] | None`. That typing gets discarded at exactly the two lines where an unvalidated value crosses into it: `message_history=continuation_state` in `claude.py:1220` and `openai.py:1053`. Any object at all flows straight into `agent.iter(message_history=...)`.

`Any` isn't the spelling of "opaque," it's the spelling of "unchecked." Annotating the parameter `object` instead produces `error[invalid-argument-type]: Expected Sequence[ModelMessage] | None, found object` under the repo's own `ty` checker — proof the current annotation opts out of a check that would otherwise fire. The failure mode when it does go wrong is opaque too: a wrong-shaped value yields `AttributeError: 'str' object has no attribute 'conversation_id'` from deep inside pydantic-ai, which `_apply_validator`'s broad `except Exception` then swallows into a fail-open. Your own new tests use `["original request", "original response"]` as the fake state — a `list[str]` that type-checks only because of `Any`, and that's exactly the shape that produces that `AttributeError` in production.

A related invariant is also unexpressed: state produced by one provider must never feed into another. Not reachable today (`_apply_validator` reuses the same executor), but `claude` and `openai` are both Pydantic-AI providers, so a cross-feed wouldn't crash — it would produce a run whose history contains tool calls for tools the new agent never registered.

**Suggested fix:** Minimum change: annotate `object` rather than `Any` on both the field and the parameter, and add a single documented `cast` at each of the two Pydantic-AI call sites. Stronger: wrap it in a frozen `ContinuationState(owner: str, payload: object)` where `owner` is the producing provider's `__qualname__`, checked on resume — that makes the provider-match invariant enforceable in one line.

src/conductor/executor/agent.py
```diff
@@ -360,6 +361,7 @@ async def execute(
             guidance_section: Optional user guidance section to append to the
                 rendered prompt. When provided, this is appended after the
                 rendered prompt text.
+            continuation_state: Optional provider-specific state from a completed run.
```
jrob5756: **RECOMMENDED**: Docstrings across the changed call chain now contradict the code.

The prose in AGENTS.md, CHANGELOG.md, and docs/workflow-syntax.md was updated carefully, but the docstrings on the functions that implement the change weren't:

- `AgentExecutor.execute` (line 352-356) still states unconditionally "This method: 1. Renders the agent's prompt template with context" — false when `continuation_state` is set.
- `guidance_section` (line 361-363) still says it's "appended after the rendered prompt text." On the continuation path it *becomes* the entire prompt. Worse than an omission — actively wrong for the caller this PR added.
- `continuation_state` (line 364) is one line and skips the caller-visible consequences: that the prompt template, prompt prefix, and eager skill injection are all skipped.
- `_apply_validator` (`workflow.py:4006-4009`, `4028-4029`) still describes the feedback as "appended" and `guidance_section` as something the feedback is appended after; on the continuation path `guidance_section` is discarded entirely.
- `_build_validation_feedback` (`workflow.py:4190`) says "appended on re-run."
- The four providers that ignore the kwarg have no `Args:` entry for it, breaking a convention this repo follows closely (`hermes.py` and `openai.py` both spell out "Ignored. …" for every unused parameter, `aca.py` additionally `del`s them).

The inline comment at line 429 is currently the only accurate description of this behavior anywhere in the file.

**Suggested fix:** Update each site. For the executor:

```python
guidance_section: Optional user guidance section. Without
    ``continuation_state`` it is appended after the rendered prompt
    text. With ``continuation_state`` it *is* the prompt — the sole
    new user turn appended to the provider's existing conversation.
continuation_state: Optional provider-specific state from a completed
    run. When not ``None``, the prompt template, workspace-instructions
    preamble and eager skill injection are all skipped — they are
    already present in the provider-held conversation.
```

For the four ignoring providers, add `continuation_state: Ignored. This provider has no in-memory continuation surface, so it never populates ``AgentOutput.continuation_state`` and is never handed one back.` plus a `del continuation_state`, matching `aca.py`.

src/conductor/engine/workflow.py
```diff
@@ -4136,7 +4136,10 @@ def _emit_v(event_type: str, data: dict[str, Any]) -> None:
                 executor.execute(
                     agent,
                     agent_context,
-                    guidance_section=new_guidance,
+                    guidance_section=(
+                        feedback.lstrip() if output.continuation_state is not None else new_guidance
```
jrob5756: **RECOMMENDED**: The most surprising line in the PR is an inline conditional with no explanation.

Three non-obvious decisions get packed into an argument list nested two calls deep, in a function that's otherwise heavily commented (an eight-line comment justifies the fail-open handler right below it).

1. `guidance_section` is deliberately dropped on the continuation path. Correct, since the guidance was in the first run's prompt and is already inside `continuation_state`, but it reads as a bug, and a future maintainer "fixing" it would duplicate the guidance.
2. `.lstrip()` is unexplained. It exists because `_build_validation_feedback` prefixes `"\n\n"` for the append case, which is wrong when the string is a standalone user turn.
3. `new_guidance` (line 4132) is still computed unconditionally but is dead on the continuation path, which visually implies it's always used.

**Suggested fix:** Hoist to a named local above the `try:` with the reasoning attached:

```python
feedback = self._build_validation_feedback(outcome.issues)
continuation_state = output.continuation_state
if continuation_state is None:
    # Stateless re-run: the whole prompt is rebuilt, so the feedback lands
    # after any guidance already appended to the primary prompt.
    rerun_guidance = (guidance_section or "") + feedback
else:
    # The provider still holds the first run's conversation, which already
    # contains the rendered prompt *and* guidance_section — re-sending them
    # would duplicate both. lstripped because _build_validation_feedback
    # prefixes newlines for the append case.
    rerun_guidance = feedback.lstrip()
```

src/conductor/engine/workflow.py
```diff
@@ -4136,7 +4136,10 @@ def _emit_v(event_type: str, data: dict[str, Any]) -> None:
...
+                    ),
+                    continuation_state=output.continuation_state,
```
jrob5756: **RECOMMENDED**: The retry now re-sends the full first-run history, and the resulting failure is nearly invisible.

The continuation re-run replays the entire first conversation, including every tool exchange bounded only by `runtime.tool_output.max_chars` (50k default), plus the feedback. The previous behavior was a fresh, bounded prompt. For a tool-heavy agent that already burned most of its context window on attempt 1, the retry can now overflow where it previously succeeded. Compaction absorbs some of this on `claude`/`openai`, but the direction of the change is clear, and the CHANGELOG presents it purely as a saving ("without repeating the original prompt").

What makes this worth fixing is where the failure lands. `_apply_validator` catches it, keeps the original output, and reports through two channels that both under-serve the user: `logger.warning` (which reaches stderr only via Python's `lastResort` handler, unformatted), and an `agent_validation_failed` event whose console rendering is gated behind `verbose_log` — so in default mode nothing prints at all. The event payload carries `{issues, will_retry, rerun_errored}`, not the exception type or message, so even in verbose mode the user learns *that* the retry failed but never *why*.

**Suggested fix:** Put the cause on the event so the one user-visible surface is actionable:

```python
_emit_v(
    "agent_validation_failed",
    {
        "issues": outcome.issues,
        "will_retry": False,
        "rerun_errored": True,
        "error": f"{type(exc).__name__}: {exc}",
        "continued": output.continuation_state is not None,
    },
)
```

Surface the reason in `cli/run.py`'s `rerun_errored` branch, and note the context-growth tradeoff in `docs/workflow-syntax.md` next to the new sentence.

tests/test_providers/test_pydantic_ai_interrupt.py
```diff
@@ -137,6 +137,40 @@ async def test_last_call_input_tokens_populated_on_partial_path(self) -> None:
         assert outcome.last_call_input_tokens > 0
 
 
+class TestMessageHistoryContinuation:
```
jrob5756, line 180: **RECOMMENDED**: Continuation is tested only against a plain-text agent, and three other new branches are unpinned.

`TestMessageHistoryContinuation` uses `_make_text_agent()` (`output_type=str`). But `validator:` only applies to agents with an `output:` schema, so in production the continued history always ends in a `ToolCallPart`/`ToolReturnPart` pair for the output tool. That's the case most likely to break on a pydantic-ai upgrade, and the only one that matters here. I confirmed it currently works (history 3 messages → 6, prior messages preserved, no duplicated system prompt), so this is cheap to add and locks in working behavior.

Three further branches are unpinned, each confirmed undetected by mutation:

- **The skills/instructions skip.** Re-adding `self._build_prompt_prefix(...)` to the continuation branch is undetected, because the fixture agent has no skills and no workspace instructions, so the prefix is `""`. Half the documented contract is unverified, and there's no executor-level test for `continuation_state` at all — `grep continuation_state tests/` returns three files, none under `tests/test_executor`.
- **Interrupt before the first iteration with history.** Reverting `list(message_history or [])` to `[]` (`interrupt.py:227`) is undetected. Worth flagging while writing it: on that path `user_prompt` — the validator feedback — gets silently dropped, so the model is asked for a partial result based on the conversation that just failed validation. It fails safe (the engine discards partial re-runs), but the test should assert it so it's a decision rather than an accident.
- **The "never serialized" invariant.** AGENTS.md asserts `continuation_state` never reaches a checkpoint or event log. It holds today, but `checkpoint.py::_make_json_serializable` `str()`s unknown objects, so a future leak wouldn't crash — it would silently write a giant stringified message list into every checkpoint.

**Suggested fix:** Add: a `ToolOutput`-based twin of `test_message_history_is_followed_by_new_user_turn`; an executor-level test using an agent with a non-empty prompt prefix asserting `captured_prompt == "FEEDBACK"` exactly; an interrupt-with-history case in `TestInterruptBeforeRun`; and a validator-retry test whose provider returns `continuation_state=object()`, asserting the saved checkpoint still `json.dumps` cleanly and no emitted event payload contains the sentinel.

jrob5756 (commented):
This PR wires `message_history` continuation through the validator retry path for Claude and OpenAI, and the happy path works. But the two things that make it safe long-term are both missing: nothing pins that the continuation state actually reaches the provider, and nothing stops a future provider from claiming continuation support without being able to honor the prompt it's handed. Both are blocking.

**Blocking:**
- `b1` — the inbound `message_history` forwarding (`runner.py:178`, `openai.py:1053`, `claude.py:1220`) has no test; deleting all three leaves the suite green.
- `b2` — the executor discards the rendered prompt on the continuation path with no guard on whether the provider can actually use it (`executor/agent.py:431`).

Everything else is recommended: prompt/event visibility on retries, an unguarded duplicate call site, an `Any`-typed field at a boundary that needs real typing, stale docstrings, an unexplained conditional, a context-growth risk with a near-invisible failure mode, and test coverage that only covers the plain-text case rather than the schema-based one validator retries actually hit. Detail on all of those is inline.

One finding has no single line to anchor to:

### RECOMMENDED — AGENTS.md gains no Provider Parity bullet for a change to the `AgentProvider.execute` contract

AGENTS.md says any change to a provider's behavior, contract, or capabilities must be applied to all providers, backed by a list meant to stay current. This PR changed the `AgentProvider.execute` abstract signature and added a field to `AgentOutput`, both squarely inside that contract, and added no bullet. Two existing bullets are exact structural precedents: the plugin-components bullet (a new `execute()` kwarg) and the context-window-measurement bullet (a new `AgentOutput` field with an explicit "or leave it `None`" rule).

The result is that "When modifying any provider, check all other providers for the same change" has nothing to check against for this field. The only statement of the rule lives in a Key Patterns bullet about the validator, which nobody adding a seventh provider is going to read. The provider roster is now duplicated in four places (`AGENTS.md:134`, `AGENTS.md:223`, `docs/workflow-syntax.md:1230`, `config/schema.py:729`), none of them near `runner.py:222`, the one line that actually decides it, and none covered by a test.

Separately, `config/schema.py:726-730` is user-facing and now contradicts itself: sentence one says the feedback is "appended to its prompt," sentence three says the prompt isn't re-sent at all. It also leaks the internal label "Pydantic AI providers" into a config schema docstring — a workflow author who writes `provider: claude` has no way to know their provider is "Pydantic AI"-based. The sibling text in `docs/workflow-syntax.md` gets this right.

**Suggested fix:** Add a parity bullet alongside the `last_call_input_tokens` one: a provider that can resume a completed run continues it and treats `rendered_prompt` as the next user turn; a provider that can't must ignore the kwarg and leave `AgentOutput.continuation_state` at `None`, the first-class "rebuild the prompt statelessly" signal the executor branches on. State that the value is provider-opaque, must never be handed to a different provider, and must never be serialized. Then rewrite the `schema.py` paragraph provider-neutrally: "On the `claude` and `openai` providers the re-run continues the completed conversation and the feedback is the next user turn; other providers rebuild the prompt with the feedback appended."

src/conductor/providers/_pydantic_ai/runner.py
hertznsk: Fixed in `2fc443f`. Both Pydantic AI providers now have an inbound test that spies on `run_with_interrupt` and asserts the exact history object arrives as `message_history`, the follow-up text arrives as `user_prompt`, and the continued run's messages extend the first run's (`test_openai.py::TestExecute::test_execute_continuation_state_reaches_the_pydantic_run`, same for Claude in `test_pydantic_ai_provider.py`). I re-ran your mutation check to confirm the pin works: replacing `message_history=continuation_state` with `None` in `openai.py` fails the new test immediately.

src/conductor/executor/agent.py
hertznsk: Fixed in `b9fe67b`. `AgentProvider.supports_continuation` (default `False`, alongside the two existing `supports_native_*` properties as you suggested) is now declared `True` on `ClaudeProvider` and `OpenAIProvider`, and `AgentExecutor` raises `ExecutionError` when handed continuation state from a provider that does not declare it — so the prompt is never discarded on a coincidence. Paired tests: the guard and the exact follow-up-only prompt at executor level (with a non-empty `instructions_preamble`, so the prefix skip is verified, not just the template skip), class-level declarations pinned for all six providers, and `test_fail_then_rerun_succeeds` now asserts `primary_prompts[1].startswith(primary_prompts[0])` so the stateless path can no longer pass with a feedback-only stub.

One follow-up on the roster: `hermes` turned out to have a real continuation surface (`run_conversation(conversation_history=...)`, the same mechanism its checkpoint resume already uses), so `d90ef63` declares it there too, with the state forwarded as `conversation_history` and withheld when a parse-recovery call produced the output (that output lives in the recovery conversation the outer result never carries).

src/conductor/executor/agent.py
hertznsk: Fixed in `3f19fb9`. `agent_prompt_rendered` now carries `continuation: true` on the continuation path, and the store appends the turn to the node's existing prompt (top-level and for-each item) instead of replacing it, so the Prompt panel shows the original task plus the feedback turn. Covered by two new store tests (continuation appends, non-continuation still replaces) and an executor-level assertion on the event payload; static assets rebuilt via `make build-frontend`.

src/conductor/executor/agent.py
hertznsk: Fixed in `632594c`. Everything after `rendered_prompt` is now keyword-only on `AgentProvider` and all six providers, `continuation_state` sits at the end of the signature (same move in `AgentExecutor.execute`), and the two call arms collapsed into one unconditional call. Test doubles updated to match, and the one genuinely positional caller — the ACA runner's inner `provider.execute` — now passes `tools=` by keyword.

src/conductor/providers/base.py
hertznsk: Fixed in `f9d463f`, taking your minimum-change option: the field and every parameter are now `object | None`, and the two Pydantic AI call sites re-narrow with a documented `cast("Sequence[ModelMessage] | None", ...)` noting the value can only be this provider's own `all_messages()` list. `make check` (ruff + `ty`) is clean.

On the stronger `ContinuationState(owner=...)` wrapper: I left it out deliberately. The cross-provider feed is not reachable today (`_apply_validator` reuses the same executor), and the `supports_continuation` guard from the other thread turns a provider that returns state it cannot resume into a loud `ExecutionError`. If a future change ever lets state cross providers, the wrapper becomes the right enforcement point — happy to add it then.

src/conductor/executor/agent.py
hertznsk: Fixed in `7b24e95`. Every site you listed is updated: `AgentExecutor.execute`'s summary and both parameter entries (including the skipped-prefix consequences and the `supports_continuation` requirement), `_apply_validator` step 4 and its `guidance_section` entry, `_build_validation_feedback` (now documents the newline prefix is for the append case), and the four ignoring providers each got an "Ignored." `Args` entry plus a `del continuation_state`, matching the `aca.py` convention.

src/conductor/engine/workflow.py
hertznsk: Fixed in `4e88acd`. The selection is hoisted to a named `rerun_guidance` local above the `try:`, with both branches commented: the stateless path appends the feedback after any existing guidance (whole prompt is rebuilt anyway), the continuation path sends the lstripped feedback alone because the provider-held conversation already contains the rendered prompt and the guidance — and `new_guidance` no longer exists to look dead on the continuation path.

src/conductor/engine/workflow.py
hertznsk: Fixed in `46b3c09`. The rerun-errored `agent_validation_failed` emission now carries `error` (`"RuntimeError: rerun boom"` style) and a `continued` flag; the console's `rerun_errored` branch prints the cause, and the dashboard activity entry shows it too. The context-growth trade-off is documented in `docs/workflow-syntax.md` right next to the continuation sentence: the retry replays the entire first conversation including tool exchanges, so an agent that already burned most of its window can overflow where a rebuilt prompt used to fit. Tests assert the payload on both the stateless and continuation fail-open paths.

tests/test_providers/test_pydantic_ai_interrupt.py
hertznsk: Fixed across `c108d75` and `b9fe67b`:

- **Structured twin**: `test_structured_output_history_is_continued` uses a `ToolOutput` agent whose model answers through the output tool, so the continued history ends in the `ToolCallPart`/`ToolReturnPart` pair production always has. (Your 3→6 message shape is exactly what it asserts. Side note: `_make_structured_agent` could not be reused — its `FunctionModel` answers in plain text, which only ever exercises the interrupt partial path, never a completed `ToolOutput` run.)
- **Skills/instructions skip**: executor-level test with a non-empty `instructions_preamble` asserting `captured_prompt == "FEEDBACK"` exactly (`b9fe67b`).
- **Interrupt before the first iteration with history**: pinned as a decision — the partial request is based on the prior conversation and the feedback prompt never reaches the model, which fails safe because the engine discards partial re-runs.
- **Never serialized**: a validator-retry test whose provider returns `continuation_state=object()` asserts no emitted event payload contains the sentinel, every payload still `json.dumps`-clean, and a checkpoint saved after the retry round-trips through JSON with no trace of it.

hertznsk:
Thanks for the thorough review — every finding was valid. All are addressed, rebased onto current `main` (`544a2bf`), and pushed. The two blocking items each have their own thread; the review-body findings:

**AGENTS.md parity bullet** — added in `7b24e95`, alongside the `last_call_input_tokens` one, using your framing: a provider that can resume declares `supports_continuation=True`, populates `AgentOutput.continuation_state` on completed outputs, and treats `rendered_prompt` as the sole next user turn; a provider that cannot ignores the kwarg (docstring entry + `del`) and leaves the field `None`. The three invariants are stated: provider-opaque and `object`-typed at the boundary, never handed to a different provider, never serialized to checkpoints or event logs.

**`config/schema.py` self-contradiction** — rewritten provider-neutrally per your suggestion ("On the `claude` and `openai` providers the re-run continues the completed conversation…; other providers rebuild the prompt with the feedback appended"), and the "Pydantic AI providers" label is gone from user-facing text.

**Roster update** — one correction to a claim I made: `hermes` *does* have a working continuation surface (`run_conversation(conversation_history=...)`, the same mechanism its checkpoint resume already ships), so `d90ef63` wires it as a third supporting provider, with the state withheld when a parse-recovery call produced the output. The parity bullet and validator docs name all three providers (`claude`, `openai`, `hermes`).

Everything is verified locally: `make check` clean, and the full suite passes (7839 + 790 fleet tests; the only failures are the four pre-existing environment ones noted in the PR body — two POSIX unreadable-path assertions and the two Claude performance thresholds).

jrob5756 (approved):

PR #512, opened 2026-09-08
fix(providers): prevent concurrent registry creation

ldavidgomez:
## Summary

Fixes a race in `ProviderRegistry` where concurrent requests for the same provider type could each construct a separate provider instance before either result reached the cache.

This is the first, independently mergeable foundation PR in a planned sequence of Claude Agent SDK hardening work. Its scope is intentionally limited to provider-instance correctness; it does not include authentication, cancellation, billing, tools, or SDK behaviour changes.

## Behaviour

- Concurrent requests for the same provider type share one constructed and cached instance.
- Cached provider reads remain lock-free.
- Provider construction, session-resume wiring, and cache publication are protected as one operation.
- Construction failures do not populate the cache or strand concurrent waiters.
- Different provider types remain distinct. Construction is intentionally serialized by one registry-local lock, following the approved design.

## Validation

- Added deterministic concurrency tests that reproduce the original unlocked race and prove the fix.
- `tests/test_providers/test_registry.py`: 21 passed.
- Lint and formatting passed locally.
- The session-key integration test passed on the clean base as non-regression evidence.

## CI status

This PR is opened as a draft pending the required Linux CI validation: lint, type checking, and the non-`real_api` / non-`performance` suite.

## Scope exclusion

No Claude authentication-preflight, schema, factory, diagnostics, billing, tools, examples, or AgentOps planning artifacts are included.


ldavidgomez:
Linux CI is the remaining validation condition. The workflow is currently awaiting maintainer approval, as required for external contributions.

jrob5756 (approved):

PR #518, opened 2026-09-09
feat: add deterministic type: mcp workflow step

### LINKED ISSUE #392: Idea: add a deterministic `type: mcp` step for direct MCP tool calls
## Summary

Add a first-class deterministic workflow step that calls one tool on a configured MCP server directly, without asking an LLM agent to choose or invoke the tool.

This would let workflows use the existing MCP integration ecosystem for operations such as querying issue trackers, updating repository metadata, reading databases, or calling internal services while keeping the operation visible and reviewable in YAML.

This is an idea for discussion, not a commitment to a particular final syntax.

## Motivation

Conductor is strongest when probabilistic work and deterministic orchestration are kept separate. Today, deterministic integration work generally falls back to `type: script`:

```yaml
- name: update_tracker
  type: script
  command: uv
  args: [run, python, -m, scripts.update_tracker]
```

That works, but the actual operation moves out of the workflow into Bash, Python, or TypeScript code. A reviewer can no longer understand the full workflow from the YAML alone, and each repository must own and maintain its integration glue.

Platforms such as n8n address this with native integrations, but building and maintaining a comparable catalog specifically for Conductor would be expensive. MCP already provides a standardized tool discovery and invocation surface across a large ecosystem of servers. A direct MCP step could therefore serve as a broad integration escape hatch while native steps remain available for common operations where Conductor can provide a better contract and user experience.

## Proposed MVP shape

```yaml
workflow:
  runtime:
    mcp_servers:
      tracker:
        type: stdio
        command: tracker-mcp
        env:
          TRACKER_TOKEN: ${TRACKER_TOKEN}

agents:
  - name: create_issue
    type: mcp
    server: tracker
    tool: create_issue
    arguments:
      project_id: 105
      subject: "{{ triage.output.title }}"
      description: "{{ triage.output.details }}"
    routes:
      - to: notify
```

The step should:

- invoke exactly one named tool without an LLM call;
- recursively render `arguments` against workflow context while preserving JSON-native values rather than coercing every value to a string;
- verify that `server` exists in `runtime.mcp_servers`;
- verify that the tool is permitted by the servers configured `tools` allowlist;
- expose the MCP result in workflow context for downstream templates and routes;
- support the normal step lifecycle in the main loop, parallel groups, and `for_each` where safe;
- emit dedicated started/completed/failed events for CLI, JSONL, dashboard, and replay visibility;
- apply the existing per-result MCP output limit where applicable.

A possible output envelope is:

```yaml
create_issue.output:
  content: [...]       # MCP content blocks, represented as JSON-safe data
  structured: {...}    # structuredContent when supplied by the server
  is_error: false      # MCP tool execution error flag
```

The exact output shape needs discussion. It should preserve structured MCP data rather than flattening everything to text. Optional `output:` validation could apply to a documented portion of that envelope.

## Why this appears feasible in the current architecture

Conductor already has most of the required building blocks:

- `runtime.mcp_servers` and `MCPServerDef` describe stdio, HTTP, and SSE servers.
- `src/conductor/mcp/manager.py::MCPManager` is a host-side MCP client that can connect, discover tools, call tools, manage process lifetime, and apply tool-output limits.
- `WorkflowEngine` already dispatches non-LLM step types such as `script`, `set`, and `wait` and stores their typed outputs in `WorkflowContext`.
- Existing routing, limits, checkpointing, event emission, and output validation can be reused around a new executor.

The direct step should be owned by the workflow engine/executor layer, not by an LLM provider. Current providers expose MCP through different mechanisms: Claude uses the host-side `MCPManager`, while Copilot and Claude Agent SDK delegate MCP configuration to their SDK/CLI. A deterministic step must have one provider-independent execution path and must also work in workflows whose LLM provider does not support MCP tools.

### No transactional guarantee

MCP standardizes tool invocation, not the semantics or atomicity of the external operation. Closing an MCP process or connection does **not** imply rollback. A successful tool may have committed an external side effect before Conductor receives, stores, or checkpoints the result.

Likewise, if a workflow is cancelled after a mutating tool call but before the step is committed to context, resume may invoke it again. The step therefore has at-least-once risk around interruption and resume unless the tool itself supports idempotency keys or reconciliation.

### No implicit retries in the MVP

Conductor cannot infer whether an arbitrary MCP tool is safe to repeat. Automatic retries could duplicate issues, comments, payments, commits, or other mutations. The first version should not retry tool calls implicitly. Any future retry support should be explicit and documented around idempotency.

### Start with the transport the shared client actually supports

Although `MCPServerDef` accepts stdio, HTTP, and SSE, the shared host-side `MCPManager` currently implements stdio only. A coherent first implementation could therefore support stdio and return a clear validation error for other transports. HTTP/SSE can follow by extending the shared manager while preserving the step syntax.

### Secrets and observability need an explicit policy

MCP does not redact secrets for Conductor. Tool arguments may contain credentials or sensitive payloads, and the current tool event formatting includes arguments. A direct step should not log full arguments or results by default. Events can safely include server/tool names, argument keys, elapsed time, result size, and truncation metadata; displaying values should require an explicit, redaction-aware policy.

### Concurrency needs a defined ownership model

A workflow-owned connection pool could keep one stdio server alive and reuse it across direct calls. Calls sharing the same server/session should initially be serialized unless client/server concurrency is known to be safe; different servers or working directories can proceed independently. Cleanup must be bounded by workflow lifecycle and robust under cancellation.

## Why `sequence` is not part of the MVP

A special multi-call sequence syntax is tempting for stateful servers, but it would immediately add nested scoping, conditional execution, partial failure policy, cleanup hooks, and misleading expectations of transactionality.

Ordinary workflow steps already express ordered calls and routing. Connection reuse can be an executor lifecycle detail, so consecutive `type: mcp` steps may share the same live stdio server without introducing another programming language inside YAML.

If real servers demonstrate a need for an indivisible connection-scoped block, sequence syntax can be discussed later as non-transactional convenience sugar.

## Relationship to #222 (`type: http`)

This proposal complements rather than replaces #222.

A native HTTP step remains preferable for direct REST calls because Conductor can provide HTTP-specific behavior that a generic MCP tool cannot guarantee consistently:

- typed request and response bodies;
- status codes, headers, and non-2xx payloads as a stable output contract;
- method-aware retry defaults;
- standard authorization-header redaction;
- lower startup and serialization overhead;
- no dependency on the quality, lifecycle, or security behavior of a third-party MCP HTTP client server.

The MCP step is the broad integration escape hatch for existing MCP tools. The HTTP step is the purpose-built primitive for a very common protocol.

### Phase 1

- one `type: mcp` step equals one tool call;
- stdio servers through an engine-owned executor/client lifecycle;
- typed recursive argument rendering;
- structured result preservation and optional output validation;
- no implicit retry and no sequence syntax;
- safe events, cancellation cleanup, routing, parallel/for-each support with per-session serialization;
- validation that the server exists and the tool is allowed.

### Phase 2

- host-side Streamable HTTP/SSE transport support;
- transport-specific authentication and lifecycle handling;
- richer static validation using discovered tool schemas where practical.

### Later, only if justified by real workflows

- explicit idempotency-aware retry policy;
- connection-scoped sequence convenience syntax;
- declarative redaction annotations for selected argument/result paths.

## Open questions

- Should the step type be named `mcp`, `mcp_call`, or `mcp_tool`?
- What stable JSON-safe representation should be used for MCP content blocks and `structuredContent`?
- Should `isError: true` become a failed workflow step by default, or remain routable output with an opt-in fail mode?
- Should a direct step reuse a long-lived workflow-level server process, or should lifecycle be scoped more narrowly by server and working directory?
- How should direct steps and provider-managed MCP servers avoid duplicate stdio processes when both use the same configuration?
- Is the Python MCP SDK promoted to a required dependency, or should this step live behind an installation extra?
- Which event fields can be shown safely without exposing tool arguments or results?

## Acceptance criteria for an eventual MVP

- [ ] A workflow can call a configured stdio MCP tool without invoking an LLM provider.
- [ ] Arguments retain JSON-native types after template rendering.
- [ ] Structured MCP results remain structured and are available to routes and downstream steps.
- [ ] Unknown servers, disallowed tools, unsupported transports, and invalid step-only field combinations fail clearly.
- [ ] No automatic retries occur for tool calls.
- [ ] Cancellation closes owned resources best-effort and documents that external side effects may remain.
- [ ] Same-session calls are concurrency-safe, with serialization as the conservative default.
- [ ] Events and logs do not expose full argument/result values by default.
- [ ] The step works independently of the selected LLM provider.
- [ ] An example demonstrates an MCP call followed by normal conditional routing.

Related: #222.

hertznsk:
Closes #392.

## What

A new workflow step type, `type: mcp`, that calls a single tool on a configured MCP server directly — no LLM in the loop, zero prompt tokens, fully deterministic. The tool result lands in the workflow context as a structured envelope and can drive routing, including branching on the tool's `isError` flag.

```yaml
- name: check_status
  type: mcp
  server: tracker
  tool: get_issue
  arguments:
    issue_id: "{{ workflow.input.id }}"
  routes:
    - when: "{{ check_status.is_error }}"
      to: handle_error
    - to: $end
```

## Key design decisions

- **Engine-owned connection pool.** The step does not depend on any provider's MCP support: the engine lazily manages its own stdio `MCPManager` pool keyed by (server, runtime working dir), with per-server `asyncio.Lock` serialization (so parallel/for_each members calling the same server don't interleave), a bounded pool with idle eviction, and unconditional cleanup in a `finally` on `run()`/`resume()`. Fail-fast groups cancel and drain in-flight siblings before pool cleanup.
- **`is_error` is data, not failure.** A tool error is routable data in the envelope (like `exit_code` for script steps). Only transport-level failures fail the step.
- **Redacted event contract.** `mcp_started`/`mcp_completed`/`mcp_failed` carry only metadata (server, tool, argument *keys*, elapsed, is_error, result_bytes, truncated, spill_path) — never argument or result *values* — in console, JSONL, web replay, and fleet summary. The existing string-based `MCPManager.call_tool` is untouched; a new `call_tool_structured` is added alongside it.
- **Off-network static validation.** `conductor validate` checks that the server is declared, the tool is allowlisted by the server's `tools:` filter, the transport is stdio (a clear error for http/sse — deferred to a future phase), and argument templates parse.
- **Structured envelope.** `{content: [...], structured: {...}|null, is_error: bool}` — `structured` keys are merged on top of the envelope without allowing the reserved envelope keys to be shadowed.

## Coverage

- Dispatch in all three engine positions: main loop, parallel groups, for_each
- Dashboard: `mcpNode` graph node + `McpDetail` panel (metadata only)
- Console, web resume/replay, and fleet summary surfaces
- Docs (`workflow-syntax.md`, `mcp-tools.md`), `examples/mcp-step.yaml`, bundled skill references, CHANGELOG
- Explicit non-goals: no http/sse transport, no automatic retries/idempotency policies, no multi-call sequence syntax, no per-step `working_dir`

## Verification

- Full test suite: 8798 passed (3 unrelated pre-existing environment failures: two permission-based unreadable-directory tests and one provider performance test)
- `make check` clean (ruff + ty; 6 pre-existing unrelated warnings)
- `make test-frontend`: 159 passed; `make build-frontend`; `make validate-examples`
- End-to-end QA against a real stdio MCP server: envelope routing on `is_error`, `--web-bg` dashboard, resume after a transport failure, and fleet visibility


src/conductor/engine/workflow.py
```diff
@@ -593,6 +687,141 @@ def _workflow_dir(self) -> Path | None:
...
+            if target is None:
+                logger.debug(
+                    "MCP step pool at cap %d with every entry locked; allowing overflow",
+                    _MCP_STEP_POOL_MAX,
+                )
+                return
+            evicted_key = target
+            evicted = self._mcp_step_managers.pop(target)
+        try:
+            await evicted.close()
```
jrob5756: **BLOCKING**: `MCPManager.close()` swallows `CancelledError` while it drains connection teardown. Awaiting it directly during eviction means that cancellation gets absorbed here instead of propagating — once cleanup finishes, `_get_mcp_step_manager()` just carries on connecting and invoking the next tool. So a workflow cancellation, or a fail-fast sibling failure, can still result in a new external side effect starting after the fact.

Run the eviction cleanup in its own shielded task. If cancellation arrives while it's running, keep draining cleanup (tolerating repeated cancellation), then re-raise `CancelledError` before this code goes on to connect or invoke anything.

src/conductor/executor/mcp_step.py
```diff
@@ -0,0 +1,229 @@
...
+        structured = envelope.get("structured")
+        if isinstance(structured, dict):
+            shadowed = set(structured) & _RESERVED_ENVELOPE_KEYS
+            if shadowed:
+                logger.debug(
+                    "MCP step '%s' structured content shadows envelope fields: %s",
+                    agent.name,
+                    ", ".join(sorted(shadowed)),
+                )
+            envelope.update(
```
jrob5756, line 212: **BLOCKING**: A structured response like `{"outputs": [1, 2], "errors": []}` gets flattened straight into the envelope, but `WorkflowContext` and for-each source resolution use those exact two keys to identify group outputs. That means this result gets misclassified and loses its normal `.output` wrapper — something like `{{ call.output.is_error }}` breaks even though the tool call succeeded. Affects `accumulate`, `last_only`, and `explicit` context modes alike.

Engine-owned step metadata should decide what counts as a group output, not the payload's own keys — that includes for-each source resolution. Alternatively, reserve `outputs`/`errors` from flattening and keep them under `structured`, documenting the restriction.

src/conductor/web/server.py
```diff
@@ -1007,6 +1007,48 @@ def _synth_agent_or_script(
             }
             return "set_started", started_data, "set_completed", completed_data
 
+        if agent_type == "mcp":
```
jrob5756, line 1184: **BLOCKING**: The metadata-only synthetic replay path handles standalone MCP steps, but `replay_synthetic_from_context()` routes saved parallel and for-each entries through `_synth_parallel()`/`_synth_for_each()` instead, and those publish the complete saved output — MCP content and structured values included. So resuming without the original JSONL log exposes results in dashboard history and `/api/state` that live execution deliberately never showed.

Group synthesis needs to know its members' step types, strip MCP envelopes from the aggregate replay events, and synthesize metadata-only member/item events instead. Needs to cover both mixed parallel groups and MCP for-each groups.

src/conductor/engine/workflow.py
```diff
@@ -1417,6 +1646,246 @@ async def _run_set_step(self, agent: AgentDef, agent_context: dict[str, Any]) ->
...
+                agent_name=agent.name,
+            ) from None
+        elapsed = _time.time() - start
+
+        content = envelope.get("content")
+        blocks = content if isinstance(content, list) else []
+        truncated = any(isinstance(block, dict) and block.get("truncated") for block in blocks)
+        spill_path = next(
+            (
+                block.get("spill_path")
```
jrob5756: **BLOCKING**: `call_tool_structured()` preserves content-block extension fields, and this code treats any block's `spill_path` as if it were metadata Conductor generated itself. A server can return `spill_path={"private_result": "value"}` and that object gets copied straight into `mcp_completed`, even with spilling disabled — which both leaks result values through what's supposed to be a metadata-only event and violates the frontend's string type for that field. Synthetic replay repeats the same extraction, and server-supplied truncation fields get the same blind trust.

Keep locally generated truncation/spill metadata separate from whatever the server sent back, or strip those reserved field names before adding local metadata. Live and replay events should build the same trusted, type-checked representation.

jrob5756 (commented):
Six blocking issues here, mostly clustered around the same theme: the new MCP step type leaks more than it should (server-controlled data into trusted metadata, saved results into replay, raw errors past the debug-log boundary that doesn't actually exist for the CLI), plus a real cancellation hole and a namespace collision in structured output. This isn't close to mergeable as-is. The recommended items are mostly follow-up gaps in interrupt handling, pool accounting, validation parity, and docs that promise more than the implementation delivers.

- `engine/workflow.py:806` — eviction cleanup swallows cancellation, letting a new tool call start after the workflow was cancelled
- `executor/mcp_step.py:153` — `outputs`/`errors` keys in a structured result collide with context's own output-shape keys
- `web/server.py:1010` — replaying a checkpoint without the original JSONL synthesizes full MCP results into parallel/for-each groups instead of metadata-only
- `engine/workflow.py:1867` — a server can put arbitrary data in `spill_path` and it gets forwarded as trusted event metadata
- `engine/workflow.py:1840` — failures point users at debug logs that `--log-file` never actually populates
- `engine/workflow.py:752` — `connect_server()`'s own exception handler logs raw server output at ERROR, bypassing the sanitization the rest of the step does

In addition, it looks like some merge conflicts crept in.

src/conductor/engine/workflow.py
```diff
@@ -1417,6 +1646,246 @@ async def _run_set_step(self, agent: AgentDef, agent_context: dict[str, Any]) ->
...
+            # from the SDK / mcp_auth / connect / renderer code (e.g.
+            # SystemExit("secret")) would otherwise bypass this handler and
+            # reach the workflow's outer except BaseException, publishing the
+            # value via str(e) in workflow_failed. Trade-off: a genuine
+            # SystemExit / KeyboardInterrupt raised inside an mcp step is
+            # reported as a redacted step failure instead of propagating as a
+            # base exception — the no-values policy outranks base-exception
+            # classification for this step type, and the engine converts real
+            # keyboard interrupts before this point.
+            logger.debug("MCP step '%s' failed", agent.name, exc_info=True)
```
jrob5756: **BLOCKING**: Transport, rendering, timeout, and output-validation failures all get replaced with "see debug logs for details," and the actual exception only goes to `logger.debug()`. But `--log-file` wires up a Rich console, not a Python logging handler, and doesn't turn DEBUG on. So even someone who explicitly asked for a log file gets nothing to diagnose the failure with.

Add an explicit private diagnostic sink for both run and resume before the exception details get discarded. Keep raw values out of lifecycle events, but preserve safe categories like timeout (and its duration), and have the user-facing message actually point at wherever the diagnostic ends up.

src/conductor/engine/workflow.py
```diff
@@ -593,6 +687,141 @@ def _workflow_dir(self) -> Path | None:
...
+            "tools": server_def.tools,
+        }
+        if server_def.env:
+            server_config["env"] = server_def.env
+        if server_def.timeout:
+            server_config["timeout"] = server_def.timeout
+        resolved = await resolve_mcp_server_config(server_name, server_config)
+
+        manager = MCPManager(tool_output=self.config.workflow.runtime.tool_output)
+        await manager.connect_server(
```
jrob5756: **BLOCKING**: This new connection path calls `MCPManager.connect_server()`, and that method's own exception handler logs raw exception text plus a traceback at ERROR — before `_run_mcp_step()` gets a chance to sanitize whatever's being propagated. That means server-supplied diagnostic values can land in stderr and background capture logs with DEBUG turned off, which breaks the debug-only treatment this feature is supposed to have for connection failures specifically.

Give deterministic MCP connections a redacted logging path: safe connection metadata at normal log levels, raw exceptions only to an explicitly enabled private diagnostic sink. Don't change existing provider logging behavior where it's relied on elsewhere.

src/conductor/engine/workflow.py
```diff
@@ -1417,6 +1646,246 @@ async def _run_set_step(self, agent: AgentDef, agent_context: dict[str, Any]) ->
...
+                        resolved_cwd,
+                        raw_cwd,
+                    )
+                    raise _McpStepRedactedCheckError(
+                        f"MCP step '{agent.name}': runtime working_dir does not exist "
+                        f"or is not a directory",
+                        agent_name=agent.name,
+                    )
+
+            async with await self._mcp_step_slot(server):
```
jrob5756: **RECOMMENDED**: Dashboard Stop sets `interrupt_event` rather than cancelling the engine task, but neither slot acquisition, connection setup, nor tool execution checks that event. So a blocked MCP call just keeps waiting, and once it finishes, the root web-mode `_check_interrupt()` clears the flag assuming provider-level interruption already handled it — the workflow can then move on to the next step instead of pausing.

Wire the MCP path into `interrupt_event` and the dashboard pause flow, including the waits for a slot or a connection. Don't auto-replay a call that may have already had a side effect. Needs to cover the actual Stop path on both run and resume, not just `task.cancel()`.

src/conductor/engine/workflow.py
```diff
@@ -593,6 +687,141 @@ def _workflow_dir(self) -> Path | None:
...
+        # Lazy import: conductor.mcp.__init__ eagerly imports the MCP SDK, and
+        # this module is imported by cli/app.py on every conductor invocation.
+        from conductor.mcp.manager import MCPManager
+
+        key = (server_name, resolved_cwd)
+        manager = self._mcp_step_managers.get(key)
+        if manager is not None:
+            return manager
+
+        if len(self._mcp_step_managers) >= _MCP_STEP_POOL_MAX:
```
jrob5756: **RECOMMENDED**: Capacity gets checked before the awaited connection work runs, with no reservation for connections still in flight. With 15 idle managers already cached, two connections to distinct servers can both pass this check and leave 17 managers cached with nothing evicted — separate from the intentional busy-pool overflow, and the oversized pool never gets trimmed back down afterward.

Reserve space for pending connections as part of atomic admission under the pool guard, releasing the reservation on failure or cancellation, and keep the actual connection I/O outside the guard. If the busy-pool overflow is intentional, document it as a soft threshold rather than a hard bound.

src/conductor/engine/workflow.py
```diff
@@ -1417,6 +1646,246 @@ async def _run_set_step(self, agent: AgentDef, agent_context: dict[str, Any]) ->
...
+                )
+            server_def = mcp_servers[server]
+            if server_def.type != "stdio":
+                raise _McpStepRuntimeCheckError(
+                    f"MCP step '{agent.name}': server '{server}' has type "
+                    f"'{server_def.type}'; type: mcp supports stdio servers only "
+                    f"(http/sse support is not implemented yet)",
+                    agent_name=agent.name,
+                )
+            if server_def.tools != ["*"] and tool not in server_def.tools:
```
jrob5756: **RECOMMENDED**: Static validation treats any `'*'` in `server_def.tools` as unrestricted access, but the runtime only recognizes the exact list `['*']`. So `tools: ['*', 'health']` passes validation for an echo step and then fails at execution, before it even connects.

Use the same wildcard-membership rule at both the validation and runtime boundaries, or reject mixed wildcard lists at the schema level. Worth covering singleton wildcard, mixed wildcard, explicit include/exclude, and empty lists with shared validation/runtime test cases.

src/conductor/config/validator.py
```diff
@@ -1284,6 +1369,13 @@ def _collect_template_strings(
         for key, expr in values.items():
             templates.append((f"agent '{agent.name}' values.{key}", expr))
 
+    # 'mcp' step arguments — values are Jinja2-rendered at runtime, including
+    # string leaves nested in dicts/lists, so collect them recursively to
+    # catch stale references at validate-time like every other rendered field.
+    arguments: dict[str, Any] | None = getattr(agent, "arguments", None)
+    if arguments:
+        templates.extend(_collect_argument_strings(f"agent '{agent.name}' arguments", arguments))
```
jrob5756, line 1399: **RECOMMENDED**: Argument strings only get fed into reference analysis, and `_extract_template_refs()` catches `TemplateSyntaxError` and just returns no references. So an argument like `'{{ workflow.input.foo'` passes `conductor validate` and only fails at execution — the offline syntax check this feature advertises for nested MCP arguments isn't actually happening.

Parse each collected argument template explicitly for syntax errors before running reference analysis, reporting the step and the nested argument path. Include malformed templates nested inside both mappings and lists.

src/conductor/web/frontend/src/stores/workflow-store.ts
```diff
@@ -1905,6 +1923,99 @@ const eventHandlers: Record<string, (state: MutableState, data: Record<string, u
...
+                mcp_spill_path: data.spill_path,
+              }
+            : i
+        );
+      }
+      replaceNode(t.nodes, data.group_name);
+    } else {
+      const nd = ensureNode(t.nodes, data.agent_name, 'mcp');
+      nd.status = 'completed';
+      t.incrCompleted();
```
jrob5756: **RECOMMENDED**: Parallel MCP completion events carry `group_name` but no `item_key`, so they fall into this branch and increment workflow completion — but `parallel_completed` increments the same counter again for the group. A single parallel group with one MCP member ends up reporting 2/1 completed.

Only increment workflow completion here for standalone MCP steps. Let `parallel_agent_completed` handle member progress and `parallel_completed` handle the top-level group.

src/conductor/web/frontend/src/components/graph/graph-layout.ts
```diff
@@ -489,6 +489,7 @@ function layoutContext(
     let flowNodeType = 'agentNode';
     if (nodeType === 'script') flowNodeType = 'scriptNode';
     else if (nodeType === 'set') flowNodeType = 'setNode';
+    else if (nodeType === 'mcp') flowNodeType = 'mcpNode';
```
jrob5756: **RECOMMENDED**: The MCP node mapping only applies to standalone layout — parallel children are still hardcoded as `agentNode`/type `'agent'`, and workflow initialization creates their stored `NodeData` the same way. `ensureNode(..., 'mcp')` doesn't replace an existing type either, so selecting a parallel MCP member opens the generic detail panel instead of `McpDetail`.

Preserve declared member types during both root and nested-context initialization, and pick the parallel child renderer from those types, so parallel MCP members actually get `McpNode`/`McpDetail`.

src/conductor/web/frontend/src/stores/workflow-store.ts
```diff
@@ -1905,6 +1923,99 @@ const eventHandlers: Record<string, (state: MutableState, data: Record<string, u
...
+    if (data.group_name != null && data.item_key != null) {
+      const nd = ensureNode(t.nodes, data.group_name, 'for_each_group');
+      if (nd.for_each_items) {
+        nd.for_each_items = nd.for_each_items.map((i) =>
+          i.key === data.item_key
+            ? {
+                ...i,
+                status: 'completed',
+                elapsed: data.elapsed,
+                mcp_server: data.server,
```
jrob5756, line 1983: **RECOMMENDED**: This handler records MCP metadata on each item, but neither `GroupDetail` nor `ForEachItemRow` renders those fields or factors them into `hasDetails`. Since MCP item completion intentionally skips output, a metadata-only completed item can end up with a disabled detail toggle — no way to see its tool-reported error, server/tool name, result size, truncation, or spill path.

Include the MCP metadata in item expandability and render it in the item detail panel. Keep `is_error: true` items showing as completed but visibly flag the tool error, without exposing the raw result body.

tests/test_engine/test_mcp_step_groups.py
```diff
@@ -0,0 +1,1088 @@
...
+            await original_close()
+
+        engine._close_mcp_step_managers = recording_close  # type: ignore[method-assign]
+
+        blocking = asyncio.Event()
+
+        async def item_call(_server: str, _tool: str, arguments: dict[str, Any]) -> dict[str, Any]:
+            if arguments["q"] == "a":
+                # Fail only once the sibling is genuinely blocked inside its call.
+                await asyncio.wait_for(blocking.wait(), timeout=5)
```
jrob5756: **RECOMMENDED**: Item a holds the single-server slot while waiting on an event that item b only sets after it acquires that same slot — so the wait times out before the canary-bearing `RuntimeError` ever gets raised. Accepting any `ExecutionError` and just asserting the canary is absent doesn't actually establish the exception-redaction behavior this test is meant to cover.

Coordinate the sibling at the slot boundary rather than requiring simultaneous calls into one server. Assert the intended `RuntimeError` actually happened, that `mcp_failed` reports that type, and that sibling teardown happens before the pool closes.

plugins/conductor/skills/conductor/references/authoring.md
```diff
@@ -569,6 +569,68 @@ Set agents **cannot** have: `prompt`, `provider`, `model`, `tools`, `system_prom
...
+    timeout: 30                             # Optional: per-call timeout in seconds
+    routes:
+      - to: handle_error
+        when: "{{ output.is_error }}"
+      - to: analyze_spec
+```
+
+### Argument Rendering and Type Coercion
+
+Dict and list structures in `arguments:` are traversed recursively. String leaves are Jinja2-rendered against workflow context and auto-coerced (e.g. `"105"` -> `105`, `"true"` -> `True`, `"null"` -> `None`). Non-scalar values and embedded templates remain strings. YAML-native scalars (integers, floats, booleans, `None`) pass through untouched.
```
jrob5756: **RECOMMENDED**: This says non-scalar values and embedded templates stay strings, but `_coerce_auto()` YAML-parses every fully rendered string. `'[1, 2]'` becomes a list, `'1{{ x }}'` becomes the integer `12` when `x=2`, and `'label: {{ x }}'` becomes a mapping. Anyone following this doc as written can end up sending arguments with a type they didn't expect.

Describe the actual behavior: whole-render YAML parsing, including collection conversion and embedded templates whose rendered text parses as something else. The syntax guide and executor docstrings should say the same thing.

docs/workflow-syntax.md
```diff
@@ -1025,6 +1025,127 @@ Per-key typing on multi `values:` is not supported.
...
+**Composition:**
+
+- **Parallel groups:** MCP steps can run inside `parallel` groups. Invocations targeting distinct servers execute concurrently; invocations targeting the same server serialize on the server slot lock.
+- **For-each groups:** MCP steps can serve as the inline agent of a `for_each` group.
+- **Working directory:** The server process inherits the workflow's `runtime.working_dir`, rendered dynamically per execution. Step-level `working_dir` is not allowed on MCP steps.
+
+**Validation rules:**
+
+- **Static validation (`conductor validate`):** Validates workflows offline without connecting to servers. Checks that the referenced server is declared in `runtime.mcp_servers`, has `type: stdio`, allows the tool in its `tools:` filter, uses valid template syntax in `arguments`, and avoids referencing sibling parallel steps.
+- **Runtime validation (`conductor run`):** Repeats all static checks and additionally verifies that the tool actually exists on the live connected server.
```
jrob5756: **RECOMMENDED**: `conductor run` doesn't repeat all of the static checks. `_run_mcp_step()` checks server declaration, transport, tool allowlisting, and live tool existence, but it doesn't do parallel-sibling dependency analysis — so a configuration `conductor validate` would reject can still execute, using a sibling's previously stored output. The current wording implies skipping validation still preserves those guarantees, which isn't true.

Enumerate the checks that actually get repeated at runtime, and state plainly that `conductor validate` is still necessary for the offline template-reference and parallel-dependency diagnostics.

docs/workflow-syntax.md
```diff
@@ -1025,6 +1025,127 @@ Per-key typing on multi `values:` is not supported.
...
+
+**Events, errors, and secrets policy:**
+
+MCP steps emit three lifecycle events:
+- `mcp_started`: contains `agent_name`, `iteration`, `server`, `tool`, and `argument_keys` (sorted list of key names only).
+- `mcp_completed`: contains `agent_name`, `elapsed`, `server`, `tool`, `is_error`, `result_bytes`, `truncated`, and optional `spill_path`.
+- `mcp_failed`: contains `agent_name`, `elapsed`, `server`, `tool`, `error_type`, and a generic redacted `message`.
+
+Argument values and result payloads are never included in event data, ensuring credentials or sensitive parameters do not leak to event logs or dashboard streams. When a call fails, `mcp_failed` records only the exception class name and a sanitized message; full tracebacks and raw error details are written to debug logs only (`exc_info=True`). The redaction extends downstream: the step re-raises a generic error, so `workflow_failed` and group failure events (`parallel_agent_failed`, `for_each_item_failed`) also carry only the sanitized message — never raw manager or SDK exception text, which can embed argument or result values.
```
jrob5756: **RECOMMENDED**: Omitting argument/result bodies from events isn't the end-to-end secrets guarantee this section implies it is. For-each `key_by` values get copied into `item_key` on MCP lifecycle events, so a `key_by: token` config exposes that token even if it's also a tool argument. Explicitly returning MCP results through workflow output also publishes them in `workflow_completed`. Both are intentional, and both contradict the unconditional wording here.

Scope the guarantee to automatic exclusion of argument/result bodies specifically. Add an explicit warning that item identifiers and authored downstream/final outputs stay visible, and advise against sensitive `key_by` values.

src/conductor/engine/workflow.py
hertznsk: Right — `close()` absorbs `CancelledError` by design (task-affine teardown), so awaiting it directly during eviction swallowed the workflow's cancellation and the code went on to connect and invoke.

Fixed in ec64c8a: the eviction close now runs as its own shielded task (`_close_evicted_mcp_step_manager`), tolerating repeated cancellation while teardown drains, then re-raises `CancelledError` before any new connect or invoke. Covered by `test_eviction_close_reraises_cancellation` in `tests/test_engine/test_mcp_step_pool.py`, which verifies the close completes, the cancellation propagates, and no new connect happens.

src/conductor/executor/mcp_step.py
hertznsk: Correct — `WorkflowContext` duck-types group outputs by exactly those two keys, so a flattened `outputs`/`errors` pair misclassified the step's output in every context mode.

Fixed in ec64c8a: `outputs` and `errors` are now reserved from flattening alongside `content`/`structured`/`is_error` (see `_RESERVED_ENVELOPE_KEYS`); they stay reachable under `output.structured.outputs` / `output.structured.errors`, and the restriction is documented in the merge-rule sections of `docs/workflow-syntax.md` and the bundled skill references. Tests: `test_outputs_errors_keys_are_reserved_from_flattening` plus a context-level regression (`test_envelope_with_structured_outputs_keys_keeps_output_wrapper`) proving `{{ step.output.is_error }}` keeps working.

src/conductor/web/server.py
hertznsk: Agreed — live `parallel_completed`/`for_each_completed` carry counts only, while the synthetic path published the full saved outputs, MCP envelopes included.

Fixed in ec64c8a: group synthesis now knows member step types. MCP members of a parallel group are stripped from the aggregate and replayed as the metadata-only sequence the live engine emits (`mcp_started`/`mcp_completed` with `group_name`, plus the LLM-less `parallel_agent_completed`); MCP for-each items get `for_each_item_started` → `mcp_started` → `mcp_completed` → `for_each_item_completed` (no `output`), with the envelopes stripped from the aggregate. Mixed groups keep non-MCP members' outputs untouched. Both live and replay events now read truncation markers through the same typed helper. Tests in `TestSyntheticReplayMcpGroups` (`tests/test_web/test_server.py`) assert no result value appears anywhere in the emitted events.

src/conductor/engine/workflow.py
hertznsk: Right — `model_dump(mode="json")` preserves server extension fields, so a forged `spill_path` was forwarded as trusted metadata (and could be a non-string, breaking the frontend contract).

Fixed in ec64c8a at ingestion: `call_tool_structured` strips `truncated`/`spill_path` from every server content block, so only Conductor's own truncation pass can set them. The read-side is unified behind `mcp_truncation_metadata()` (shared by `_run_mcp_step` and the synthetic replay), which additionally type-checks `spill_path` as a string — so even envelopes persisted before this change replay safely. Tests: `test_server_supplied_truncation_fields_are_stripped` (manager) and `test_forged_spill_path_in_stored_envelope_is_dropped` (replay).

src/conductor/engine/workflow.py
hertznsk: Accurate — `--log-file` wires a Rich console, not a logging handler, and nothing turns DEBUG on, so the details were discarded into the void.

Fixed in ec64c8a with an explicit private sink: the engine writes the full exception (including the connect-time cause chain) to a per-run `*.mcp-diagnostics.log` next to the run's `*.events.jsonl`, on both run and resume (both wire the event log path into the engine). The redacted message now points at the concrete file (`MCP step 'x' failed; full diagnostic: <path>`). Safe categories keep authored detail: a per-call timeout surfaces as `McpStepTimeoutError` with its duration, and the name-only runtime checks propagate verbatim. The workflow-syntax secrets section documents the file's sensitivity. Tests assert both directions: canary never in events/raised errors, and present in the diagnostic file (`test_failure_event_is_redacted`, `test_schema_mismatch_fails_workflow_with_single_redacted_event`).

src/conductor/engine/workflow.py
hertznsk: Right — that ERROR record carried raw exception text plus a traceback, and Python's lastResort handler puts it on stderr even with nothing configured.

Fixed in ec64c8a: `connect_server` gained `redact_errors=True` for the deterministic step path — the failure log carries only the server name, while the raised `RuntimeError` still chains the original exception so the engine's diagnostic sink captures the full detail. Provider-facing behavior (the default) is byte-identical, covered by `test_unredacted_connection_error_keeps_full_log`; the redacted path is pinned by `test_redacted_connection_error_logs_safe_metadata_only`.

src/conductor/engine/workflow.py
hertznsk: All three waits (slot acquisition, connect, call) now race `interrupt_event` — see `_invoke_mcp_interruptible` in ec64c8a. A Stop cancels the in-flight call and drains it (re-cancellation-proof via the shared group-drain helper), then raises a control-flow signal (`_McpStepInterrupted`) instead of failing the step. The main loop routes it into the existing pause flow: `agent_paused` → Resume/Kill on the dashboard, or the CLI interrupt menu otherwise. The cancelled call is never auto-replayed inside that execution — Resume re-enters the step from the loop top, matching the documented at-least-once semantics. Group members mirror LLM members (they never receive the interrupt signal mid-call either; Stop reaches them through the group's cancellation/drain). Covered on both surfaces: `test_stop_during_call_pauses_and_resume_reexecutes` and `test_stop_during_call_kill_unwinds`.

src/conductor/engine/workflow.py
hertznsk: Fixed in ec64c8a as suggested: admission now reserves a slot for the in-flight connect atomically under the pool guard (`_mcp_step_pending`), so two concurrent first-time connects can no longer both pass a size-only check — the second evicts and the pool finishes at the cap. The reservation is released on failure or cancellation and folds into the pool on success; connect I/O stays outside the guard. The all-locked overflow is now documented as a soft threshold (in `_MCP_STEP_POOL_MAX`, `_pick_mcp_eviction_target`, and `docs/workflow-syntax.md`). Covered by `test_concurrent_first_connects_reserve_capacity`, which forces two overlapping connects at cap−1 and asserts exactly one eviction and a cap-sized pool.

src/conductor/engine/workflow.py
hertznsk: Aligned: the runtime check now uses wildcard membership (`"*" not in server_def.tools`), exactly matching the validator, so `["*", "health"]` behaves identically at both boundaries. Covered with the shared matrix you asked for: singleton wildcard, mixed wildcard, explicit include, and empty list — `TestToolAllowlist` in `tests/test_config/test_mcp_step_validation.py` (static) plus `test_mixed_wildcard_tools_list_allows_any_tool` in `tests/test_engine/test_mcp_step_workflow.py` (runtime).

src/conductor/config/validator.py
hertznsk: Fixed in ec64c8a: `_validate_mcp_steps` now explicitly parses every collected argument template with `_JINJA_ENV.parse` and reports the step plus the nested argument path (`arguments.opts.labels[1]`) on `TemplateSyntaxError` — covering templates nested in both mappings and lists. Reference analysis keeps its deliberate leniency for semantics; syntax is now checked before it. Tests: `TestArgumentTemplateSyntax` in `tests/test_config/test_mcp_step_validation.py`.

src/conductor/web/frontend/src/stores/workflow-store.ts
hertznsk: Fixed in 79d09fd: `mcp_completed` only calls `incrCompleted()` for standalone steps (`group_name == null`); parallel members update their node and leave the increment to `parallel_completed`, and the for-each item branch was already correct. `mcp_failed` never incremented, so no change was needed there. Covered by `does not double-count completion for parallel MCP members` in `workflow-store.test.ts`.

src/conductor/web/frontend/src/components/graph/graph-layout.ts
hertznsk: Fixed in 79d09fd on both halves: `workflow_started` seeds parallel members' nodes from the declared types (a name→type map off `data.agents`) in both the root and sub-workflow initialization branches, and `graph-layout.ts` now maps each parallel child's declared type through the shared `flowNodeTypeFor` helper (extracted so the standalone loop uses the same mapping) and sets `data.type` accordingly. A parallel MCP member now renders as `McpNode` and opens `McpDetail`. Covered by `assigns the declared node type to parallel group members in workflow_started` and a new `graph-layout.test.ts` case.

src/conductor/web/frontend/src/stores/workflow-store.ts
hertznsk: Fixed in 79d09fd: `ForEachItemRow` now treats `mcp_server != null` as having details (so a metadata-only completed item expands), renders Server/Tool/Result Bytes (with the truncation marker)/Spill Path in the item's metadata grid, and shows a visible `Tool reported an error (is_error)` warning when `mcp_is_error` is true while keeping the item's `completed` status. No raw result body is rendered — none exists on the item by design.

tests/test_engine/test_mcp_step_groups.py
hertznsk: The old coordination was indeed impossible: one shared slot meant item b's call could only start after item a's call finished, so the wait hit its 5s deadline and the canary `RuntimeError` never fired — the redaction assertions were vacuous.

Rewritten in ec64c8a with roles following call order instead of item identity: the first item to acquire the slot fails immediately with the canary `RuntimeError`; the sibling is cancelled wherever the fail-fast drain finds it (blocked at the slot boundary or inside its call — both are asserted via the `order` log). The test now asserts the intended `RuntimeError` actually happened (`mcp_failed` reports `error_type == "RuntimeError"`), the canary appears in no event, and sibling teardown lands before the pool close (`["call-raised", "sibling-blocked", "sibling-cancelled", "pool-close"]`).

plugins/conductor/skills/conductor/references/authoring.md
hertznsk: Correct — the description didn't match `_coerce_auto`, which YAML-parses the *whole rendered string*. Rewritten in ec64c8a (authoring.md, the executor docstrings, and `docs/workflow-syntax.md` now agree): `"[1, 2]"` becomes a list, `"1{{ x }}"` with `x=2` renders `"12"` and becomes the integer `12`, `"label: {{ x }}"` becomes a mapping, and only renders that parse as plain strings stay strings. Both docs now advise keeping rendered text unambiguous when the exact type matters.

docs/workflow-syntax.md
hertznsk: Fixed in ec64c8a: the Validation rules section now enumerates what `conductor run` actually repeats (server declared, stdio transport, tool allowlisted, plus the live-only tool-existence check) and states plainly that template syntax checking and the same-parallel-group reference analysis are validate-only diagnostics — with the caveat that skipping validation forfeits them.

docs/workflow-syntax.md
hertznsk: Fixed in ec64c8a: the guarantee is now scoped to automatic exclusion of argument/result bodies from event payloads. The section explicitly warns that `key_by` values land on item-scoped `mcp_*` events as `item_key` (advising against sensitive identifiers) and that explicitly surfacing a result — via workflow `output:` or a later step's `prompt`/`arguments` — publishes it in `workflow_completed` and beyond by design. The new diagnostic file's sensitivity is documented alongside.

src/conductor/executor/mcp_step.py
```diff
@@ -0,0 +1,288 @@
...
+        locally truncated; ``spill_path`` is the first local spill path (a
+        non-empty string) or ``None``.
+    """
+    blocks = content if isinstance(content, list) else []
+    truncated = any(isinstance(b, dict) and b.get("truncated") is True for b in blocks)
+    spill_path = next(
+        (
+            b["spill_path"]
+            for b in blocks
+            if isinstance(b, dict) and isinstance(b.get("spill_path"), str) and b["spill_path"]
```
jrob5756, line 125: **BLOCKING**: Server-supplied truncation and spill markers are still trusted on replay.

`manager.py:458-460` now strips `spill_path` and `truncated` from newly ingested results, so the live path is closed. But this file still accepts any nonempty `spill_path` string and `truncated=true` straight off saved checkpoint content (lines 115-125), and `server.py:1085-1104` passes that saved content through unchanged and publishes the markers. Since the original implementation persisted these fields into checkpoints, an existing checkpoint can still replay a server-supplied string as trusted metadata without any tampering.

Needs: suppress untrusted legacy markers on replay, or distinguish persisted Conductor-generated metadata from server extensions before publishing it.

src/conductor/engine/workflow.py
```diff
@@ -5122,6 +5811,83 @@ async def _execute_loop(self, current_agent_name: str) -> dict[str, Any]:
...
+                                # unwinds via InterruptError.
+                                pause_outcome = await self._handle_web_pause(agent.name, None)
+                                if pause_outcome.handled:
+                                    # Pause resolved (Resume / guidance —
+                                    # guidance was applied to context inside
+                                    # _handle_web_pause — / disconnect).
+                                    if self._interrupt_event is not None:
+                                        self._interrupt_event.clear()
+                                    continue
+                                if self._web_dashboard is not None:
```
jrob5756, line 5898: **RECOMMENDED**: Stop handling still misses groups and can auto-replay a call.

Standalone invocations now race against `interrupt_event` (lines 1950-1968, 2129-2165), so Stop reaches those. But the parallel and for-each callers at 7039-7043 and 7574-7578 still don't wire interruption into their group waits, so Stop is silent there. And this line re-enters an interrupted standalone step automatically when no browser is connected — the shared pause handler does the same on disconnect — so an interrupted call can replay without anyone actually deciding to resume it. Applies to both run and resume.

Needs: connect group execution to the Stop/pause flow, and require an explicit resume decision before repeating an interrupted MCP call, including the no-client and disconnect cases.

src/conductor/web/frontend/src/stores/workflow-store.ts
```diff
@@ -1528,13 +1547,14 @@ const eventHandlers: Record<string, (state: MutableState, data: Record<string, u
...
         const agentNames = new Set<string>();
+        const agentTypes = new Map<string, NodeType>(ctx.agents.map((a) => [a.name, (a.type || 'agent') as NodeType]));
 
         for (const pg of ctx.parallelGroups) {
           for (const a of pg.agents) groupAgents.add(a);
           agentNames.add(pg.name);
           ensureNode(ctx.nodes, pg.name, 'parallel_group');
           ctx.groupProgress[pg.name] = { total: pg.agents.length, completed: 0, failed: 0 };
-          for (const agentName of pg.agents) ensureNode(ctx.nodes, agentName, 'agent');
+          for (const agentName of pg.agents) ensureNode(ctx.nodes, agentName, agentTypes.get(agentName) || 'agent');
```
jrob5756, line 1558: **RECOMMENDED**: Static nested MCP members still carry the generic detail type.

Root-level graph mapping and fresh initialization now use the declared type. But `buildStaticChildContext` (line 631) still seeds parallel members as `'agent'`, normal subworkflow startup reuses those placeholders (1430-1443), and `ensureNode` (532-539) doesn't update an existing node's type. So nested members still end up typed `'agent'` here, and `DetailPanel.tsx:59-71` routes them to `AgentDetail` instead of `McpDetail`.

Needs: preserve declared types when building static child contexts, and update reused placeholders during initialization.

jrob5756 (commented):
14 points from the last pass landed: eviction now preserves cancellation, structured results no longer masquerade as group outputs, redacted failures point somewhere real, pool admission counts pending connections, wildcard allowlisting agrees at both boundaries, and the docs now match whole-render parsing and the runtime/offline split, among others.

Three points are still open: one blocking, two recommended.

- **Blocking**: checkpoint replay still trusts server-supplied truncation/spill markers (`src/conductor/executor/mcp_step.py:121`)

Details on all three are inline.

src/conductor/executor/mcp_step.py
hertznsk: Fixed in b300d35. `_synth_mcp_pair` no longer republishes markers from a stored envelope: synthetic `mcp_completed` events now always report `truncated: false` / `spill_path: null`, and `mcp_truncation_metadata` is documented as the live-path-only reader — its trust comes from the ingestion stripping, which a persisted checkpoint may predate. The stored envelope itself stays intact in the workflow context for routing and templates; only the published synthetic event loses the badge. The test that pinned the old behavior was updated, and a new test asserts suppression on the standalone, parallel, and for-each synthetic paths (all three converge on `_synth_mcp_pair`).

src/conductor/engine/workflow.py
hertznsk: The auto-replay half is fixed in 78c55ec. `WebPauseOutcome` now carries an explicit `reason` (`resume` / `guidance` / `disconnect` / `unavailable`), and an interrupted mcp call is re-entered only on an explicit resume decision: a dashboard Resume/guidance, or the CLI interrupt menu (every menu outcome is a decision). On a mid-pause disconnect or a dashboard with zero connected clients the run parks instead of replaying: a new `_McpStepOutcomeUncertain` (an `InterruptError` subclass, so `workflow_failed` is flagged `stopped_by_user` and a failure checkpoint is saved) makes `conductor resume` the explicit at-least-once boundary. No `mcp_failed` is emitted — the tool reported no failure, the call's outcome is simply unknown — and no `agent_resumed` on the parking path; an explicit Resume completed in the same wait batch wins over a simultaneous disconnect. The LLM auto-resume behavior and its event stream are unchanged, since re-running an LLM agent only costs tokens.

The group half I deliberately did not change here. No step type — LLM agents included — receives a mid-call interrupt inside parallel/for-each today: Stop takes effect at the next between-step check, and Kill works via engine-task cancellation. Wiring Stop into group waits is a cross-cutting change (cancel one member or the whole group, draining siblings, `fail_fast` interaction, N concurrent pause UIs) that deserves its own issue rather than an mcp-only divergence. The cancellation paragraph in `docs/workflow-syntax.md` now states the exact semantics, including the no-implicit-replay rule.

src/conductor/web/frontend/src/stores/workflow-store.ts
hertznsk: Fixed in 5a245b7. `buildStaticChildContext` now seeds parallel members with their declared types (the same declared-type map the root and child `workflow_started` handlers already use), and the child `workflow_started` re-syncs every declared agent's node type from the runtime topology onto reused placeholder nodes — `ensureNode` intentionally still never updates a type, so untyped event paths can't downgrade a typed node, and the sync walks only `ctx.agents`, leaving group nodes untouched. Vitest coverage added for both halves: declared types in static previews, and a deliberately stale placeholder being corrected when the real `workflow_started` lands on it.

jrob5756 (approved):

PR #522, opened 2026-09-10
feat(claude-agent-sdk): add deterministic auth modes and readiness

ldavidgomez:
## Summary

- Adds deterministic `auth_mode` support for `claude-agent-sdk`:
  - `subscription` selects the child-process subscription path by clearing API-key/token variables only through `ClaudeAgentOptions.env`.
  - `api_key` requires an inherited API key and clears the competing auth token.
  - `auto` preserves the CLI's inherited credential resolution.
- Adds bounded, sanitized auth-readiness checks and structured `conductor doctor --check` diagnostics.
- Hardens subprocess cleanup, including cancellation during the spawn window.
- Makes ordinary tests hermetic: no local Claude CLI, login, or credentials required.
- Documents the supported contract and limitations.

## Validation

- Independent review accepted the implementation.
- `ruff check .` passes after rebase.
- Focused provider, diagnostics, doctor, plugin, and factory tests: 436 passed; two Windows symlink-privilege failures are pre-existing and unrelated.
- A hermetic full-suite comparison against the rebase base found no ticket-attributable regression; upstream CI remains the final gate.

## Notes

- Builds on the ProviderRegistry concurrency fix already released in v0.1.37.
- This change does not claim billing attribution or rely on a real model invocation as evidence.


ldavidgomez:
### Example configurations

Use `subscription` when the workflow should explicitly use the Claude subscription login available to the child process:

```yaml
provider:
  type: claude-agent-sdk
  auth_mode: subscription
```

Use `api_key` when the runner supplies `ANTHROPIC_API_KEY` and API-key authentication is required:

```yaml
provider:
  type: claude-agent-sdk
  auth_mode: api_key
```

Use `auto` only when preserving the Claude CLI's inherited credential resolution is intentional:

```yaml
provider:
  type: claude-agent-sdk
  auth_mode: auto
```

`subscription` and `api_key` make the intended authentication path explicit for the provider subprocess. `auto` preserves existing CLI precedence and is therefore intentionally less deterministic.

jrob5756 (commented):
The deterministic-auth mechanism still has the gaps described inline: settings-file `env` entries and inherited credential/backend selectors can override an explicit auth mode, the subscription readiness probe can check a credential that execution removes, and the broad conftest exemption lets ordinary execution tests reach the real Claude CLI.

CI update: I've rebased this branch onto main after the shared telemetry-test fix in #532. Your four commits retain their original patches; I haven't changed the implementation or applied formatting fixes. Fresh CI is at https://github.com/microsoft/conductor/actions/runs/35010353127.

Blocking findings:
- `src/conductor/providers/claude_agent_sdk.py:1623`: settings sources can reload a credential after the override blanks it.
- `src/conductor/providers/claude_agent_sdk.py:1217`: OAuth token and cloud-provider env vars aren't handled, so they can override the explicit mode.
- `src/conductor/providers/claude_agent_sdk.py:1083`: the readiness probe and execution can check different credentials.
- `tests/conftest.py:212`: the module-wide exemption makes ordinary tests depend on local Claude auth state.
- Ruff formatting rejects `tests/test_cli/test_doctor.py`. This is a branch-specific issue, not the shared telemetry failure. Please format that file and commit the result. The workflow-level failure and Lint failure in my original summary were the same blocker, counted twice; most downstream jobs cannot run until lint passes.

The inline recommendations remain open too. I have left all contributor-specific changes to you; this rebase only brings in the main-branch fix.

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -1162,23 +1610,25 @@ async def _execute_session(
...
+            # authentication path (TICKET-20260816-0002). Built by
+            # ``_auth_env_override`` from ``self._auth_mode`` alone — never
+            # from ``auth_status`` — so ``auto`` is guaranteed to contribute
+            # no override at all. The SDK copies ``os.environ`` and spreads
+            # this mapping last when launching the ``claude`` subprocess
+            # (``claude_agent_sdk`` 0.2.87,
+            # ``_internal/transport/subprocess_cli.py``), so parent
+            # ``os.environ`` is never mutated and every other agent/process
+            # keeps its own inherited credentials untouched.
+            env=auth_env,
```
jrob5756, line 1623: **BLOCKING:** The environment override runs before Claude loads its enabled settings sources, and a settings-file `env` entry beats an inherited environment value. So `auth_mode: api_key` with `setting_sources: [user]` can have `ANTHROPIC_AUTH_TOKEN` reloaded from user settings after Conductor clears it, and that token wins over the API key you asked for. Subscription mode has the same problem in reverse — a settings-file API key can reload after Conductor clears it.

This breaks the feature's core promise: pick a mode and Conductor will actually use it. Enforce the selection through the effective CLI settings configuration too, not just the subprocess environment — a securely written per-call settings override where that's supported, and a hard rejection where a higher-priority conflict can't be overridden. Add a test that pairs each explicit mode with a conflicting settings-file credential.

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -810,6 +976,254 @@ def skills_require_plugin_root(self) -> bool:
...
+                for name in ("ANTHROPIC_API_KEY", "ANTHROPIC_AUTH_TOKEN")
+                if _nonblank_env(name) is not None
+            ]
+            if overridden:
+                logger.warning(
+                    "auth_mode=subscription: overriding inherited %s for the Claude "
+                    "child process so the logged-in subscription session is used.",
+                    " and ".join(overridden),
+                )
+            return {"ANTHROPIC_API_KEY": "", "ANTHROPIC_AUTH_TOKEN": ""}
```
jrob5756, line 1217: **BLOCKING:** Blanking `ANTHROPIC_API_KEY` and `ANTHROPIC_AUTH_TOKEN` isn't enough to guarantee subscription mode uses the local login. `CLAUDE_CODE_OAUTH_TOKEN` is still inherited and outranks it. Inherited `CLAUDE_CODE_USE_BEDROCK`, `CLAUDE_CODE_USE_VERTEX`, or `CLAUDE_CODE_USE_FOUNDRY` will also select a cloud backend ahead of either explicit mode. The requested mode can end up running against a different account or backend entirely, silently.

Handle these selectors explicitly: clear the competing OAuth token when subscription mode is selected, and reject incompatible cloud-provider selections with a clear error unless you're actually implementing deterministic support for them. Leave `auto` alone, and document whatever combinations you're choosing not to support.

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -810,6 +976,254 @@ def skills_require_plugin_root(self) -> bool:
...
+                requested_mode=self._auth_mode,
+                inferred_mode="subscription",
+                ready=False,
+                error=(
+                    "Claude CLI not found. Install with: npm install -g @anthropic-ai/claude-code"
+                ),
+            )
+
+        try:
+            stdout_bytes, stderr_bytes, returncode = await _run_auth_status_subprocess(cli_path)
```
jrob5756, line 1083: **BLOCKING:** The auth-status subprocess inherits the parent environment as-is, but execution later applies different credential overrides. That means with an ambient API key and no subscription login, the probe can report `loggedIn=true` off the API key, this method treats that as subscription readiness, and then execution strips the key and fails to authenticate. The probe also doesn't get the execution child's working directory or effective settings sources, so it can end up checking a completely different auth context than the one that will actually run.

Build the child's auth configuration once, before preflight, and pass that same environment, working directory, and settings sources into both the probe and execution — don't mutate `os.environ` as a side channel between the two. Add a regression test where API-key-backed status must not count as subscription readiness.

tests/conftest.py
```diff
@@ -199,3 +199,71 @@ def _isolated() -> Path:
...
+
+# Test modules that genuinely exercise `ClaudeAgentSdkProvider._check_auth_readiness`
+# itself (calling the real method against a mocked subprocess, or asserting on its
+# exact CLI-probe/timeout/interrupt behavior) and must therefore receive the *real*
+# implementation rather than the hermetic stub below. Matched by module basename
+# (without `.py`) so it is independent of which directory collects the file.
+_CLAUDE_AUTH_READINESS_HERMETIC_EXEMPT_MODULES = frozenset(
+    {
+        "test_claude_agent_sdk_auth",
+        "test_claude_agent_sdk",
```
jrob5756, line 212: **BLOCKING:** Exempting the entire `test_claude_agent_sdk` module from the readiness stub takes it away from tests that never needed real auth in the first place. `TestExecute.test_execute_text_only`, for example, mocks `query` and `ClaudeAgentOptions` but not the new preflight — with no `ANTHROPIC_API_KEY` set, it now launches the actual installed Claude CLI and depends on whatever's logged into the developer's keychain. On a machine without that login, it fails before it ever reaches the behavior it's testing, which is exactly the hermetic-test guarantee this PR claims to have.

Apply the readiness stub to ordinary provider tests by default, and let the handful of readiness/CLI-discovery tests that need the real thing opt out explicitly via a marker or fixture — with those exceptions controlling their own credential env vars and mocking subprocess creation. A fail-fast subprocess sentinel would be good insurance here: assert that ordinary execution tests never spawn an auth process at all.

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -810,6 +976,254 @@ def skills_require_plugin_root(self) -> bool:
...
+                inferred_mode="subscription",
+                ready=False,
+                error=(
+                    f"Claude CLI could not be started ({type(exc).__name__}). "
+                    "Check installation and permissions."
+                ),
+            )
+
+        stderr_text = stderr_bytes.decode("utf-8", errors="replace").lower()
+        if returncode != 0:
```
jrob5756, line 1109: **RECOMMENDED:** `claude auth status` exits 1 when the user isn't logged in, and this branch returns before the JSON body is even parsed. So an ordinary logged-out response — `loggedIn=false` — gets the generic "CLI may not be accessible" message instead of the actual remedy (`Run: claude auth login`). The existing logged-out test doesn't catch this because it returns exit code 0 with `loggedIn=false`, which isn't the real shape of that response.

Parse the bounded, sanitized status output before falling back to the generic nonzero-exit error, and treat an explicit `loggedIn=false` as the actionable logged-out case — while still making sure a nonzero exit can never be read as readiness success. Update the regression test to use exit code 1 to match reality.

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -1402,77 +1852,19 @@ async def _execute_session(
...
-            *(() if is_windows else (Path("/usr/local/bin/claude"),)),
-            Path.home() / ".local/bin/claude",
-            Path.home() / "node_modules/.bin/claude",
-            Path.home() / ".yarn/bin/claude",
-            Path.home() / ".claude/local/claude",
-        )
-        for path in fallbacks:
-            if path.exists() and path.is_file():
-                return True
+        status = await self._check_auth_readiness()
```
jrob5756, line 1860: **RECOMMENDED:** `validate_connection` now defers entirely to `_check_auth_readiness`, which returns success immediately for `auto` or `api_key` as soon as `ANTHROPIC_API_KEY` is non-blank — neither path ever checks whether the Claude executable exists. An install with the Python SDK importable but no discoverable CLI now passes factory validation and shows as connected in `doctor`, even though execution can't start. The old validation caught exactly this.

Keep a non-spawning CLI-availability check in `validate_connection` for every auth mode. The API-key paths can still skip the auth-status subprocess call — they just shouldn't skip checking that the executable required for execution actually exists.

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -810,6 +976,254 @@ def skills_require_plugin_root(self) -> bool:
...
+            return ClaudeAuthStatus(
+                requested_mode=self._auth_mode,
+                inferred_mode="subscription",
+                ready=False,
+                error="Authentication check returned an unreadable status response.",
+            )
+
+        filtered = {key: payload.get(key) for key in _AUTH_STATUS_WHITELIST if key in payload}
+        logged_in = filtered.get("loggedIn")
+        auth_method = filtered.get("authMethod") or filtered.get("apiProvider")
```
jrob5756, line 1150: **RECOMMENDED:** When `authMethod` is missing, this fallback stores `apiProvider` in `auth_method`, and `auth_status_diagnostic` publishes that under `sdk_observed.authMethod`. A CLI response with only `loggedIn=true` and `apiProvider='firstParty'` ends up reporting an observed auth method of `'firstParty'` — a value the CLI never actually returned for that field. That breaks the diagnostic's as-observed contract and conflates a provider identifier with an authentication mechanism.

Populate `auth_method` only from the CLI's actual `authMethod` field. If `apiProvider` is worth surfacing, keep it under its own name instead of relabeling it. Add a test asserting that a missing `authMethod` stays absent from `sdk_observed`, rather than getting backfilled.

docs/workflow-syntax.md
```diff
@@ -592,6 +592,76 @@ agents:
...
+succeeded — plus a third note when `requested_mode` is `auto` stating that
+the effective credential path follows the SDK/CLI's inherited-environment
+precedence. This is what distinguishes a subscription session from an
+API-key-present session, since `authMethod` alone reports identically for
+both:
+
+```text
+$ conductor doctor --check
+claude-agent-sdk   installed   ✓ connected
+                                Conductor: requested_mode=subscription, inferred_mode=subscription
```
jrob5756, line 652: **RECOMMENDED:** This example shows `conductor doctor --check` reporting `requested_mode=subscription`, but `diagnostics.gather_provider` constructs the provider without `provider_settings` and never reads a workflow's `auth_mode` — doctor always shows `auto`. On top of that, when an API key is present the `auto` readiness branch skips the CLI status call entirely, so `sdk_observed` comes back empty. As written, this tells readers that doctor verifies a workflow's explicit auth choice, which it doesn't.

Change the example to `requested_mode=auto` with a note explaining that doctor checks the default provider configuration, not a specific workflow's `auth_mode`, and explain when the SDK-observed fields will be empty.

docs/configuration.md
```diff
@@ -255,6 +256,26 @@ name rather than accepted and dropped. An enabled tier brings that tier's
...
+spawns — it is rejected on every other provider name, the same way
+`setting_sources` is. `subscription` selects the child-process subscription
+path by passing empty `ANTHROPIC_API_KEY` and `ANTHROPIC_AUTH_TOKEN` through
+`ClaudeAgentOptions.env`; `api_key` requires a non-empty inherited
+`ANTHROPIC_API_KEY` and clears the competing `ANTHROPIC_AUTH_TOKEN` the same
+way; `auto` preserves inherited credential resolution and is intentionally
+non-deterministic (no override is contributed at all). **No global
+`os.environ` mutation occurs** in any mode — the override is scoped to the
+per-call `ClaudeAgentOptions.env` mapping. This selects a credential path, not
+an endpoint — it composes with, and is orthogonal to, the routing fields
```
jrob5756, line 270: **RECOMMENDED:** This says `auth_mode` composes with the routing fields described above it, but `ProviderSettings` actually rejects those routing fields for `claude-agent-sdk` and rejects `auth_mode` on the providers that do support them. Someone follows this guidance, sets `auth_mode` alongside `base_url` or `api_key`, and gets a schema error instead of the setup described here.

State plainly that `auth_mode` doesn't enable custom routing and doesn't change field compatibility on either provider. If this line is actually trying to describe endpoint configuration the Claude CLI itself inherits, name that mechanism directly and separate it from Conductor's (unsupported) YAML routing fields.

src/conductor/providers/claude_agent_sdk.py
ldavidgomez: Addressed in `d4758a4` / `c2b2eeb`. The SDK exposes no per-call setting verified to take precedence over a settings file's `env` block, so explicit modes now reject a non-empty `setting_sources` outright. The check runs at `conductor validate` and again at run time, before any probe or session, with one shared error message. The run-time check uses the provider's configured tiers, so an agent's `skills: []` can't bypass it. Tests cover both explicit modes with a configured tier, with and without the opt-out, and assert that no probe is spawned and no session is queried.

src/conductor/providers/claude_agent_sdk.py
ldavidgomez: Addressed in `d4758a4` and `caa7d18`.
- Both explicit modes blank `ANTHROPIC_AUTH_TOKEN` and `CLAUDE_CODE_OAUTH_TOKEN` in the child environment, and `subscription` also blanks `ANTHROPIC_API_KEY`.
- An inherited non-blank `CLAUDE_CODE_USE_BEDROCK`, `CLAUDE_CODE_USE_VERTEX` or `CLAUDE_CODE_USE_FOUNDRY` is now **rejected** in both explicit modes. Readiness fails before the probe runs or any session starts. The error names the variable, never its value, and suggests unsetting it or using `auth_mode: auto` to keep the inherited backend.
- The selectors are still sent as empty strings, which covers a value set in the parent after capture.
- `auto` is unchanged and passes them through.
- Tests cover each explicit mode with each selector (no spawn, no query, name without value), several selectors at once, blank values, and `auto` preservation.
- The auth-mode table in `workflow-syntax.md` documents the rejection.

src/conductor/providers/claude_agent_sdk.py
ldavidgomez: Addressed in `d4758a4` / `c2b2eeb`. One immutable `EffectiveAuthContext` is captured per operation, before preflight. It holds the environment, working directory, effective settings sources and CLI path, and both the probe and `ClaudeAgentOptions` read from it; `os.environ` is never modified. In subscription mode the probe runs with `ANTHROPIC_API_KEY` blanked, so an API-key-backed status can't count as subscription readiness. The probe also runs as `claude --setting-sources=<csv> auth status --json`, with the same tiers the session gets, so an ambient settings-file credential can't satisfy it. Tests assert that the probe and the options receive identical env, cwd, CLI and sources, even when the parent environment changes in between. A regression test puts the only credential in an ambient tier and shows that readiness now fails.

tests/conftest.py
ldavidgomez: Addressed in `d4758a4`. The readiness stub is now autouse for every test. Only tests marked `@pytest.mark.claude_auth_readiness_mocked` run the real readiness method, and for those, process creation is replaced with a guard that fails the test if it's ever reached. No test can launch a real Claude CLI.

src/conductor/providers/claude_agent_sdk.py
ldavidgomez: Addressed in `d4758a4`. Stdout is now parsed before the exit code, so `{"loggedIn": false}` with exit 1 reports "Not logged in to Claude Code. Run: claude auth login". A nonzero exit can never be read as success. The regression test now uses exit code 1, alongside a test for non-JSON output with a nonzero exit.

src/conductor/providers/claude_agent_sdk.py
ldavidgomez: Addressed in `c2b2eeb`. Every mode now requires a discoverable CLI, checked without spawning anything, and that includes `auto` with `ANTHROPIC_API_KEY` set. The API-key paths still skip `claude auth status`. `validate_connection()` and `conductor doctor` report not-connected when the SDK is installed but the CLI is missing. Covered by tests for `auto` + key with and without a CLI, and by a Doctor test that builds the real provider.

src/conductor/providers/claude_agent_sdk.py
ldavidgomez: Addressed in `d4758a4`. `auth_method` is now filled only from the CLI's own `authMethod`. `apiProvider` is surfaced under its own name, and Doctor labels it as the backend. A test checks that a response with only `loggedIn` and `apiProvider` reports no `authMethod` and shows `apiProvider` separately in `sdk_observed`.

docs/workflow-syntax.md
ldavidgomez: Addressed in `d4758a4` and `caa7d18`.
- The example now shows `requested_mode=auto`.
- Doctor prints a `Scope:` line stating that it checks the default provider configuration and doesn't inspect a workflow's `auth_mode`, and the docs say the same.
- The docs now state explicitly that with `auth_mode: auto` and `ANTHROPIC_API_KEY` set, the `SDK:` fields are empty, because that path doesn't run `claude auth status`. Since Doctor always checks `auto`, that's what it shows whenever the key is set.

docs/configuration.md
ldavidgomez: Addressed in `d4758a4` and `caa7d18`. The "composes with" wording is gone. `configuration.md` now says plainly that `auth_mode` selects a credential, not an endpoint, doesn't enable custom routing, and doesn't change field compatibility:
- Conductor's YAML routing fields stay rejected for `claude-agent-sdk` whatever `auth_mode` is.
- `auth_mode` stays rejected on `copilot` and `claude`.

It also separates this from the CLI's own environment. An inherited `ANTHROPIC_BASE_URL` reaches the `claude` child process in every mode, because no `auth_mode` sets or clears it; that's the CLI's behaviour, not a Conductor routing field.

ldavidgomez:
Thanks for the thorough review. All nine threads are addressed in `d4758a4`, `c2b2eeb` and `caa7d18`, with replies inline.

- **One effective auth context** is captured per operation: environment, working directory, effective settings sources and CLI path. The readiness probe and the SDK session both read only from it; `os.environ` is never modified.
- **Explicit modes are deterministic:**
  - they blank competing API-key, auth-token and OAuth credentials in the child environment;
  - they reject inherited cloud-backend selectors, naming the variable but never its value;
  - they refuse `setting_sources`, including for an agent with `skills: []`.
- **The probe loads the session's settings tiers** via `--setting-sources`, including the empty selection and `skills: []`.
- **Every mode requires the Claude CLI**, checked without spawning anything, including `auto` with an API key.
- **Status handling:** a logged-out response is reported even on a nonzero exit, and `apiProvider` is reported separately from `authMethod`.
- **Doctor** states that it checks the default configuration only, and the docs explain when its `SDK:` fields are empty.
- **The docs** state that `auth_mode` doesn't affect routing or field compatibility.
- **Tests are hermetic:** real readiness only runs on opt-in, and process creation is guarded.

**Validation:**
- The focused provider, auth and Doctor suites pass (524).
- With fake API key, auth token, OAuth token and all three cloud selectors set in the parent environment, the auth suites pass (314) and the Doctor/diagnostics suites pass (120).
- `ruff check` and `ruff format --check` are clean, and `ty` reports 0 errors.

Could you take another look when you have a chance?

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -810,6 +1101,315 @@ def skills_require_plugin_root(self) -> bool:
...
+                error="Authentication check returned an unreadable status response.",
+            )
+
+        filtered = {key: payload.get(key) for key in _AUTH_STATUS_WHITELIST if key in payload}
+        logged_in = filtered.get("loggedIn")
+        auth_method = filtered.get("authMethod")
+        api_provider = filtered.get("apiProvider")
+        subscription_type = filtered.get("subscriptionType")
+        api_key_source = filtered.get("apiKeySource")
+        if logged_in is True:
```
jrob5756, line 1415: RECOMMENDED (partially addressed): The `loggedIn=false` case now gets the right login guidance even on a nonzero exit, and the exit code is consulted when JSON parsing fails. But the `loggedIn=true` branch here still returns `ready=True` regardless of `returncode` — a parsed `loggedIn=true` response with exit code 1 is accepted as ready.

Keep the specific logged-out message, but reject any nonzero exit before returning readiness success, including when the JSON says `loggedIn=true`. Add a test case for `loggedIn=true` with exit code 1 to lock this in.

jrob5756 (commented):
Since the last review, the doctor formatting issue is fixed, explicit auth modes now reject conflicting settings sources and inherited cloud backends, competing OAuth credentials are neutralized, readiness and execution share one auth context, ordinary provider tests no longer hit real auth readiness, API-key readiness now requires the CLI, backend and auth method stay distinct, the doctor docs match its actual scope, and auth mode is no longer described as YAML routing. The earlier rebase/CI note is moot.

CI is still failing on the reviewed commit, which blocks merge on its own:

- CI run 35454230118, attempt 2 (completed/failure): https://github.com/microsoft/conductor/actions/runs/35454230118
- CI / Test (Python 3.13, windows-latest): failure — https://github.com/microsoft/conductor/actions/runs/35454230118/job/106331932656
- Changelog run 35454230132, attempt 2 (completed/failure): https://github.com/microsoft/conductor/actions/runs/35454230132
- Changelog / Changelog: failure — https://github.com/microsoft/conductor/actions/runs/35454230132/job/106331861454

For all four, the fix is the same: resolve whatever's failing on this commit, then rerun and confirm checks pass before merging.

One open item left inline on `claude_agent_sdk.py`, marked recommended rather than blocking.

ldavidgomez:
The latest follow-up is pushed in `b13228c`. Both workflow runs are waiting for approval before any jobs can start:

- CI: https://github.com/microsoft/conductor/actions/runs/35608685755
- Changelog: https://github.com/microsoft/conductor/actions/runs/35608685840

Could someone with write access approve these runs when convenient? The follow-up addresses the Windows home-resolution failure and migrates the direct `CHANGELOG.md` edit to the required changelog fragment.

src/conductor/providers/claude_agent_sdk.py
ldavidgomez: Addressed in `39d6ef0`. A `loggedIn: true` response is now accepted only when the CLI exits successfully. A contradictory nonzero exit returns a sanitized readiness failure, while `loggedIn: false` retains the existing login guidance. Regression coverage includes `loggedIn: true` with exit code 1. The Windows Python 3.13 job and the complete CI suite are green on the rebased head.

ldavidgomez:
Thanks for the follow-up review. The branch is now rebased onto the current `main`, and CI is fully green on `39d6ef0`: all 14 checks passed, including Windows Python 3.13, Changelog, Build Package, Linux tests, lint and typecheck.

CI run: https://github.com/microsoft/conductor/actions/runs/35632645001

GitHub reports the PR as mergeable with `mergeStateStatus: CLEAN`. The final independent review verdict is Accepted.

jrob5756 (approved):

PR #524, opened 2026-09-10
fix(providers): clarify structured output completion

hertznsk:
## Summary

- give Pydantic AI's `final_result` output tool an explicit completion contract
- keep `ToolOutput`, schema sanitization, output retries, and provider parity unchanged
- cover both the shared builder metadata and the serialized OpenAI Chat Completions request

Pydantic AI already exposes the intended customization seam through `ToolOutput(..., description=...)`, but its default description only says that the tool is the final response. Some models behind OpenAI- or Anthropic-compatible endpoints still finish with plain text unless the requirement to call the tool is explicit. This change puts that requirement on the tool definition rather than modifying workflow-authored system prompts.

This is an alternative to the backend-wide `PromptedOutput` switch proposed in #508. It preserves forced tool output, nested-schema sanitization, parse recovery, telemetry, and continuation message formats.

## Validation

- `make check`
- `uv run pytest tests/test_providers/test_pydantic_ai_agent_builder.py tests/test_providers/test_pydantic_ai_structured_output.py tests/test_providers/test_openai_http_stub.py -q` (84 passed)
- full non-performance suite: 8919 passed, 49 skipped; two unrelated POSIX permission tests failed because this environment did not produce their expected `chmod(0o000)` behavior
- direct builder smoke test confirmed the same `final_result` description for both `backend="openai"` and `backend="anthropic"`

## Notes

The description improves adherence for models that support tool calling. It cannot compensate for a gateway that strips tool descriptions or does not implement tool calls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   [`docs/workflow-syntax.md`](docs/workflow-syntax.md#mcp-steps) and
   [`examples/mcp-step.yaml`](examples/mcp-step.yaml).
 
+### Fixed
+
+- **Pydantic AI structured-output agents explicitly require `final_result`** —
+  the generated output tool now tells models that they must call it before
+  finishing and that plain-text responses are not accepted. This improves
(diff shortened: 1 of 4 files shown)


jrob5756 (approved):

PR #525, opened 2026-09-11
fix(engine): inherit background mode in subworkflows

arturleao:
## Summary

- Propagate background-mode context from parent workflows into child workflow engines.
- Prevent nested `human_gate` steps from racing a CLI stdin prompt under `--web-bg`.
- Add regression coverage for a dashboard-backed nested gate with TTY-like stdin.

## Why

Only the CLI builds `RunContext`. Child engines previously saw `bg_mode=False`, so a nested gate could call `Prompt.ask`, raise `EOFError`, and fail a background workflow even while a web dashboard was available.

## Testing

- `tests/test_engine/test_subworkflow.py` (42 passed)
- Scoped Ruff and formatting checks


jrob5756 (approved):

PR #526, opened 2026-09-13
refactor(config): split step definitions into a discriminated union

### LINKED ISSUE #517: refactor(schema): replace monolithic AgentDef with Pydantic discriminated unions for step kinds
## Problem

`AgentDef` in `src/conductor/config/schema.py` is currently a single monolithic `BaseModel` (~1,500 lines) representing every kind of workflow step: LLM agents, shell scripts, human gates, question forms, MCP direct calls, sub-workflows, terminations, setters, and wait steps.

Because all step-specific fields (`command`, `args`, `options`, `questions`, `server`, `tool`, `workflow`, `status`, `duration`, `value`, etc.) are declared on this single model with `extra="forbid"`, Pydantic cannot automatically reject fields that belong to a different step kind.

To enforce field exclusivity, `validate_agent_type` (`schema.py:1990–2623`) contains over **600 lines** of repetitive manual assertions:
```python
if self.type == "script":
    if self.prompt: raise ValueError("script agents cannot have 'prompt'")
    if self.provider: raise ValueError("script agents cannot have 'provider'")
    if self.model: raise ValueError("script agents cannot have 'model'")
    if self.tools is not None: raise ValueError("script agents cannot have 'tools'")
    # ... ~30 more if-statements per step kind
```

Whenever a new step kind is introduced (e.g. `wait`, `questions`, `mcp`, `terminate`), authors must manually add dozens of cross-rejection checks across all other branches.

## Proposed Architecture: Discriminated Unions

In Pydantic v2, this pattern is idiomatically solved using **Discriminated Unions** keyed on `type` (with `discriminator="type"` and `extra="forbid"` on each model).

### Proposed Model Structure

```python
class StepBase(BaseModel):
    """Base fields shared by all workflow steps."""
    model_config = ConfigDict(extra="forbid")

    name: str
    description: str | None = None
    input: list[str] = Field(default_factory=list)
    routes: list[RouteDef] = Field(default_factory=list)


class LLMAgentDef(StepBase):
    type: Literal["agent"] | None = "agent"
    prompt: str
    system_prompt: str | None = None
    model: str | None = None
    provider: ProviderName | None = None
    tools: list[str] | None = None
    reasoning: ReasoningConfig | None = None
    dialog: DialogConfig | None = None
    validator: ValidatorConfig | None = None
    sandbox: SandboxConfig | None = None
    skills: list[str] | None = None
    plugins: list[PluginDef] | None = None
    output: dict[str, OutputField] | None = None
    retry: RetryPolicy | None = None
    timeout_seconds: int | None = None
    context_tier: ContextTier | str | None = None
    session_key: str | None = None
    max_session_seconds: int | None = None
    max_agent_iterations: int | None = None


class ScriptStepDef(StepBase):
    type: Literal["script"]
    command: str
    args: list[str] = Field(default_factory=list)
    env: dict[str, str] = Field(default_factory=dict)
    working_dir: str | None = None
    timeout: int | None = None
    stdin: str | None = None
    output: dict[str, OutputField] | None = None


class HumanGateStepDef(StepBase):
    type: Literal["human_gate"]
    prompt: str
    options: list[GateOption]


class MCPStepDef(StepBase):
    type: Literal["mcp"]
    server: str
    tool: str
    arguments: dict[str, Any] = Field(default_factory=dict)
    output: dict[str, OutputField] | None = None
    timeout: int | None = None


class QuestionsStepDef(StepBase):
    type: Literal["questions"]
    questions: list[QuestionDef] | None = None
    source: str | None = None
    allow_back: bool | None = None
    allow_skip: bool | None = None
    allow_skip_all: bool | None = None
    allow_abort: bool | None = None
    abort_route: str | None = None


class TerminateStepDef(StepBase):
    type: Literal["terminate"]
    status: Literal["success", "failed"]
    reason: str
    output_template: str | None = None


class SetStepDef(StepBase):
    type: Literal["set"]
    value: Any | None = None
    values: dict[str, Any] | None = None
    output_type: Literal["raw", "json"] = "raw"


class WaitStepDef(StepBase):
    type: Literal["wait"]
    duration: str | int | float | None = None
    wait_until: str | None = None
    poll_interval: str | int | float | None = None
    reason: str | None = None


class WorkflowStepDef(StepBase):
    type: Literal["workflow"]
    workflow: str
    input_mapping: dict[str, str] = Field(default_factory=dict)
    max_depth: int | None = None


# Discriminated union type
AgentDef = Annotated[
    LLMAgentDef
    | ScriptStepDef
    | HumanGateStepDef
    | MCPStepDef
    | QuestionsStepDef
    | TerminateStepDef
    | SetStepDef
    | WaitStepDef
    | WorkflowStepDef,
    Field(discriminator="type"),
]
```

## Benefits

1. **Automatic Field Exclusion & Validation:** `model_config = ConfigDict(extra="forbid")` on each model automatically forbids foreign fields per step type, eliminating over 600 lines of error-prone manual `if getattr(self, ...) is not None` assertions in `validate_agent_type`.
2. **Schema & IDE Ergonomics:** Static type checkers (`mypy`, `pyright`, `ty`) and IDEs can narrow step types accurately (e.g. `step.command` only exists on `ScriptStepDef`).
3. **Cleaner Extensibility:** Adding a new step kind only requires defining its model and adding it to the Union, without touching existing step definitions or cross-matrix guards.
4. **Significant LOC Reduction:** Reduces `src/conductor/config/schema.py` by 600–800 lines of boilerplate validation logic.

## Compatibility & Migration Considerations

- `WorkflowConfig.agents` can use `list[AgentDef]` transparently since Pydantic parses YAML dicts into the concrete discriminated model based on `type` (defaulting to `type: agent` when `type` is omitted or `None`).
- Existing call sites inspecting `isinstance(step, ...)` or using duck-typing across steps will either continue working or benefit from more precise pattern matching / `match step:`.

hertznsk:
Closes #517.

## What

Replaces the monolithic `AgentDef` — one model carrying every step type's fields, with a hand-maintained validator matrix deciding which fields each `type:` accepts — with concrete Pydantic models united by a static `StepDef` discriminated union:

- **`AgentDef`** is now the provider-backed LLM step only.
- New public models: **`HumanGateStepDef`, `QuestionsStepDef`, `ScriptStepDef`, `MCPStepDef`, `WaitStepDef`, `SetStepDef`, `TerminateStepDef`** (deliberately no `routes`), **`WorkflowStepDef`**, plus `StepBase` / `RoutableStepBase`. All exported from `conductor.config` and `conductor.config.schema`.
- `WorkflowConfig.agents: list[StepDef]` and `ForEachDef.agent: StepDef` parse into concrete variants through `Field(discriminator="type")` with an outer `BeforeValidator` that canonicalizes the legacy shorthand (missing or explicit-null `type`) to `"agent"` at the boundary.
- Every variant owns exactly the fields meaningful for its kind with `extra="forbid"`, so a misplaced field is rejected next to its step instead of being silently ignored — this is what makes new step types cheap to add (no validator matrix to extend).
- The published JSON Schema now exposes `oneOf` + `discriminator` with per-variant `additionalProperties: false`, so editors and external tooling get per-type completion and validation.

## Compatibility

- **Workflow YAML is unchanged.** Omitted `type` and `type: null` still load as LLM agents (canonicalized to `type: "agent"` at parse time).
- **`AgentDef(...)` keeps working for LLM agents**, including `AgentDef(type=None, ...)`, `model_dump(exclude_none=True)` round-trips, `model_copy(update=...)`, and the synthetic agents built by `OutputValidator` and the ACA runner.
- Intentional tightenings: constructing a non-LLM step via `AgentDef(type="script", ...)` no longer works (use the named step class); a field belonging to a different step type now fails with Pydantic's standard `extra_forbidden` rather than the custom `"<type> agents cannot have '<field>'"` messages (variant-owned invariant messages are unchanged); `HumanGateStepDef` lost the inert `model` field; `human_gate` is rejected as a `for_each` inline agent (concurrent iterations would compete for one interactive channel — route to a gate from the group's `routes:` instead).

## Engine and consumers

- Main / parallel / for-each dispatch narrows via `isinstance`, with defensive `ExecutionError` guards for variants the static validator would reject — a directly-constructed engine never runs it. The for-each guard fires before the empty-source early return so a forbidden inline variant can never pass silently.
- Generic consumers narrowed the same way: validator routing graph and subworkflow/template scans, `conductor show`/`run`/`plugin`, the MCP `conductor_plan_tree` (a terminate step now renders with empty routes), executor skills/plugins helpers, and the synthetic provider agents (`max_depth` removed).
- `FileString` (`!file`) preservation restored on `human_gate`/`questions` prompts so file-backed gate prompts keep resolving relative `{% include %}` paths.

## Testing

- Full suite: **9113 passed, 3 failed** — all three failures reproduce identically on `main` in this environment (two `chmod(0o000)` permission tests and one timing-sensitive performance test); none touch this change.
- `ruff check` clean; `ty check src` shows only the same pre-existing diagnostics as `main`; `make validate-examples` green.
- New regression coverage: JSON Schema `oneOf`/discriminator/`additionalProperties` shape, `ForEachDef.agent` variant parsing, legacy missing/null-`type` canonicalization, `model_dump` round-trips for all nine variants, resume round-trip from a legacy no-type YAML, runtime defensive guards for parallel/for-each groups (including ahead of the empty-source early return), `conductor_plan_tree` with a terminate step, and an explicit-empty `output_template: {}` coverage-override regression.


jrob5756 (commented):
I love this! Just a couple of things that came up. The generated JSON Schema rejects agent configs the loader still accepts, and there are validation and test-matrix gaps described below and inline.

CI update: I've rebased this branch onto main after the shared telemetry-test fix in #532. Your three commits retain their original patches; I haven't formatted or changed the implementation. Fresh CI is at https://github.com/microsoft/conductor/actions/runs/35010355726.

Blocking:
- `src/conductor/config/schema.py:1412`: the generated schema rejects supported untagged agents (see inline comment).
- Ruff formatting fails in seven PR-modified files: `src/conductor/cli/run.py`, `src/conductor/config/validator.py`, `src/conductor/mcp/serve/introspect.py`, `tests/test_config/test_questions_validation.py`, `tests/test_config/test_step_def_union.py`, `tests/test_integration/test_examples.py`, and `tests/test_skills/test_schema.py`. This is specific to the branch, not the shared telemetry issue. Please run the formatter and commit the changes. The workflow-level failure and Lint failure in my original summary were the same blocker, counted twice. Most downstream jobs cannot run until lint passes.

The inline recommendation about revalidating copied step instances is unchanged.

Recommended: migrate the terminate-field rejection matrix to the actual step variants (`tests/test_config/test_schema.py`).

`TestTerminateAgent.test_terminate_fields_rejected_on_other_step_types` still runs script, workflow, and human_gate payloads through `AgentDef.model_validate`. Now that `AgentDef` is only the LLM variant, all nine cases pass, but for the wrong reason. `AgentDef` is rejecting the discriminator and the foreign fields, not the concrete step type rejecting `status`/`reason`/`output_template` the way the test claims to check. If `ScriptStepDef` or `WorkflowStepDef` started accepting terminate-only fields tomorrow, this suite would stay green.

Suggested fix: validate these tagged payloads through `TypeAdapter(StepDef)`, or select the matching concrete class directly. Assert the baseline payload is valid first, then inject the forbidden field and check for an `extra_forbidden` error at that field's location.

src/conductor/config/schema.py
```diff
@@ -1171,1622 +1068,384 @@ def _validate_skill_entries(entries: list[str]) -> list[str]:
...
+    AgentDef
+    | HumanGateStepDef
+    | QuestionsStepDef
+    | ScriptStepDef
+    | MCPStepDef
+    | WaitStepDef
+    | SetStepDef
+    | TerminateStepDef
+    | WorkflowStepDef,
+    Field(discriminator="type"),
```
jrob5756, line 1547: **BLOCKING**: This oneOf doesn't preserve the loader's missing/null-type compatibility. Every variant defaults its `type`, so none of the generated schemas actually require that field. An agent like `{"name": "x", "prompt": "Do the task"}` matches both `AgentDef` and `QuestionsStepDef`, and `oneOf` rejects anything matching more than one branch. Explicit `type: null` fares no better — it fails every branch's literal check. `WorkflowConfig.model_validate` accepts both forms fine, so the break is one level up: anyone validating against `WorkflowConfig.model_json_schema()` (editors, external linters, schema-aware tooling) now rejects workflows Conductor runs without complaint. Inline `for_each` agents have the same gap.

Suggested fix: customize the input schema so the non-LLM branches require their explicit discriminator, while the LLM branch keeps accepting `type` omitted, null, or `"agent"`. Leave the runtime types and concrete-constructor defaults alone. Add schema tests that validate full workflows and inline agents across all three accepted LLM type forms.

src/conductor/config/schema.py
```diff
@@ -1171,1622 +1068,384 @@ def _validate_skill_entries(entries: list[str]) -> list[str]:
...
+    output: dict[str, OutputField] | None = None
+    command: str
+    args: list[str] = Field(default_factory=list)
+    env: dict[str, str] = Field(default_factory=dict)
+    working_dir: str | None = None
+    stdin: str | None = None
+    timeout: int | None = Field(None, gt=0)
 
-    Supports Jinja2 templating against accumulated context.
+    @field_validator("command")
```
jrob5756, line 1304: **RECOMMENDED**: Moving these semantic checks out of the after-model validator and into field/before validators quietly drops them for instances that already exist. Pydantic skips replacement validators on revalidation by default, so `ScriptStepDef(name="x", command="echo").model_copy(update={"command": ""})` now sails through `WorkflowConfig.model_validate` — the identical construction used to fail. A copied `WaitStepDef` with `duration="25h"` gets the same pass. Anything that builds workflows programmatically by copying and mutating step instances, rather than parsing dicts, can now push invalid execution config past a boundary that used to catch it.

Suggested fix: keep these invariants as after-model validators, or turn on instance revalidation for the affected variants. Add regression cases that pass invalid copied instances through `WorkflowConfig` and `ForEachDef` — `model_copy` itself doesn't need to become validating.

src/conductor/config/schema.py
hertznsk: Fixed in 59cf0f0. The input schema is now customized per variant, as suggested: every non-LLM branch requires its explicit discriminator (`_require_step_type_in_schema` adds `type` to `required`), while the LLM branch keeps accepting `type` omitted, `null`, or `"agent"` (`_agent_step_type_in_schema` widens the property to `anyOf: [{const: agent}, {type: null}]`, still optional with its default). Runtime types and concrete-constructor defaults are untouched.

New `TestPublishedSchemaAcceptsLoaderForms` (tests/test_config/test_step_def_union.py) validates full workflows and inline `for_each` agents against `WorkflowConfig.model_json_schema()` via the `jsonschema` package across all three accepted LLM type forms, plus: tagged non-LLM variants still validate; an untagged mapping with foreign fields is rejected by both the schema and the loader (they stay in agreement); and a shape assertion that only the `agent` branch leaves `type` optional.

src/conductor/config/schema.py
hertznsk: Fixed in 71cc6d3, with a follow-up in 96f7e90. Went with after-model validators on the affected variants: ScriptStepDef (command), MCPStepDef (server/tool), WaitStepDef (duration bounds via a shared `_check_wait_duration` helper its field validator also uses), WorkflowStepDef (workflow path), and TerminateStepDef (status + reason — written defensively, since a copied instance never passed through coercion: `status=None`/`"bogus"` and `reason=None` are now ValidationErrors instead of being accepted or crashing with AttributeError).

The field/before validators stay for error location on mapping input, and `model_copy` itself remains non-validating. New `TestInstanceRevalidation` passes invalid copied instances through both `WorkflowConfig` and `ForEachDef` (9 parametrized cases), plus a valid-copy control.

hertznsk:
@jrob5756 Thanks for the thorough review — all four items are addressed in the pushed commits:

- **CI / Lint** — c23320b: the seven files the pinned ruff formatter flagged are reformatted; `make check` (ruff check + `ruff format --check` + ty) and the full test suite (9100+ tests) are green locally.
- **Generated JSON Schema rejects supported untagged agents** — 59cf0f0, details in the thread reply above.
- **Instance-revalidation gap** — 71cc6d3 + 96f7e90, details in the thread reply above.
- **Terminate-field rejection matrix** — eb7b1f5: the nine cases now validate through `TypeAdapter(StepDef)`, assert the baseline payload is valid first, then inject the forbidden field and assert an `extra_forbidden` error at the exact `(step_type, field)` location.

The branch was also rebased onto current main (picks up the #532 telemetry flake fix) before pushing.

jrob5756 (approved):

PR #529, opened 2026-09-14
fix(config): add !rawfile tag to bypass YAML-sniffing on prompt includes

### LINKED ISSUE #528: !file include parses YAML-shaped Markdown prompts as mappings, breaking prompt/system_prompt validation
## Summary

The `!file` tag in `src/conductor/config/loader.py` (`construct_file_tag`) does not just read the referenced file — it attempts to parse the content as YAML and returns the parsed structure whenever it happens to be a mapping or a list:

```python
parsed = sub_yaml.load(content)
if isinstance(parsed, (dict, list)):
    return parsed
# Scalar YAML or None → return raw string content
return FileString(content, source_path=file_path)
```

For structured includes (`tools:`, `output:`) that is the intended behavior, but for `prompt:` / `system_prompt:` — which usually reference Markdown — it is a footgun: a Markdown prompt whose line ends with a colon followed by `- ` bullets **is** a valid YAML mapping, so `system_prompt: str | None` rejects it.

The resulting error blames the schema value, not the include heuristic, and is very hard to trace back to the actual cause:

```
1 validation error for WorkflowConfig
for_each.1.agent.agent.system_prompt
  Input should be a valid string [type=string_type, input_value={'Сожми changelog д...обновления.']}, input_type=dict]
```

(The doubled `agent.agent` in the path is just the discriminated-union tag from #517 — cosmetic. The heuristic itself predates the v0.1.x releases, so this reproduces on any recent version.)

## Why it bites in both directions

The failure mode is **silent in both directions**, and an ordinary content edit can flip a file across the boundary without any config change:

- A Markdown file that is *invalid* YAML (e.g. contains free-floating prose paragraphs) is returned as raw text and works fine.
- Remove or reword that prose — as a normal prompt edit would — and the file can become *valid* YAML, at which point validation starts failing with the error above.

Observed in the wild exactly like that: a prompt file worked for weeks because a trailing paragraph made it invalid YAML; deleting that paragraph (an unrelated prompt cleanup) turned it into a valid 3-key mapping and broke `conductor validate` on a freshly installed build, presenting as an apparent version regression.

## Reproduction

`prompts/system.md.jinja`:

```text
Summarize the changelog, keeping only what affects the triage of a dependency update:
- breaking changes and migration notes;
- security fixes.

Answer in markdown:
- a short header with the package name and version range;
- one or two closing sentences on upgrade risk.
```

`workflow.yaml`:

```yaml
workflow:
  name: repro
  entry_point: worker

agents:
  - name: worker
    system_prompt: !file prompts/system.md.jinja
    prompt: "Summarize {{ workflow.input.text }}"
    routes:
      - to: $end
```

`conductor validate workflow.yaml` fails with:

```
agents.0.agent.system_prompt
  Input should be a valid string [type=string_type, input_value={'Summarize the changelog, ...risk.']}, input_type=dict]
```

The Markdown parses as a two-key YAML mapping (`Summarize the changelog, ... update:` → list, `Answer in markdown` → list).

## Suggested directions

Any of these would defuse the trap; they are not mutually exclusive:

1. **A raw-text include tag** (e.g. `!rawfile`) that guarantees the file content is returned verbatim, so prompt files have an include form whose type cannot silently change under them.
2. **A targeted error message**: when a string-typed field (`prompt`, `system_prompt`, `command`, `reason`, …) receives a dict/list produced by a `!file` include, fail with a message naming the included file and explaining the YAML heuristic — instead of a bare pydantic `string_type` that blames the user's value.
3. **Documentation**: call the heuristic out in the `!file` section of `docs/workflow-syntax.md`, including the "a prose edit can flip YAML validity" trap.

## Workaround

Make the prompt file invalid as YAML — e.g. ensure it starts with a plain prose line without a colon. That forces the `except YAMLError` path and the raw text is returned. It works, but it is exactly as fragile as the failure it fixes: any later edit that restores YAML validity re-breaks the workflow.

AmirF194:
Fixes the YAML-sniffing trap you described: added a `!rawfile` tag (your suggested direction 1) that always returns a file's content verbatim, so a prompt file's type can't flip depending on whether its prose happens to parse as YAML.

`construct_file_tag`'s file-reading and cycle-detection logic is now shared through a new `_read_included_file` helper, used by both `!file` and `!rawfile`. `!rawfile` skips the YAML sub-parse entirely, so it never recurses and doesn't need the nested-include bookkeeping `!file` has.

Documented it in `workflow-syntax.md`, including the "an unrelated prose edit flips it" trap, and updated the Jinja-includes note since a `!rawfile`-loaded prompt gets the same include search root as `!file` (it keys off `isinstance(x, FileString)`, not which tag produced it).

Didn't touch direction 2 (a targeted error message on the existing `!file` heuristic). `!rawfile` is the actual escape hatch, so it seemed like the more useful half to ship first, and it doesn't preclude adding the error message later.

Added tests for both sides: your exact repro (a line ending in `:` followed by a bulleted list) still trips `!file` as before, and the same file loaded through `!rawfile` stays a string. `pytest tests/test_config/` passes on the branch (1371 passed, 3 skipped, none related to this). `ruff check`/`format` and `ty check` are clean on the changed file.

Fixes #528


jrob5756 (commented):
Thanks for tackling this. Before I review the implementation, could we adjust the approach to make the two include behaviors explicit?

- **`!file`** — always read the file as text, preserving the existing environment substitution and file-backed Jinja behavior.
- **`!yamlfile`** — explicitly parse the file as YAML, supporting nested includes and reporting malformed YAML rather than falling back to text.

For example:

```yaml
prompt: !file prompts/review.md
system_prompt: !file prompts/system.md
output: !yamlfile schemas/review.yaml
```

This addresses the underlying ambiguity rather than adding a special tag to avoid it. Reading a prompt as text should be the unsurprising behavior; interpreting a file as structured configuration should be explicit.

I recognize this breaks existing structured `!file` includes, and I am comfortable taking that change here with a clear migration path. Please update the relevant documentation, repository examples, and bundled authoring skill references to explain the new semantics, including nested includes, path resolution, and the fact that text includes still undergo environment substitution and prompt Jinja rendering. Please also add an entry to `CHANGELOG.md` explicitly identifying this as a **breaking change**, with before/after examples showing how existing structured `!file` references migrate to `!yamlfile`. Where practical, validation errors should point users toward `!yamlfile` when an old structured include now supplies text.

Assuming `!rawfile` has not shipped, we can skip introducing that name entirely.

Could you revise the PR in that direction? I would like to settle the YAML contract first, then review the implementation against it.

AmirF194:
Went with the split you described: `!file` always returns text now, `!yamlfile` explicitly parses YAML and raises a `ConfigurationError` naming the file if it's malformed instead of falling back to a string. Dropped `!rawfile` entirely.

Updated `docs/workflow-syntax.md`, both bundled skill references (`authoring.md`, `yaml-schema.md`), and added a `### Changed` entry to `CHANGELOG.md` with before/after examples, flagged as a breaking change. The test suite's structured-content cases (output schemas, tool lists, nested includes) now go through `!yamlfile`; added a fixture with genuinely malformed YAML to cover the new error path, plus a test confirming `!file` on the yaml-shaped markdown file no longer traps.

`pytest tests/test_config/` passes (1372 passed, 4 skipped), `ruff check`/`format` and `ty check` are clean.

src/conductor/config/loader.py
```diff
@@ -154,26 +157,54 @@ def construct_file_tag(self, node: Any) -> Any:
...
+                        suggestion="Use !file instead of !yamlfile if this file is meant "
+                        "to be loaded as plain text.",
+                    ) from e
                 if isinstance(parsed, (dict, list)):
                     return parsed
-                # Scalar YAML or None → return raw string content
-                return FileString(content, source_path=file_path)
-            except YAMLError:
-                # Not valid YAML → return as raw string
+                # A scalar YAML document (plain string/number/bool/None): return
```
jrob5756: **BLOCKING**: `!yamlfile` still falls back to raw text for YAML scalars.

The parse step at line 190 runs correctly, but the result is only kept when it comes back as a dict or list. A scalar document — `true`, `42`, a quoted string — returns the original text instead of the parsed value, so `!yamlfile` only honors its "always parse as YAML" contract for mappings and sequences. Quoted strings keep their quotes, and `true` stays the string `"true"` rather than becoming a boolean.

Return the parsed value for scalar documents as well, and leave verbatim text loading to `!file`.

jrob5756 (commented):
Since the last pass, the bulk of this landed cleanly: `!file` returns text consistently, structured includes support nesting, malformed YAML now fails with a real error instead of silently misbehaving, and the docs/skill references/migration notes all reflect the split between the two tags. `!rawfile` was also dropped rather than added on top of `!yamlfile`, which keeps the surface area smaller than the original design floated.

One thing from the original review is still open, and it's blocking: `!yamlfile` doesn't consistently parse as YAML yet. It works for mappings and sequences, but scalar documents (a bare `true`, a quoted string, a bare number) come back as raw text instead of the parsed value — so the tag's contract only holds for part of the input space.

**Blocking:**
- `src/conductor/config/loader.py:199` — `!yamlfile` scalar documents aren't parsed

AmirF194:
Pushed at 07a8a40: construct_yamlfile_tag now returns the parsed value unconditionally instead of checking isinstance(parsed, (dict, list)) first, so a scalar document is parsed like everything else. Added a regression test on a bool and a number fixture; both come back as raw text on main and as the parsed value on this branch.

jrob5756 (approved):

AmirF194:
Glad the scalar fix closed it out. The two-pass review on the yaml contract was worth it, the final split reads a lot cleaner than the original `!rawfile` idea.

PR #531, opened 2026-09-15
fix(providers): inherit run usage limits in compaction summarizer

hertznsk:
## Problem

On the `claude` / `openai` providers, the summarizing tier of context compaction **always degrades to the sliding-window fallback on long agent runs**. The chain:

1. `pydantic-ai-harness` 0.24.0's `SummarizingCompaction._summarize` calls its nested summarization agent as `agent.run(prompt, usage=ctx.usage)` — sharing the parent run's usage, but passing **no `usage_limits`**.
2. Pydantic AI normalizes the missing argument to `UsageLimits()`, whose default `request_limit` is **50** (`pydantic_ai/agent/__init__.py`: `usage_limits = usage_limits or UsageLimits()`).
3. Conductor correctly resolves a per-agent `max_agent_iterations` (e.g. 200) into the main run's limits — but the nested summarizer never sees them. Once the parent run has made ≥50 requests, the summarizer's `check_before_request` raises `UsageLimitExceeded: ... request_limit of 50`.
4. `_TierWrapper` catches it, names `summarizing` in `degraded_tiers`, and the chain falls through to the deterministic sliding-window tier. Usage is shared, limits are not — so this fires on the **first** compaction of any run with `max_agent_iterations > 50` that reaches the summarizing tier, i.e. exactly when a long run needs real summarization.

Observed in production: an agent with `max_agent_iterations: 200` and a 250k context window degraded at 186k tokens (74%), the sliding window measured 125k tokens with the ~4-chars/token heuristic against a real server-side count of 261k (Cyrillic-dense content), and the next request failed with `HTTP 400: Input exceeds maximum input tokens` — classified non-retryable, run lost after 27 minutes.

## Fix

Inside `build_tiered_compaction`, a local `_RunLimitsInheritingSummarizingCompaction(SummarizingCompaction)` overrides `_summarize` so the nested run **inherits the limits the parent run is already enforcing**:

```python
usage_limits = (
    ctx.usage_limits
    if ctx.usage_limits is not None
    else UsageLimits(request_limit=None)
)
result = await agent.run(prompt, usage=ctx.usage, usage_limits=usage_limits)
```

- Inheriting (rather than uncapping) keeps the documented semantics: the summary call consumes one request slot from the agent's `max_agent_iterations` budget, and a genuinely exhausted budget still refuses the summarizer and degrades honestly to sliding-window. The `None` fallback covers only synthetic `RunContext`s with no run behind them (passing `usage_limits=None` would restore the default 50).
- The subclass lives inside `build_tiered_compaction` to preserve the module's deliberate lazy-harness-import boundary; the method body mirrors the upstream `_summarize` behavior and control flow and is commented with the harness version and the removal condition.
- Kept minimal per review: the follow-up for oversized newest responses (harness `ClampOversizedMessages`) is deliberately out of scope.

## Tests

New `TestSummarizerUsageLimits` in `tests/test_providers/test_pydantic_ai_compaction.py`:

- `test_summarizer_runs_after_fifty_parent_requests` — `RunUsage(requests=50)` + `UsageLimits(request_limit=200)`, the real `build_tiered_compaction` stack, a real nested `Agent.run` on `TestModel` (no mocks). Asserts the summarizing tier does **not** degrade, that the summary consumes exactly one shared request slot (`requests == 51`), and that history is compacted. **Confirmed to fail without the fix** with the exact production error (`UsageLimitExceeded: ... request_limit of 50`).
- `test_summarizer_refused_when_parent_budget_is_exhausted` — usage at the inherited limit refuses the summarizer and degrades to sliding-window; pins that the fix inherits the budget rather than granting an uncapped one.

## Verification

- `tests/test_providers/` — 1651 passed
- `ruff check` / `ruff format --check` — clean
- `ty` — no new diagnostics (6 pre-existing warnings on `main`, unchanged)

## Follow-ups (out of scope)

- Upstream the same fix to `pydantic-ai-harness` (`agent.run(..., usage_limits=ctx.usage_limits or UsageLimits(request_limit=None))` in `_summarize`); drop this subclass once the minimum pinned harness includes it.
- Evaluate harness `ClampOversizedMessages` for a single oversized newest response, which neither the summarizer's nor the sliding window's kept tail can reclaim.


src/conductor/providers/_pydantic_ai/compaction.py
```diff
@@ -610,6 +612,81 @@ def build_tiered_compaction(config: CompactionConfig) -> AbstractCapability[Any]
...
+                _INCREMENTAL_UPDATE_INSTRUCTION,
+                _format_messages,
+            )
+
+            formatted = _format_messages(
+                messages, skip_previous_summary=previous_summary is not None
+            )
+            prompt = self.summary_prompt.format(messages=formatted)
+
+            if previous_summary is not None:
```
jrob5756, line 654: **RECOMMENDED**: The local override now owns the incremental-summary prompt construction, but neither new regression test exercises a previous summary or inspects the summarization prompt. Both use a constant model response and only check request counts, tier failure state, and reduced history length — they'd still pass if the override stopped carrying forward what an earlier compaction had preserved. Incremental summarization is on by default, so this behavior needs coverage.

Add a two-cycle regression through `build_tiered_compaction` using a recording `FunctionModel`. Return a distinctive first summary, append enough history to trigger a second summarization, and assert the second prompt contains the previous summary exactly once plus the newly summarized turns. Pick a `target_tokens` the sliding-window tier can actually hit, so it doesn't discard the summary before you can assert it survived, and confirm each summarization still consumes one shared request.

jrob5756 (commented):
CI update: the Python 3.12 failure I flagged was caused by a flaky test already on main, not by this PR. `test_cross_task_detach_ownership` recorded token IDs without retaining the token objects, so Python could reuse an ID and make the ownership assertion fail. I reproduced it on unchanged main and fixed it in #532.

I've rebased this branch onto the corrected main without changing your patch. Fresh CI is at https://github.com/microsoft/conductor/actions/runs/35010354769. The two CI blockers in my original summary were duplicate reports of that baseline failure; please disregard the request to fix it in this PR.

The inline recommendation on `compaction.py:654` still stands: add coverage for repeated compactions carrying a previous summary into the next summarization prompt. I haven't changed that code or the tests in your patch.

src/conductor/providers/_pydantic_ai/compaction.py
hertznsk: Done in 167646d. Added `test_second_summarization_receives_previous_summary_once` to `TestSummarizerUsageLimits`, driving two full compaction cycles through `build_tiered_compaction` with a recording `FunctionModel`:

- Cycle one returns a distinctive summary (`ANCHOR-SUMMARY-ALPHA …`); the test asserts it survives the whole tier chain into the compacted history as the single summary `SystemPromptPart`, and that the first summarization prompt carries no `<previous-summary>` block.
- Cycle two appends ten large turns plus a fresh 20-message tail so the new turns land inside the summarized prefix; the second recorded prompt is asserted to contain the first summary exactly once — verbatim inside the `<previous-summary>` anchor block (`second_prompt.count(summary) == 1` plus an exact `<previous-summary>\n…\n</previous-summary>` match) — together with the newly summarized turns. The final history holds only the second (BRAVO) summary, so the coverage fails if the override stops carrying a previous summary forward.
- Each cycle asserts `usage.requests` advances by exactly one (10 → 11 → 12), confirming both summarizations consume one shared request slot.

On sizing: `target_tokens=150` is chosen so the summarizing tier alone brings the history under target (~58 estimated tokens), which stops escalation before the sliding-window tier runs — with `keep_messages=20` and a 22-message post-summary history, that tier would otherwise drop the summary at index 0 before the second cycle could read it. `trigger_tokens=100` keeps both cycles well above the outer gate.

Verified locally: `pytest tests/test_providers/test_pydantic_ai_compaction.py` → 43 passed; `make check` (ruff + `ty check src`) clean.

jrob5756 (approved):

PR #534, opened 2026-09-16
Preserve custom Copilot provider routing when resuming sessions

laymonage:
`conductor resume` reused prior Copilot sessions without reapplying `runtime.provider`, so resumed runs silently fell back to the default Copilot model routing instead of the configured BYOK/OpenAI-compatible endpoint. In workflows that depend on custom `base_url`, `api_key`, or `wire_api`, this breaks resume while fresh sessions continue to work.

- **Resume path parity**
  - apply the same `_apply_provider_config(...)` step on `resume_session(...)` that already exists on `create_session(...)`
  - keep resumed Copilot sessions pinned to the workflow's configured provider routing instead of only the SDK default

- **Behavioral impact**
  - resumed sessions now preserve custom routing fields such as `type`, `wire_api`, `base_url`, `api_key`, and related provider settings
  - this closes the mismatch where initial agent executions used the configured endpoint but `conductor resume` did not

- **Regression coverage**
  - add a focused test that reproduces the missing `provider` kwargs on `resume_session(...)`
  - assert the resumed-session call carries the same provider config as new-session creation

```python
resume_kwargs = {
    "on_permission_request": self._default_permission_handler,
    "working_directory": resolved_cwd,
    ...
}
self._apply_provider_config(resume_kwargs)
self._apply_github_token(resume_kwargs)
session = await self._client.resume_session(resume_sid, **resume_kwargs)
```


laymonage:
I encountered this issue myself while trying to resume a conductor workflow run with my BYOK setup. The provided fix works fine, so I asked Copilot to write a PR to fix the issue upstream.

jrob5756 (approved):

PR #535, opened 2026-09-17
chore(changelog): replace conflicting CHANGELOG.md edits with towncrier fragments

hertznsk:
## Summary

Every PR editing the shared `CHANGELOG.md` `[Unreleased]` section conflicts with every other such PR. This PR replaces those edits with **towncrier fragments** — one small file per change in `changelog.d/`, compiled into `CHANGELOG.md` only in release-prep PRs — and makes the compiled section the curated source of GitHub Release notes.

- **No new dependencies** — towncrier runs via `uvx`, pinned to a single literal `towncrier==25.8.0` shared by the Makefile and CI. `uv.lock` is untouched.
- **No product code changes** — process and CI only. Rollback = delete `changelog.d/` and revert.

## Changes (4 commits)

1. **`chore(changelog): adopt towncrier fragment workflow`** — `[tool.towncrier]` config in `pyproject.toml` (four types: added/fixed/changed/removed, `title_format` without compare links, `issue_format = "#{issue}"` so the default template renders one clickable `(#NNN)`); static `[Unreleased]` header with the towncrier insertion marker; migration of the three pending entries into fragments (`392.added.md`, `+otel-mcp-spans.added.md`, `+pydantic-final-result.fixed.md`); `changelog.d/README.md` fragment contract (agent-first: DO/DON'T, free-form naming, copy-paste examples); Makefile targets `changelog-draft` / `changelog-build VERSION=X.Y.Z`.
2. **`ci: enforce changelog fragment contract on pull requests`** — new fork-safe workflow `Changelog` (`pull_request` only, `contents: read`, no secrets, PR code never executed): feature mode (≥1 valid fragment, `CHANGELOG.md` edit rejected, `towncrier check` on any real fragment change incl. rename destinations) and release mode (detected by any `pyproject.toml` version-value change: exactly one new section, single exact marker, fragments consumed, extraction rehearsal with the **base-branch** script, `uv.lock` pins the version). Plus an sdist assertion that `CHANGELOG.md` ships in the tarball.
3. **`ci(release): publish curated changelog section as GitHub Release notes`** — `extract-release-notes.sh` (strict `## [X.Y.Z] - YYYY-MM-DD` header — intentionally refuses legacy compare-link headers so a tag without a release-prep PR fails loudly; single exact towncrier marker positioned above the section; non-empty content check that understands CommonMark headings); `release.yml` uses `--notes-file` (create **and** existing-release edit) and sets `--title "Conductor X.Y.Z"`; `--generate-notes` removed.
4. **`docs: document changelog fragment contract and release process`** — five normative rules in `AGENTS.md` (fragment per user-facing change; never edit `CHANGELOG.md` in a feature PR; free-form naming; CI failure messages are actionable; exemption only via maintainer-applied `changelog-not-required` label — which waives both the fragment requirement and the edit prohibition, added fragments still validated); README contributing step with `#changelog` anchor; release checklist rewritten for the towncrier flow incl. open-PR conversion runbook, advisory→required rollout, kill-switch, two new failure runbooks, and the `Conductor X.Y.Z` title check.

## Verification

- Six rounds of independent adversarial review (structural traceability of every plan requirement, bash robustness incl. `pipefail`/SIGPIPE and git-diff edge cases, evidence-driven QA re-running towncrier 25.8.0 and the extractor on isolated fixtures, cross-artifact coherence) — all findings fixed, final wave all-approve.
- Fixture-verified behaviors: fragment-rename smuggling an invalid name under the exemption label is caught by `towncrier check`; a docs-only edit to `changelog.d/README.md` does not trigger `towncrier check`; marker typo/missing/duplicated/misplaced all fail; a heading-only release section fails extraction.
- `uv.lock` unchanged; no new dependencies; `make check` (ruff + ty) passes.

## Rollout notes (maintainer steps, documented in the release checklist)

- The `changelog-not-required` label currently exists only in the fork; create it upstream before relying on exemptions.
- The workflow cannot run on this PR itself (GitHub does not execute workflows from a PR branch); after merge: convert open PRs per the runbook, then add the `Changelog` job (job name, not workflow name) to required status checks. The check is `pull_request`-based and not merge-queue-compatible — exclude it from merge-queue requirements if a merge queue is ever enabled.


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,192 @@
...
+  cancel-in-progress: true
+
+jobs:
+  changelog:
+    name: Changelog
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410b181273ba90b1 # v7.0.1
```
jrob5756: **BLOCKING**: This checkout SHA doesn't match any real commit — `76410b181` vs. the valid `76410c181` already used in ci.yml. GitHub can't resolve it, so the job fails during setup, before it ever gets to enforce the fragment or release contract.

```suggestion
        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
```

jrob5756 (commented):
This is a solid direction — moving off a single shared CHANGELOG.md to towncrier fragments is the right call — but the workflow that enforces it isn't working yet. One issue is a broken action pin that fails the job outright, and two more (the version comparison, the fragment-deletion exemption) will reject valid PRs once people start using this. None of it can merge as-is.

Blocking, by location:
- `.github/workflows/changelog.yml:46` — checkout action pinned to an unresolvable SHA
- `.github/workflows/changelog.yml:135` — lockfile version check rejects normalized prerelease versions
- `.github/workflows/changelog.yml:186` — `changelog-not-required` exemption still fails on fragment-only deletions
- No test coverage for the new release gate or the release-notes extractor (see below)
- Two failing CI checks on the reviewed commit (see below)

**b4 — BLOCKING: no test coverage for the new gate and extractor.** This PR adds a multi-branch contribution gate and a release-notes parser, both now load-bearing for whether contributions and releases can proceed, and neither has committed tests. The release-mode extraction rehearsal also runs the base branch's script, so it can't actually verify the extractor changes in this PR. Add executable tests using the existing pytest/subprocess conventions: run the real gate against temporary git repos covering feature, exemption, rename, deletion, release, leftover-fragment, and normalized-version cases, and test the extractor with exact-output assertions for LF/CRLF input, section boundaries, malformed or duplicate headings and markers, and heading-only bodies. Include at least one round trip through a pinned towncrier build into the extractor.

**b5 — BLOCKING: PR CI did not pass.** Changelog #35189040711 attempt 2 completed as failure: https://github.com/microsoft/conductor/actions/runs/35189040711. Fix the CI problem on this commit, then re-verify before merge.

**b6 — BLOCKING: PR check did not pass.** Changelog / Changelog: FAILURE — https://github.com/microsoft/conductor/actions/runs/35189040711/job/105222689627. Same as above: resolve, rerun, confirm green before merge.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,192 @@
...
+            else
+              fail "Release mode: .github/scripts/extract-release-notes.sh is missing on the base branch $BASE — it must land on main before release-prep PRs can pass."
+            fi
+            if ! bash "$RUNNER_TEMP/extract-release-notes.sh" "$VERSION" CHANGELOG.md "$RUNNER_TEMP/notes.md"; then
+              fail "Release mode: the base-branch release-notes extractor could not extract the '## [$VERSION] - YYYY-MM-DD' section from CHANGELOG.md. Run: make changelog-build VERSION=$VERSION and commit the result. Contract: changelog.d/README.md"
+            fi
+
+            # 6. uv.lock must pin OUR package at the bumped version.
+            lock_version=$(awk '/^name = "conductor-cli"$/{f=1} f && /^version = /{print; exit}' uv.lock)
+            if ! grep -q "\"$VERSION\"" <<< "$lock_version"; then
```
jrob5756: **BLOCKING**: The release checklist explicitly supports prerelease versions like `0.2.0-beta.1` in pyproject.toml, but uv normalizes that to `0.2.0b1` in uv.lock. This comparison treats the two as a mismatch, so a correctly-synced lockfile still fails the check — and re-running `uv lock`, which is what the diagnostic tells you to do, won't change the outcome.

Parse both versions with a TOML parser and compare their normalized PEP 440 forms instead of comparing raw strings. Keep the authored spelling (`0.2.0-beta.1`) for the tag and changelog heading — only the comparison itself needs normalizing.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,192 @@
...
+              fi
+            done < "$NEW_FRAGMENTS_FILE"
+
+            # 4. towncrier validates the whole fragments directory against
+            #    the base branch. `towncrier check` validates the whole
+            #    resulting directory (including rename destinations).
+            #    Skipped when changelog.d/ is completely untouched, or when
+            #    the only change under changelog.d/ is the contract README.
+            if ! git diff --quiet "$BASE"...HEAD -- changelog.d/ ':(exclude)changelog.d/README.md'; then
+              uvx --from towncrier==25.8.0 towncrier check --compare-with "$BASE"
```
jrob5756: **BLOCKING**: A PR that only deletes an obsolete fragment still runs `towncrier check` here, even with the `changelog-not-required` label attached. Towncrier needs a changed fragment surviving in the resulting directory to consider the check satisfied, so it exits with "No new newsfragments found on this branch" — which reintroduces exactly the fragment requirement the label is supposed to waive.

Split this into two checks: always validate that any surviving fragment filenames are well-formed, but only require a *new* fragment when the exemption label isn't present.

docs/release-checklist.md
```diff
@@ -185,8 +187,62 @@ a **pre-release** automatically (`--prerelease`).
...
 
+- **Release workflow failed during release notes extraction**:
+  If the `release` job fails at the extraction step with "section not found or ambiguous", the tag was pushed without a compiled changelog section or the tag version does not match `pyproject.toml`.
+  - Cause: `make changelog-build` was omitted during the release-prep PR, or the tag was created with a version mismatch.
+  - Remedy: Delete the tag locally and remotely (see below), create a release-prep PR on `main` to run `make changelog-build VERSION=X.Y.Z`, merge it, and re-tag the new merge commit.
+
+- **Release-prep PR blocked by the `Changelog` workflow**:
+  If CI rejects the release-prep PR in release mode, check the job's log message:
+  - Cause: Leftover fragments remain under `changelog.d/` (for instance, if another PR merged after `make changelog-build` ran), or the version in `uv.lock` does not match `pyproject.toml`.
+  - Remedy: Re-run `make changelog-build VERSION=X.Y.Z` on the branch to consume the new fragments, run `uv lock`, commit the changes, and update the PR.
```
jrob5756: **RECOMMENDED**: Once a release section is compiled, re-running towncrier for the same version won't append fragments that arrived afterward — same day, it rejects the existing header; later, it can produce a duplicate version section that CI then rejects. So the recovery instructions here leave the release-prep PR stuck, and the fragments already consumed by the first compile aren't available for a clean rebuild.

Document reverting the previous compilation, restoring all the fragments it consumed, pulling in whatever landed on the base branch since, and compiling the full set in one pass. If the mismatch is lockfile-only, the fix is just `uv lock`, nothing more. The equivalent recovery message in changelog.yml should say the same thing.

docs/release-checklist.md
```diff
@@ -56,39 +59,39 @@ so there is **no** separate `__version__` string to edit.
...
   ```
 
-- [ ] Add a **fresh, empty** `[Unreleased]` section above it, comparing the new
-      tag to `HEAD`:
+- [ ] Review the diff:
 
-  ```markdown
-  ## [Unreleased](https://github.com/microsoft/conductor/compare/v0.1.20...HEAD)
+  ```bash
+  git diff CHANGELOG.md changelog.d/
```
jrob5756: **RECOMMENDED**: Towncrier stages the generated CHANGELOG.md and the fragment deletions it makes. A plain `git diff` doesn't show staged changes, so right after compilation this command usually shows nothing at all — not the release content the maintainer is supposed to be reviewing at this step.

```suggestion
git diff HEAD -- CHANGELOG.md changelog.d/
```
Worth a line noting that compilation stages its own changes automatically, so readers aren't confused by the empty diff they'd otherwise get.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,192 @@
+# Changelog fragment contract enforcement for pull requests.
+#
+# Every PR adds a towncrier fragment in changelog.d/ instead of editing
+# CHANGELOG.md (feature/fix PRs), or compiles the fragments into CHANGELOG.md
+# (release-prep PRs, detected by a pyproject.toml version bump). See
+# changelog.d/README.md for the naming contract this job enforces.
+#
+# Fork-safe by construction: `pull_request` (never `pull_request_target`),
+# `contents: read` only, no secrets, and code introduced by the PR is never
```
jrob5756: **RECOMMENDED**: Saying PR-introduced code is never executed isn't quite true for a `pull_request` workflow — the PR can edit the workflow file itself. Loading the extractor from the base branch protects that one script, not the inline Bash around it, so this comment gives maintainers a more confident picture of the trust boundary than actually exists.

```suggestion
# This workflow requests read-only permissions and avoids invoking checked-out application
# code. The release-notes extractor is loaded from the base branch rather than the PR's
# checkout. The workflow definition itself is still under PR control, though — a modified
# workflow file could still run arbitrary steps.
```

.github/workflows/changelog.yml
hertznsk: Correct — the typo made the action unresolvable, so the job died during setup before it could enforce anything. Fixed in 321ecf6 by pinning the same v7.0.1 SHA used throughout `ci.yml` and `release.yml` (`3d3c42e5aac5ba805825da76410c181273ba90b1`, verified byte-identical to the 13 existing pins).

.github/workflows/changelog.yml
hertznsk: Correct — uv normalizes `0.2.0-beta.1` to `0.2.0b1` at lock time, so the raw string compare rejected a correctly-synced lockfile and the diagnostic's own remedy (`uv lock`) could never change the outcome. Fixed in 9dd4dd2: both versions are now parsed (tomllib + `packaging.Version`) and compared under PEP 440 normalization, while the authored spelling stays authoritative for the tag and the CHANGELOG.md section header. Missing lockfile entry and parser failure are distinct failure branches. This is pinned by `test_release_with_normalized_prerelease_passes` (pyproject `0.2.0-beta.1` + lock `0.2.0b1` passes).

.github/workflows/changelog.yml
hertznsk: Correct — verified against the towncrier 25.8.0 source: `check` exits 1 with "No new newsfragments found on this branch" whenever the branch adds no fragment, even for a perfectly valid deletion-only change, so the label path reintroduced exactly the requirement it waives. Fixed in 4fc623a with the split you proposed: whenever `changelog.d/` changed, the gate now validates every surviving fragment filename in the resulting tree itself (additions, rename destinations, and names inherited from the base branch; mirroring towncrier's own ignore list, case-insensitively), and `towncrier check` runs only when the PR adds at least one fragment, remaining the ground-truth validation on that path.

A second self-review round (7f046ae) hardened two adjacent edges found while testing this split: the scan now rejects fragments nested in a subdirectory (a rename into `changelog.d/sub/` under the label would pass the scan but plant a tree on main where `towncrier check` fails for every later PR — towncrier parses the subdirectory's name as a fragment), and the label match is now case-insensitive, matching GitHub's own label-name uniqueness semantics. Covered by executable tests: deletion-only with/without the label, valid and invalid renames, the nested-path case, and README-only changes.

docs/release-checklist.md
hertznsk: Correct — verified empirically against towncrier 25.8.0: a same-day re-run fails with "It seems you've already produced newsfiles for this version?", and a re-run on a later date appends a duplicate version section that CI then rejects. The runbook is rewritten in 8395702 to prescribe exactly the recovery you described: rebase onto the base branch, `git checkout origin/main -- CHANGELOG.md changelog.d/` (reverts the previous compilation and restores every fragment it consumed, including any that landed on the base branch since), then compile the full set in one pass and `uv lock`; a lockfile-only mismatch is just `uv lock`. The gate's leftover-fragment and duplicate-section failure messages say the same thing and name the runbook section (the messages moved into `.github/scripts/enforce-changelog-contract.sh` when the gate was extracted for testability). The round-trip test `test_towncrier_build_then_extract_round_trip` pins both the refusal and that the documented remedy actually produces a single clean section.

docs/release-checklist.md
hertznsk: Correct — verified: towncrier stages the compiled `CHANGELOG.md` and the fragment deletions, so a plain `git diff` shows nothing right after compilation. Applied your suggestion in 8395702 (`git diff HEAD -- CHANGELOG.md changelog.d/`) plus the line noting that towncrier stages its own changes, and the towncrier round-trip test pins the staging behavior so the doc cannot silently drift from reality.

.github/workflows/changelog.yml
hertznsk: Correct — on `pull_request` the workflow definition itself is under PR control, so the claim overstated the boundary. Applied your suggested wording in 321ecf6. Note the gate logic has since moved into `.github/scripts/enforce-changelog-contract.sh` (extracted verbatim so the test suite can execute the real gate); that script is likewise loaded from the PR checkout, so the caveat covers it. The only base-branch loading that remains is the release-notes extractor rehearsal, as before.

hertznsk:
Thanks for the thorough review — every blocking item was correct. All fixed and pushed (the branch was rebased onto current `main` first, see below). Replies with details are in the individual threads; here is the summary-level picture:

**b4 — test coverage.** Added in 5ff445e, extended in 7f046ae. The gate logic moved verbatim from `changelog.yml` into `.github/scripts/enforce-changelog-contract.sh` (3aaf24b) so the tests execute the real gate, not a copy: `tests/test_integration/test_changelog_gate.py` runs it via subprocess against temporary git repositories and covers feature mode (valid/invalid/missing fragments, CHANGELOG.md edit, exemption label, deletion-only and rename-only changes with and without the label, README-only changes, nested-path rejection, duplicate-counter rejection via towncrier) and release mode (happy path, missing CHANGELOG change, duplicate section, missing marker, leftover fragments, lockfile mismatch, PEP 440-normalized prerelease, missing base-branch extractor). `tests/test_integration/test_extract_release_notes.py` pins the extractor with exact-output assertions (LF and CRLF, section boundaries, missing/duplicate sections and markers, section above the marker, heading-only body, legacy compare-link headers, regex-metacharacter decoy) and includes the requested round trip through a pinned `towncrier==25.8.0` build into the extractor — which additionally pins the two towncrier behaviors the recovery runbook depends on (builds stage their changes; a repeated same-version build fails until the pre-compile state is restored). 36/36 pass locally; `make check` is green.

**b5/b6 — failing CI.** Root cause was the checkout pin typo (thread 1), fixed in 321ecf6; the job never got past action resolution. One heads-up on re-verification: this PR edits `CHANGELOG.md` (the migration), so the fixed gate will fail this PR in feature mode until the documented bootstrap step happens — a maintainer creating the `changelog-not-required` label upstream and applying it here (the workflow re-runs on `labeled`). I verified the gate locally against this exact branch: it fails without the label (as designed) and passes with it, with `towncrier check` exiting "Checks SKIPPED" because the news file changed — the fragments are validated by the surviving-tree name scan.

**Rebase.** The branch is rebased onto `origin/main` (e8f02e1). The `[Unreleased]` entries that landed meanwhile (#531 compaction fix, #517 step-definition union) are converted to fragments (`531.fixed.md`, `517.changed.md`) per the new contract, so `CHANGELOG.md` carries only the static header and marker; fragment texts are verbatim with bullets/indentation stripped and the issue reference dropped where towncrier renders it from the filename.

**Second-round self-review.** After addressing your comments I ran an adversarial review pass over the fixes themselves; the edge cases it found (nested-path rename hole in the surviving-tree scan, lockfile parser errors misreported as mismatches, towncrier check failures lacking an `::error::` annotation, case-sensitive label match) are fixed in 7f046ae with tests.

PR #536, opened 2026-09-17
fix(installer): improve private-index recovery guidance

oleksandr-bilyk:
## Summary
- show a credential-safe command for reusing an existing pip mirror when uv cannot reach the public package index
- recommend `UV_SYSTEM_CERTS=1` for current uv releases while retaining the legacy setting for uv versions older than 0.11
- cover the new recovery guidance in installer integration and static tests

## Validation
- `uv run --frozen ruff check tests/test_integration/test_install_scripts.py tests/test_integration/test_install_script_files.py`
- `uv run --frozen ty check tests/test_integration/test_install_scripts.py tests/test_integration/test_install_script_files.py`
- PowerShell parser validation for `install.ps1`
- `sh -n install.sh`
- `uv run --frozen pytest -q tests/test_integration/test_install_script_files.py` (12 passed)
- `uv run --frozen pytest -q -m install_scripts tests/test_integration/test_install_scripts.py -k "blocked_package_index_reports_guidance or blocked_index_offers_credential_safe_pip_bridge"` (2 passed)
- Windows rename-fallback integration test (1 passed with the configured package mirror)


install.sh
```diff
@@ -316,6 +316,11 @@ print_index_guidance() {
         printf '      curl -sSfL https://aka.ms/conductor/install.sh | sh\n\n' >&2
         printf '  Add the same export to your shell profile so future installs and\n' >&2
         printf "  'conductor update --apply' inherit it.\n\n" >&2
+        printf "  If pip already works through your organization's mirror, reuse its\n" >&2
+        printf '  configured URL without printing credentials into this installer log:\n\n' >&2
+        printf '      pip_index="$(pip config get global.index-url)"\n' >&2
+        printf '      export UV_DEFAULT_INDEX="internal=${pip_index}"\n' >&2
```
jrob5756: **BLOCKING**: This assumes pip already has `global.index-url` set, but pip can just as easily be working through `PIP_INDEX_URL` or `install.index-url` alone. In that case the lookup returns nothing, and the lines after it still export `UV_DEFAULT_INDEX=internal=` and rerun the installer — so a working pip config turns into a broken uv one, swapping a clear network error for an invalid-index error that persists for the rest of the shell.

Only assign `UV_DEFAULT_INDEX` and rerun when the lookup actually returns a nonempty value. If it comes back empty, leave the environment untouched and point the user at the explicit approved-mirror instructions instead. The PowerShell block and the README have the identical problem and need the same fix.

tests/test_integration/test_install_scripts.py
```diff
@@ -381,6 +381,21 @@ def test_blocked_package_index_reports_guidance_and_skips_retries(
...
+    """Offer a bridge without reading or printing the configured index URL."""
+    shim_env = _install_uv_shim(sandbox, mode="network-always")
+
+    result = run_install_script(sandbox, source=wheels.new, force=True, extra_env=shim_env)
+
+    assert result.returncode != 0, f"install should have failed:\n{result.combined}"
+    combined = result.combined
+    assert "If pip already works" in combined
+    assert "pip config get global.index-url" in combined
+    assert "UV_DEFAULT_INDEX" in combined
```
jrob5756: **RECOMMENDED**: This test still passes if the `UV_DEFAULT_INDEX` assignment this PR adds gets deleted — `UV_DEFAULT_INDEX` shows up elsewhere in the guidance text already, and the rest of the assertions only check for the intro copy and the pip lookup. So the one thing this PR actually changes, wiring the captured pip URL into uv, isn't under test.

Assert the full rendered capture-and-assign sequence per platform, including the `internal=` prefix and the literal `${pip_index}` / `$pipIndex` reference, so a regression here actually fails the test.

jrob5756 (commented):
One blocking issue and one test-coverage gap, both around the new pip-to-uv index bridging logic.

**Blocking:**
- `install.sh:322` — the pip index lookup isn't guarded before `UV_DEFAULT_INDEX` gets set and the installer reruns.

install.sh
oleksandr-bilyk: Fixed in d398aab. The rendered PowerShell and POSIX snippets now suppress lookup errors, check that the captured value is nonempty, and only then set UV_DEFAULT_INDEX and rerun. The empty case leaves the environment untouched and points back to the explicit approved-mirror instructions.

tests/test_integration/test_install_scripts.py
oleksandr-bilyk: Fixed in d398aab. The integration test now asserts the complete rendered capture, nonempty guard, internal= assignment, rerun, and empty-value fallback sequence separately for PowerShell and POSIX.

jrob5756 (commented):
Two things from last time are done: the pip-to-uv bridge now guards against an empty lookup, and there's a test asserting the full rendered bridge sequence, not just pieces of it.

Everything else is still open.

CI is red on this commit, reported four separate ways:
- Install Scripts (ubuntu-latest) failed on run 35366975168, attempt 2: https://github.com/microsoft/conductor/actions/runs/35366975168/job/105685803280
- Changelog failed on run 35366975254, attempt 2: https://github.com/microsoft/conductor/actions/runs/35366975254/job/105685698456
- The check "CI / Install Scripts (ubuntu-latest)" is showing FAILURE for that same job.
- The check "Changelog / Changelog" is showing FAILURE for that same job.

All four point at the same two failing jobs. Get the commit green, then re-run review and verify checks before merging.

One more item carried over, recommended rather than blocking: `install.sh:322` — the pip index lookup isn't guarded before `UV_DEFAULT_INDEX` gets set and the installer reruns. There's a test-coverage gap alongside it. The follow-up pass didn't examine this one, so it's coming through unexamined — take a look yourself before deciding whether it needs to block.

oleksandr-bilyk:
@jrob5756 Addressed the two red checks in a6a4303:

- fixed the POSIX nested `printf` rendering by using a fixed outer format, so the generated snippet preserves the literal `%s\n`
- added `changelog.d/+installer-index-guidance.fixed.md` for the new changelog contract

Local validation passed: changelog contract, `sh -n install.sh`, direct POSIX rendering assertion, static installer tests (12 passed), and focused blocked-index integration tests (2 passed).

The refreshed CI and Changelog runs are currently `action_required` with no jobs created and need maintainer approval. Please approve the fork workflows and re-review once they complete.

jrob5756 (commented):
The empty pip index lookup is guarded now, coverage requires the full rendered bridge, and the changelog issue is fixed. Good progress.

One item from the earlier pass is still open. It was flagged recommended, but it actually bundles a blocking issue with a coverage gap, both in the new pip-to-uv index bridging logic: at `install.sh:322`, the pip index lookup still isn't guarded before `UV_DEFAULT_INDEX` gets set and the installer reruns. The follow-up analysis didn't check whether this landed, so I can't confirm it from here — worth a manual look before closing it out.

**Still open:**
- `install.sh:322` — pip index lookup guard (blocking, per the original point)

oleksandr-bilyk:
Manual Windows validation completed on September 22, 2026:
• Uninstalled the existing  conductor-cli  uv tool.
• Reinstalled from the current  fix/installer-index-guidance  checkout using:
 .\install.ps1 -Source C:\dev\conductor -Force -SkipPathUpdate 
• The installer detected  UV_DEFAULT_INDEX=internal=https://packagefeedproxy.microsoft.io/pypi/simple/ .
• Installation completed successfully.
• Verification passed:  Conductor v0.1.37 .
This validates the local-source installer path and internal-index handling on Windows. All CI checks are also green

Issue #540, opened 2026-09-20
Test suite: chmod-based unreadable-directory tests fail (Path.exists() swallows EACCES) and performance-marked tests run under tight thresholds


hertznsk:
## Summary

Four tests in the default `make test` suite fail reproducibly (2) or intermittently (2) in this environment. None are caused by application behavior changes; they are test-suite fragility issues.

## Environment

- Linux (x86_64, /tmp mounted as tmpfs)
- Python 3.14.7, pytest 9.0.3
- Reproduces on current `main` (`87f7788`)

## 1. chmod-based "unreadable directory" tests fail on any standard POSIX setup

Failing tests:

- `tests/test_skills/test_path_entries.py::TestUnreadableParent::test_unreadable_parent_is_reported_not_raised_raw`
- `tests/test_plugins/test_registry.py::TestUnreadableTrees::test_unreadable_skill_subdirectory_is_reported`

Both do `parent.chmod(0o000)` and expect `SkillNotFoundError` / `PluginManifestError` with the message "could not be read".

**Root cause.** In `src/conductor/skills/registry.py::_resolve_path_entry` (and the plugins counterpart) the code comment states:

> Python re-raises EACCES from ``exists``, ``is_dir``, ``is_file`` and ``iterdir`` alike

That is only true for `is_dir()` / `is_file()` / `iterdir()`. `pathlib.Path.exists()` **swallows every `OSError` (including `PermissionError`) and returns `False`**. So with an unreadable parent directory, `resolved.exists()` returns `False` and the code raises the *"does not exist"* variant instead of the *"could not be read"* variant — the tests' `match="could not be read"` fails.

Minimal repro of the underlying behavior:

```python
import os, pathlib, tempfile
t = tempfile.mkdtemp()
os.makedirs(f"{t}/locked/acme")
os.chmod(f"{t}/locked", 0o000)
os.stat(f"{t}/locked/acme")            # raises PermissionError(13, 'Permission denied')
print(pathlib.Path(f"{t}/locked/acme").exists())   # -> False (EACCES swallowed)
```

Observed failure:

```
E  conductor.skills.registry.SkillNotFoundError: Skill path '.../locked/acme' resolved to .../locked/acme, which does not exist. ...
```

**Suggested fixes** (any one):

- Probe readability explicitly before the exists check (e.g. attempt `os.stat(resolved)` inside the existing `except OSError` guard and raise "could not be read" from there), or
[...]

Issue #527, opened 2026-09-13
Idea: GitOps-native runner backends and reproducible execution environments


hertznsk:
# Idea: GitOps-native runner backends and reproducible execution environments

## Summary

Conductor should support reproducible local and CI execution through pluggable runner backends without becoming a mandatory hosted platform.

The proposed direction keeps Conductor CLI-first and GitOps-native:

- workflow YAML remains the reviewable orchestration definition;
- the existing `WorkflowEngine` remains the single owner of routes, loop-backs, parallel and for-each groups, gates, checkpoints, and final output;
- a selected execution environment resolves logical profiles into concrete images, resources, network/security policy, and secret bindings;
- local, Docker, and Kubernetes runners implement the same typed execution contract;
- an immutable run manifest and content-addressed bundle make the exact execution inputs inspectable and auditable;
- no database, broker, dashboard, or always-on Conductor control plane is required.

The intended product model is closer to **Ansible or GitLab CI for AI agents** than to a visual workflow platform: source-controlled declarative workflows, inexpensive local execution, and consistent orchestration semantics in developer and production environments.

This issue proposes architecture and behavioral contracts for discussion. It intentionally does not prescribe a PR sequence.

## Motivation

Running Conductor on a developer workstation makes automation depend on mutable host state: installed Python and Node versions, Bash/system tools, stdio MCP binaries, ambient credentials, filesystem layout, and the lifetime of the developer's process.

A single universal image does not solve this cleanly. A minimal image lacks common dependencies; a broadly useful image becomes a combination of multiple language runtimes, cloud CLIs, and organization-specific MCP servers.

Per-script containers alone are also insufficient. A script may clone or modify a repository in one filesystem while the next agent and its filesystem MCP server remain on the host and see another filesystem. Execution placement must cover the complete runtime consuming a workspace, not only `type: script`.

The current implementation exposes the missing contracts:

- `ScriptExecutor` starts local subprocesses and inherits the host environment.
- Agent providers may own SDK/CLI processes, sessions, tools, stdio MCP lifecycle, skills, and plugins.
- Direct MCP steps pool sessions by server and working directory.
- `working_dir` has different effective meanings for local agents, scripts, and ACA sandboxes.
- Checkpoints preserve logical workflow state, but not filesystem state required by completed steps.

These behaviors are valid for today's local runner, but insufficient for reproducible remote execution and workspace-safe resume.

## Product boundary

[...]

PR #541, opened 2026-09-20
refactor(execution): runner backend contract with engine-owned workspace lease

hertznsk:
> **Part of #527** — the first PR in the planned execution-backend cycle (the 9-step sequence outlined in the issue discussion). It lands the contract and the local reference implementation only; the remaining steps (profiles, remote realms, run manifest, workspace persistence, agent/MCP placement, ACA convergence) build on this seam.

## Summary

Introduces `conductor.execution` — a stdlib-only execution-backend seam that script steps now run through, with the workflow engine owning the backend and a run-scoped workspace lease. This is the foundation for the future execution architecture (profiles, Docker/Kubernetes realms, ACA convergence): everything downstream will plug into this seam instead of growing conditional branches in the engine.

**No behavior change.** Rendered commands, environment handling, timeouts, error messages, and exit-code routing are byte-identical; the only internal difference is that the script child process is now always killed and reaped when the run task is cancelled (previously it could be left orphaned).

## Changes (3 atomic commits)

**`refactor(execution): introduce the runner backend contract`**
- New package `src/conductor/execution/` (pure leaf: stdlib-only, no conductor imports): frozen dataclass contract types (`CommandSpec`, `CommandResult`, `StartError` — a serializable, transport-neutral start-failure record used to reconstruct `__cause__`, `WorkspaceLease`, `RunSpec`, `RunnerCapabilities`), `CommandOutcome`/`RunOutcome` literal aliases, and the `RunnerBackend` protocol (`capabilities` / `prepare_run` / `run_command` / `finalize_run`).
- Contract tests: frozen-instance guarantees, leaf purity (subprocess + AST negative control), `StartError` round-trip reconstruction.

**`refactor(executor): delegate script-step execution to the local runner backend`**
- `LocalRunnerBackend`: the subprocess logic extracted verbatim from `ScriptExecutor` (env merge order, PATH resolution, spawn, `communicate`, per-command timeout → kill+wait, `utf-8`/`errors="replace"` decode, duration), plus kill+reap on `asyncio.CancelledError` under `suppress(ProcessLookupError)`.
- `ScriptExecutor` keeps its public contract (`execute(agent, context) -> ScriptOutput`, no-arg constructor) and delegates to the backend; outcomes map back to byte-identical `ExecutionError` messages with `__cause__` reconstructed from `StartError`. Module-level `os`/`shutil` remain as patch anchors so the existing test suite patches the same singletons the backend uses.
- Contract tests: all failure outcomes, kill-on-cancel, `__cause__` equivalence, diagnostics callback, legacy patch-target compatibility.
- `AGENTS.md`: new `execution/` bullet + corrected `script.py` bullet.

**`refactor(engine): own the execution backend and workspace lease at run scope`**
- `WorkflowEngine` accepts an optional `execution_backend` and prepares a `WorkspaceLease` in `run()`/`resume()` (root engine only), finalizing it in the matching `finally` with the run outcome (`succeeded`/`failed`/`cancelled`; best-effort, logged, never masking the run outcome). Both sub-workflow child constructors inherit backend + lease like `_guidance_channel`; `_execute_script` threads the lease through. `cli/run.py` untouched.
- Integration tests: lease lifecycle on run **and** resume, exactly-one finalize on cancellation, no finalize when `prepare_run` raises, root-only prepare/finalize, both child-constructor inheritance paths.
- `AGENTS.md` engine-ownership facts + changelog fragment.

## Verification

- **Parity gate:** `tests/test_executor/test_script.py` (~700 lines) passes **unmodified**; base-aware test-scope check shows only added files under `tests/test_execution/` (61 new tests, every one with an English requirement-pinning comment). No existing test was touched.
- Full suite green: 9341 passed — the only failures are 4 pre-existing environment-sensitive tests unrelated to this change (filed separately as #540).
- `make check` (ruff + ty) and `make validate-examples` green.
- **Real-run parity:** a purpose-built provider-free workflow (`set` → `script` echo → `script` exit 42 with `exit_code` routing → `$end`) produces byte-identical output and exit codes on this branch and on the base revision.
- **Per-commit atomicity:** `make check` + the full test suite verified green on each of the three commits individually (9275 / 9302 / 9311 passed), not just the final state.

## Notes for reviewers

- One lease per run is a deliberate step-1 simplification (comment in `workflow.py`); the profiles step will replace it with a backend/realm→lease mapping.
- `StartErrorKind`/aliases carry attribute docstrings following the `RunMode` convention in `fleet/records.py`.


jrob5756 (commented):
This one isn't mergeable as-is: CI is red on the reviewed commit, and there's a real hang risk in the new cancellation path. The workspace-lease reuse bug is worth fixing too, but it's lower stakes than the other two.

**Blocking findings:**
- `src/conductor/execution/local.py:147` — cancellation can hang waiting on undrained subprocess pipes (inline)
- CI failure on this commit (no line to anchor to, written out below)

**PR CI did not pass** — CI #35512086754 attempt 2: completed/failure (https://github.com/microsoft/conductor/actions/runs/35512086754). Test (Python 3.13, windows-latest): completed/failure (https://github.com/microsoft/conductor/actions/runs/35512086754/job/106332310505). Fix the CI problem on this commit, then rerun review and confirm checks before merging.

**PR check did not pass** — CI / Test (Python 3.13, windows-latest): FAILURE (https://github.com/microsoft/conductor/actions/runs/35512086754/job/106332310505). Same issue: resolve it on this commit and verify checks pass before merge.

The two recommended items are inline: a resumed/re-run engine reusing an already-finalized workspace lease (`workflow.py:3424`), and cancellation during finalization interrupting the backend's own cleanup (`workflow.py:3615`).

src/conductor/execution/local.py
```diff
@@ -0,0 +1,165 @@
...
+            return CommandResult(
+                outcome="timed_out",
+                resolved_command=resolved_command,
+                duration_seconds=time.monotonic() - started_at,
+            )
+        except asyncio.CancelledError:
+            with contextlib.suppress(ProcessLookupError):
+                process.kill()
+            with contextlib.suppress(BaseException):
+                await process.wait()
```
jrob5756: **BLOCKING**: Cancelling `run_command()` cancels `communicate()`, which stops stdout/stderr drainage. If a high-output child fills the pipe buffers during shutdown, asyncio stops reading and `process.wait()` can stay blocked even after `process.kill()` has already killed the child — a continuously-writing Python child reproduces this with `returncode` already `-9`. So this handler can turn a dashboard Kill or a workflow timeout into a hang instead of propagating the cancellation, and it also blocks workspace finalization from ever running.

Suggested fix: run `communicate()` in a separately shielded task. On cancellation, kill the child, then await pipe drainage and process termination before re-raising `CancelledError`. Add a regression test that cancels a subprocess that keeps writing continuously.

src/conductor/engine/workflow.py
```diff
@@ -3351,13 +3415,36 @@ async def resume(self, current_agent_name: str) -> dict[str, Any]:
         # Fresh timeout window for resumed execution
         self.limits.start_time = _time.monotonic()
 
+        # Same outcome tracking as :meth:`run` — resume must lifecycle the
+        # execution backend identically, or resumed runs would silently skip
+        # prepare/finalize (run/resume parity).
+        outcome: RunOutcome = "failed"
         try:
+            if self._subworkflow_depth == 0:
+                self._workspace_lease = self._workspace_lease or (
```
jrob5756, line 3428: **RECOMMENDED**: Finalization leaves `self._workspace_lease` populated. If an embedding caller resumes a failed run on the same engine instance, this expression skips `prepare_run()` and hands the already-finalized lease to the next commands. A backend that releases its workspace resources at finalization time gets an invalid handle here, and the engine goes on to finalize that same handle a second time. Repeated `run()` calls on the same engine hit this too.

Suggested fix: clear the lease from engine state as soon as finalization starts (keep a local reference for cleanup), and make sure every subsequent root `run()`/`resume()` acquires a fresh lease. Add coverage using a backend that rejects already-finalized handles.

src/conductor/engine/workflow.py
```diff
@@ -3514,6 +3601,26 @@ def _save_checkpoint_on_failure(self, error: BaseException) -> None:
...
+
+        A no-op unless a lease was actually prepared: a ``prepare_run`` that
+        raised leaves no lease to finalize, and ``finalize_run`` must never be
+        handed ``None``. Failures are logged, never raised — realm cleanup
+        cannot mask the run's own outcome.
+        """
+        if self._workspace_lease is None:
+            return
+        try:
+            await self._execution_backend.finalize_run(self._workspace_lease, outcome)
```
jrob5756: **RECOMMENDED**: The backend finalizer is awaited directly on the engine task. If cancellation arrives while an async `finalize_run()` is mid-cleanup, it cancels that cleanup outright — `except Exception` doesn't catch `CancelledError`. Same problem if a second cancellation lands during teardown of a run that's already been cancelled once. A resource-owning backend can end up leaving its workspace allocated, which breaks the engine's own lifecycle guarantee.

Suggested fix: run finalization in a shielded task and await it to completion even under cancellation, then re-raise the pending cancellation afterward. Add a test for cancellation arriving after finalization has already started, not just during command execution.

src/conductor/execution/local.py
hertznsk: Fixed in 6c1f394. `communicate()` now runs as its own shielded task, so neither the per-command `wait_for` timeout nor an outer cancellation stops pipe drainage. Both teardown paths go through a shared `_kill_and_reap`: kill the child, keep draining the still-running `communicate()` task to EOF, then finish a shielded `process.wait()` — the read side never pauses, so a flooded child can no longer wedge the reap. A cancellation absorbed during that cleanup is re-raised afterward (on the timeout path it is converted back from the data-shaped `timed_out` result into `CancelledError`, never swallowed).

Regression coverage: `test_cancel_drains_flooded_pipes_and_reaps_child` cancels a real continuously-writing child (cross-platform, including the Windows proactor path) and asserts prompt cancellation, a reaped returncode, and EOF observed on both `StreamReader`s. Verified against the pre-fix code, where it fails with exactly the state you described — `<StreamReader 196608 bytes transport=<_UnixReadPipeTransport ...> paused>`, EOF never fed. A second mock-based test (`test_cancel_during_timeout_cleanup_reraises_cancellation`) pins a cancellation arriving mid-timeout-cleanup.

src/conductor/engine/workflow.py
hertznsk: Fixed in 97b4bc7. `_finalize_execution_backend` now detaches the lease from engine state *before* finalization begins (`lease = self._workspace_lease; self._workspace_lease = None`, keeping the local reference for cleanup), so a repeated root `run()`/`resume()` on the same engine instance always calls `prepare_run` fresh, and an already-finalized handle is never handed to another command or finalized twice.

Covered by `test_second_run_prepares_a_fresh_lease`, which uses a backend that mints a fresh lease per `prepare_run` and raises if a finalized handle ever reaches `run_command`/`finalize_run` — two sequential `run()` calls on one engine produce two distinct leases and two finalize calls, and `engine._workspace_lease` is `None` after each run.

src/conductor/engine/workflow.py
hertznsk: Fixed in 97b4bc7. `finalize_run` now runs as a shielded task, mirroring the existing `_close_evicted_mcp_step_manager` pattern in this file: a racing cancellation is absorbed until the backend's teardown completes, then re-raised afterward, so an async finalizer can't be interrupted mid-cleanup. The finalize call also moved into an inner `finally` relative to `_close_mcp_step_managers()`, so a cancellation landing in the MCP close can't skip realm cleanup either.

Covered by `test_cancellation_during_finalize_completes_cleanup` (cancel arrives while `finalize_run` is blocked mid-cleanup: the engine task stays alive behind the shield, teardown runs to completion, then `CancelledError` propagates) and `test_cancellation_during_mcp_close_still_finalizes` (cancel during the MCP-manager close: finalization still runs exactly once).

hertznsk:
All review findings are addressed and pushed (ec75df3..97b4bc7):

- **Blocking — cancellation hang in `local.py`** (6c1f394): `communicate()` now runs as a shielded task and both teardown paths (per-command timeout and outer cancellation) go through `_kill_and_reap` — kill, drain the pipes to EOF, then a shielded `process.wait()`. A racing cancellation absorbed during cleanup is re-raised afterward, never collapsed into a data-shaped `timed_out`. Regression coverage: a cross-platform test cancelling a real continuously-writing child (verified to fail on the pre-fix code, which leaves the pipe transport paused with 196 KB unread and EOF never fed — exactly the wedged state from the review), plus a mock test for cancellation arriving mid-timeout-cleanup.
- **Recommended — finalized lease reuse** (97b4bc7): `_finalize_execution_backend` detaches the lease from engine state before finalizing, so a repeated `run()`/`resume()` on the same engine prepares a fresh handle; pinned by a backend that rejects finalized leases.
- **Recommended — shielded finalization** (97b4bc7): `finalize_run` runs as a shielded task mirroring `_close_evicted_mcp_step_manager`, re-raising a pending cancellation after teardown; the call also sits in an inner `finally` so a cancellation inside `_close_mcp_step_managers()` can't skip realm cleanup. Both are covered by new engine tests.
- **CI failure (Python 3.13, windows-latest)** (27a4b32): both were Windows-only test bugs, not product bugs — `shutil.which()` only resolves PATHEXT-suffixed names on Windows (the fake `backend-python` executable stayed unresolved), and the executor appends the Windows resolution hint only on win32 while the test's expectation didn't branch on the platform. Both tests are now platform-aware; they were no-ops on Linux, which is why the Ubuntu leg stayed green.

Verification: full suite green locally — 9345 passed; the only 4 failures are the pre-existing environment-sensitive ones that reproduce identically on `main` (tracked in #540, fix proposed in #542). `make check` (ruff + ty) is clean, and every new regression test was confirmed to fail on the pre-fix code.

jrob5756 (approved):

PR #542, opened 2026-09-20
fix: restore unreadable-path diagnostics on Python 3.14

### LINKED ISSUE #540: Test suite: chmod-based unreadable-directory tests fail (Path.exists() swallows EACCES) and performance-marked tests run under tight thresholds

hertznsk:
Fixes #540.

## Summary

Two independent fixes for the test-suite failures reported in the issue, both reproduced locally on Python 3.14.7.

### 1. chmod-based unreadable-directory tests

Python 3.14's rewritten `pathlib` delegates `Path.exists()` / `is_dir()` / `is_file()` to `os.path`, which swallows **every** `OSError` — including `PermissionError` — and returns `False`. On 3.13 and earlier those probes re-raised anything outside a small allowlist (`ENOENT`/`ENOTDIR`/`EBADF`/`ELOOP`, plus winerrors 21/123/1921 per `Lib/pathlib/_abc.py::_ignore_error`), so a permissions problem surfaced instead of impersonating a missing path.

Conductor's path diagnostics distinguish *"does not exist"* from *"could not be read"*, and the 3.14 change broke that in observable ways beyond the two failing tests: an unreadable skills-root subdirectory was silently misfiled as a near-miss `skipped` entry, and an unreadable plugin `skills/` directory silently resolved as a plugin with no skills.

The fix adds `conductor/filesystem.py` — a small stdlib-only leaf module (in the style of `duration.py`/`console.py`/`rundir.py`) providing `stat_or_none()` plus `exists_strict()`/`is_dir_strict()`/`is_file_strict()` predicates that mirror the pre-3.14 allowlist exactly, on every interpreter. All call sites where the two error variants must stay distinct now use the strict probes:

- `skills/registry.py` — `_resolve_path_entry` (one stat answers both "exists?" and "is a directory?"), `expand_skills_root`
- `plugins/registry.py` — `_resolve_path_entry`, `_plugin_skills`
- `plugins/resolution.py` — `_local_root`
- `skills/discovery.py` — `_scan_root` (keeps the `failed=True` branch reachable), `_has_repo_marker` (keeps the warning reachable)

Call sites deliberately left alone: `find_manifest`'s probe is already wrapped in `except OSError: continue` with identical fall-through semantics on both versions; `manifest.py`'s `.mcp.json` fallback and `frontmatter.py`'s `SKILL.md` check still raise via the immediately-following `read_text`; `_subdirectory` never had a "could not be read" contract (raw `PermissionError` escaped there pre-3.14 too).

### 2. Performance-marked tests under `make test`

`make test` / `make test-cov` now run `-m "not install_scripts and not performance"`, matching the marker expression CI already uses and the exclusion mechanism `AGENTS.md` documents. The stale references to the old expression in `tests/conftest.py` and `test_install_scripts_marker.py` are updated (the latter now exercises the exact new `make test` expression).

## Test plan

- Both originally-failing chmod tests pass on Python 3.14.7
- New `tests/test_filesystem.py` (23 tests): monkeypatch-based, chmod-independent — errno/winerror allowlist policy, `EACCES` propagation from every probe, symlink semantics, and call-site wiring including a mutation-checked regression test for the plugin `skills/` case
- Full `make test`: **9275 passed, 35 skipped**
- `make lint` and `ty check` clean on changed files; `make validate-examples` passes; the changelog contract script passes against `origin/main`


jrob5756 (commented):
One recommended finding here, no blockers, so this can merge as-is if the author wants to leave the follow-up for later.

### Recommended (no line to anchor to)

**Complete the strict-probe migration for plugin agents and MCP configuration** — `src/conductor/plugins/agents.py::read_plugin_agents()` and `src/conductor/plugins/manifest.py::_load_mcp_servers()` still have the Python 3.14 diagnostic bug this PR fixes elsewhere. If `agents/`, an agent definition file, or the conventional `.mcp.json` is a symlink into a directory you can't read, the remaining pathlib probes swallow `PermissionError` instead of raising `PluginManifestError`. The result is silent: agents or MCP servers just disappear from resolution rather than the run failing loudly. The MCP case is specifically the path where the manifest has no `mcpServers` key — the fallback returns before `read_text()` runs, so that read never gets a chance to preserve the diagnostic.

Suggested fix: swap in `is_dir_strict()` for the agents directory check and `is_file_strict()` for agent candidates and the conventional MCP file, inside the error handlers that already exist for this. Add a regression test that injects `EACCES` at `os.stat` and checks for `PluginManifestError`, not an empty result.

hertznsk:
Thanks for the careful read — addressed in dfe0c1d.

**What changed:**

- `plugins/agents.py::read_plugin_agents()` — `agents_dir.is_dir()` → `is_dir_strict()`, and the per-candidate `entry.is_file()` → `is_file_strict()`. Both sit inside the pre-existing `except OSError` handlers, so an unreadable `agents/` or an unreadable candidate (e.g. a symlink into an unreadable directory) now surfaces as `PluginManifestError("... could not be read: ...")` on every interpreter instead of silently resolving as "no agents".
- `plugins/manifest.py::_load_mcp_servers()` — the conventional `.mcp.json` fallback probe `fallback.is_file()` → `is_file_strict()`. As you noted, this was the one MCP path where the early return skipped the `read_text()` that used to preserve the diagnostic; the string-path form was already safe.
- Three regression tests in `tests/test_filesystem.py::TestCallSiteWiring` inject `EACCES` via `Path.stat` (matching the PR's existing wiring tests) and assert `PluginManifestError` rather than an empty result. Mutation-checked on Python 3.14.7: all three fail with `DID NOT RAISE` against the unfixed code and pass with the fix.

**Verification:** `tests/test_filesystem.py` 26/26, `test_plugins` + `test_skills` + `test_filesystem` 716/716, full `make test` 9278 passed / 35 skipped, ruff + ty clean on the changed files.

**Remaining probes audit:** I re-checked the call sites the finding didn't name. `find_manifest`'s `except OSError: continue` and `_has_agent_definitions`/`_dropped_components` are deliberately tolerant discovery/reporting paths with identical fall-through semantics on both interpreter generations, so they stay as-is. One borderline case is `plugins/resolution.py::_subdirectory()` for `plugin_sources.<name>.path`: pre-3.14 a raw `PermissionError` escaped there (no "could not be read" contract ever existed), while 3.14 misreports it as "does not exist" inside `PluginSourceError` — same divergence class, though resolution still fails loudly rather than dropping a capability. Happy to fold that one in here too if you'd like; otherwise I'd leave it for a small follow-up to keep this PR scoped to the reported failures.

jrob5756 (approved):

PR #552, opened 2026-09-22
fix(gates): keep a dialog open when the agent still has a question; add dialog.conversation_prompt

throup:
## What

An agent following the dialog system prompt routinely appends `[READY_TO_CONTINUE]` to a message like "Two tickets it is. Ready to move on to the next question when you are.", and `_extract_ready_marker` honoured it. The operator then saw a bare `Continue? (yes/no)` directly under a message that had invited an answer, and a conversational "yes" ended the dialog. In a real interview the agent announced three questions, asked one, and the dialog closed on the operator's "yes" with the other two unasked.

- **A trailing marker is withheld when the message reads as asking or announcing a question** (`_asks_or_announces_question`: a `?` outside a URL anywhere in the message, or a clause matching a short list of announcement patterns — `next question`, `question 2 of 3`, `let me ask`, `one more question`, `further questions` — in a clause carrying no negation, so "no further questions" passes and "No problem, next question is…" does not). The marker is stripped from the transcript either way, and the prompt now tells the agent not to combine the marker with a question.
- **The continue proposal names the agent and says what each reply does**: `Let <agent> continue? (yes ends the dialog; anything else is sent to <agent>)`. `yes`/`y` or EOF lets the agent continue, a dismiss keyword dismisses, an empty reply re-asks, and anything else is sent to the agent as its own turn. The web path's proposal message says the same.
- **The opening banner lists every dismiss keyword**, generated from `DISMISS_KEYWORDS`, since `continue` and `proceed` are natural mid-conversation words.
- **`dialog_completed` carries `agent_question_outstanding`** (also on `DialogResult`) when the dialog closed while the agent's last message was still asking or announcing a question; the terminal prints a line saying so.
- **The continue prompt is read on a daemon thread**, as the main turn already is: Ctrl-C there previously cancelled the task while the worker stayed blocked in `input()`, and the loop's shutdown then hung joining it.
- **New optional `dialog.conversation_prompt`**: `trigger_prompt` reaches only the evaluator that decides whether to *open* the dialog (`engine/dialog_evaluator.py`), so a workflow author had no field that reached the agent holding the conversation. The new field is appended to the built-in dialog system prompt through one shared builder for the terminal and web paths, stated to take precedence over the built-in rules where they conflict.

Two changelog fragments, one per user-facing change.

## What to scrutinise

1. **The predicate is a heuristic**, deliberately biased towards keeping the dialog open (the user can always send a dismiss keyword; a dialog ended early cannot be reopened). Known false positives that hold a closing message open: "I'll ask the team to review the draft", and an active-voice closing such as "I have addressed both follow-up questions." (a settled participle negates only after a copula or "all", so the passive "both follow-up questions are addressed" closes). The pattern list, the clause-scoped negation and the whole-message scan (a question followed by a recommendation and "Let me know." is still a question) are pinned by `TestAsksOrAnnouncesQuestion`. The alternative is to drop the guard and rely on the new prompt sentence alone; the guard exists because the prompt instruction is the one the model was already ignoring.
2. **Terminal-path behaviour change on a non-approval reply.** Previously the reply was recorded and the loop read *another* input, so the provider saw two consecutive user turns; now the reply is the next turn, matching the web path. `test_agent_proposes_continue_user_declines` is amended to script the agent's reply to that turn.
3. **`""` no longer counts as approval** on either path. On the terminal it re-asks; on the web an empty message at the proposal is ignored, a `null` `content` is read as `""` at every wait site rather than raising (an empty ordinary turn is still sent to the agent, as before), and the dashboard's leave-dialog control at the proposal now records `user_dismissed`, as it does everywhere else on the web path.

## Verification

- `tests/test_gates/test_dialog.py`: 134 passed (39 existing, unchanged apart from the amendment above; 95 new). Mutation: removing the guard branch in `_extract_ready_marker` fails exactly the six guard tests; deleting the transcript append for a reply at the proposal, the negation scoping, or the empty-reply loop each fail at least one.
- `make check`: ruff clean; `ty` reports the same 6 diagnostics as `main`.
- `make test` equivalent (`pytest -m "not install_scripts and not performance"`): 9577 passed, 43 skipped, 1 failed — `tests/test_skills/test_path_entries.py::TestSkillsRootDiagnostics::test_mis_cased_skill_md_is_reported`, a case-sensitive-filesystem test that does not hold on macOS APFS and touches nothing in this diff. `tests/test_interrupt/test_cli_terminal_pty.py::test_run_restores_after_provider_teardown[*]` was excluded from that run: its two cases fail identically on unmodified `main` in this environment.
- Real pty, scripted provider, no LLM: a withheld marker leaves the dialog open and the following "yes" reaches the agent as a turn; Enter at the proposal re-asks; "no, one more thing" is sent to the agent; Ctrl-D lets the agent continue; `done` under a question dismisses and prints the outstanding-question line; Ctrl-C at the proposal exits, where it hung before.
- Frontend: type-only addition to `DialogCompletedData`; `static/` is unchanged and needs no rebuild.


jrob5756 (commented):
Dialog fix is solid overall, but the URL-stripping regex has a real hole that reproduces the bug this PR is meant to close, so it shouldn't merge as-is. There's also a parity gap on the new diagnostic flag worth fixing before it lands.

Blocking:
- `src/conductor/gates/dialog.py:166` — the URL regex strips a Markdown link's closing delimiter along with the URL, so a question mark right after a link gets swallowed and the message is misread as a completion proposal.

src/conductor/gates/dialog.py
```diff
@@ -119,19 +135,116 @@ def _dismiss_instruction() -> Text:
...
+)
+# A clause that negates the announcement, or talks about questions already
+# dealt with ("the remaining questions were answered above").
+_NEGATION_RE = re.compile(
+    r"\b(?:no|not|nothing|never|without|answered|resolved|addressed|settled)\b|n't\b",
+    re.IGNORECASE,
+)
+_SENTENCE_BOUNDARY_RE = re.compile(r"(?<=[.!?])\s+")
+_CLAUSE_BOUNDARY_RE = re.compile(r"[,;:\u2014\u2013]")
+_URL_RE = re.compile(r"https?://\S+")
```
jrob5756: **BLOCKING**: This regex strips more than the URL — it also eats the Markdown closing delimiter and whatever punctuation follows it. So `Should we use [this repository](https://example.com/repo)? [READY_TO_CONTINUE]` gets read as a completion proposal instead of a question. Answering `yes` on either the terminal or web path then closes the dialog instead of forwarding the answer to the agent, which is exactly the premature-completion bug this PR sets out to fix. `agent_question_outstanding` comes out false in this case too.

Exclude Markdown link and autolink closing delimiters from the URL removal so trailing punctuation stays visible to the question detector. Add regression cases for `[label](https://example.com)?` and `<https://example.com>?`, and keep the existing coverage for question marks inside query strings.

src/conductor/gates/dialog.py
```diff
@@ -623,6 +773,7 @@ async def _web_handle_dialog(
                 result.user_dismissed = True
                 break
 
+        result.agent_question_outstanding = result.last_agent_message_asks_question()
```
jrob5756, line 789: **RECOMMENDED**: `agent_question_outstanding` is only computed after the conversation loop runs, but the web handler can return before that when the first user message is a dismiss keyword. So an opening question like `Which repository should I use?` followed immediately by `done` reports `agent_question_outstanding=False` and the completion event omits the field entirely — the same interaction on the terminal path correctly reports `true`.

Route first-message dismissal through the shared completion logic, or compute and include the flag before that early return. A terminal/web parity regression for dismissal right after the opening question would catch this going forward.

src/conductor/gates/dialog.py
throup: Fixed, thanks. A URL now ends before a Markdown link target's `)` or an autolink's `>`, and never on a `?`, so both of your cases read as questions and `_extract_ready_marker` withholds the marker; a `?` inside a query string (bare, in a link target, or in an autolink) is still removed with the URL. Regression cases cover both shapes, the query-string negatives, a link target with balanced parentheses (`Foo_(bar)?x=1`), and a bolded link question, which is what the delimiter exclusion alone protects.

A further pass turned up the same shape one construct over: a `?` after a bare URL (`Have you seen https://example.com?`) was also swallowed, so the pattern never ends a URL on `?`. I then applied RFC 3986's excluded set, so a `?` before a closing `"` or backtick is the message's too. Remaining residual: a `?` followed by another non-space character that is not `)`, `>`, `"`, a backtick or whitespace stays inside the URL. `'` and `*` are legal URL characters, so `Did you mean 'https://example.com?'` and `**Have you seen https://example.com?**` are still not read as questions.

src/conductor/gates/dialog.py
throup: Fixed, taking the first of your two remedies: the first-message dismiss branch no longer has its own `dialog_completed` emit and early return. It sets `user_dismissed` and skips the loop, so it reaches the same completion block as every other exit and the event carries the flag (and `agent_proposed_continue`) like the rest. A parity test runs `Which repository should I use?` followed by `done` through both paths and asserts `true` on the result and the event with no agent turn; a second case checks `false` for a closing opening message on the web path. The docs now say the line is printed in terminal mode and that the dashboard does not yet surface the flag.

throup:
Thanks for the review. Since 7142e96, six commits. The two findings: `_URL_RE` no longer removes a link's `)` or an autolink's `>`, never ends a URL on `?`, and treats `"` and backtick as non-URL characters per RFC 3986 (I also found the bare-URL case, `Have you seen https://x.com?`); and the web path's first-message dismissal reaches the shared completion, so `agent_question_outstanding` is reported there as on the terminal.

Beyond those, I also found and fixed false negatives in the announcement guard's negation rule (`one more question I need answered`, `haven't asked the next question yet`, a bullet's `No …` reaching the next line): a line break is now a sentence boundary, the settled participles count only after a copula or `all`, and a negated participle reads as still open. The trade-off is that an active-voice closing such as `I have addressed both follow-up questions.` now holds the dialog open, the heuristic's stated bias. The engagement prompt reads on the daemon thread like the other prompts, the changelog states the web path's dismissal parity at the continue proposal, the docs qualify the terminal-only line, and the three lines Codecov flagged are covered.

I've updated the description's test counts (134 passed, six guard tests) and its known-false-positive sentence to name the active-voice case. Ready for another look.

jrob5756 (approved):

PR #553, opened 2026-09-22
fix(cli): suppress BlockingIOError from console writes on non-blocking stderr

### LINKED ISSUE #543: BlockingIOError on large agent prompt logging in non-interactive / piped environments
### Describe the bug

When running a workflow in a non-interactive environment (such as `cron` or when `stderr`/`stdout` is connected to a pipe), `conductor run` fails with `BlockingIOError: [Errno 11] write could not complete without blocking` if an agent's rendered prompt is large (e.g. > 64 KB).

### Context & Reproduction Details

- **Workflow scenario:** A multi-agent workflow where parallel collector agents gather data and pass aggregated outputs to a downstream synthesis agent.
- **Prompt size:** The rendered prompt for the synthesis agent was 58,161 characters (~105 KB UTF-8).
- **Execution environment:** Running via `cron` (`0 7 * * * /bin/bash -lc "... conductor run workflow.yaml --no-interactive"`).
- In cron/piped environments, standard streams are pipes without a TTY. When writing large continuous blocks exceeding the OS pipe buffer capacity (typically 64 KB on Linux) to a pipe descriptor in non-blocking mode, `sys.stderr.write()` / `self.file.write()` raises `BlockingIOError`.

### Observed Behavior & Stack Trace

1. Orchestrator completes collector agents, routes to synthesis, prepares the agent prompt, and attempts to log it via `verbose_log_section`.
2. Logging to `_verbose_console` / Rich `Console` fails on `self.file.write(text)`:
```text
  File ".../site-packages/conductor/executor/agent.py", line 490, in execute
    _verbose_log_section(
        f"Prompt for '{agent.name}'",
        rendered_prompt,
    )
  File ".../site-packages/conductor/executor/agent.py", line 40, in _verbose_log_section
    verbose_log_section(title, content)
  File ".../site-packages/conductor/cli/run.py", line 401, in verbose_log_section
    _verbose_console.print(
        Panel(
            Text(content),
            title=styled("[cyan]{}[/cyan]", title),
            border_style="dim",
        )
    )
  ...
  File ".../site-packages/rich/console.py", line 2124, in _write_buffer
    self.file.write(text)
BlockingIOError: [Errno 11] write could not complete without blocking
```
3. The engine catches the exception and attempts error recovery / saving checkpoint, but the teardown handlers (`_print_resume_instructions` and `print_error`) also invoke console printing, triggering repeated `BlockingIOError` exceptions.

### Suspected Root Cause (Presumed)

- `verbose_log_section` unconditionally prints the full rendered prompt Panel to console when in full verbose mode (which is default).
- Diagnostic / verbose logging to standard output/error streams does not handle `BlockingIOError` / non-blocking pipe semantics.
- A failure during verbose logging aborts the entire workflow execution rather than being handled gracefully or suppressed.

### Expected Behavior

- Verbose console logging should not crash the workflow execution if `write()` cannot complete immediately or if the pipe buffer is full.
- Error recovery handlers (`_print_resume_instructions`, `print_error`) should survive stream write errors without crashing.

AmirF194:
Wrapped the two places that write your prompt panel to stderr, `_SilentAwareConsole.print` (covers `verbose_log_section`, `verbose_log`, and `_print_resume_instructions`) and `print_error`'s panel print, in `contextlib.suppress(BlockingIOError)`. Both of the teardown handlers you named already route through `_verbose_console.print`, so they get the fix without a separate change.

Repro was a plain `TextIOWrapper`/`BufferedWriter` write to a real `O_NONBLOCK` pipe with a stuck reader: it raises the identical `BlockingIOError(11, ...)`, no full cron/pipe setup needed to hit it.

Only `BlockingIOError` is swallowed. An unrelated write failure, say the disk fills, still surfaces, in both the print path and in `print_error`.

Fixes #543


src/conductor/cli/run.py
```diff
@@ -80,7 +80,11 @@ def print(self, *args: Any, **kwargs: Any) -> None:
         from conductor.cli.app import is_verbose
 
         if is_verbose():
-            super().print(*args, **kwargs)
+            # stderr can be a non-blocking pipe with a slow/absent reader
+            # (same root cause as copilot.py's _fix_pipe_blocking_mode); a
+            # large panel then raises here instead of blocking (#543).
+            with contextlib.suppress(BlockingIOError):
```
jrob5756, line 102: **BLOCKING**: Rich only clears its internal buffer once write() and flush() both succeed. Suppressing `BlockingIOError` outside `super().print()` leaves the whole rendered panel queued, prefix and all. Every subsequent print appends to that same buffer and retries it from the top, so once a panel outgrows the pipe's capacity, draining the pipe between calls just replays old prefixes instead of surfacing newer messages. Left running, the backlog grows without a bound, which is the opposite of what this fix is supposed to protect against.

Handle backpressure at the stream write/flush boundary so Rich can actually finish clearing its buffer, or explicitly discard the pending segments under its lock after the `BlockingIOError`. `print_error` needs the same treatment. Also worth a test against a real non-blocking pipe that forces repeated failures and then drains it, asserting the pending output stays bounded and later messages actually get through.

jrob5756 (commented):
One blocking issue here, in the BlockingIOError handling in `run.py`. As written, it suppresses the error but doesn't account for Rich's internal buffering, which can let output accumulate unbounded instead of actually recovering.

**Blocking:**
- `src/conductor/cli/run.py:86` — discards a failed write but leaves Rich's buffer (and the growing backlog behind it) in place

src/conductor/cli/run.py
AmirF194: You're right, and I checked it: catching BlockingIOError without clearing Rich's buffer meant a still-full pipe would replay old panel content ahead of each new print, growing without bound.

Pushed a fix: clear O_NONBLOCK on stderr's fd at console construction, mirroring what copilot.py already does for the Copilot CLI's own subprocess pipes. Writes block for the reader to drain instead of failing, so nothing needs to be discarded. The suppress calls stay as a fallback in case the flag gets reset on us mid-run.

jrob5756 (approved):

AmirF194:
Thanks for the review on the buffering issue, clearing O_NONBLOCK at construction is a cleaner fix than what I had.

PR #554, opened 2026-09-22
feat(claude-agent-sdk): secure native tools by default

ldavidgomez:
## Problem

On the experimental `claude-agent-sdk` provider, an agent that simply **omitted `tools:`** received the full Claude Code preset — filesystem read/write, shell, web and editing — with permissions bypassed (`bypassPermissions`). Every agent on this provider was therefore a filesystem/shell/web agent without the workflow ever saying so, and `working_dir` gave the impression of a boundary it does not provide.

## Change

A new provider setting, `runtime.provider.native_tools`, makes that grant explicit.

```yaml
runtime:
  provider:
    name: claude-agent-sdk
    native_tools: claude_code   # or: none (the default)
```

- **`none` — the new default.** Applies to an omitted field, YAML `null`/`~`, and the bare `provider: claude-agent-sdk` shorthand, which keeps serializing as the bare string.
- **`claude_code` — the explicit compatibility opt-in** to the full preset. It forces the structured serialization form.
- The field is rejected on every other provider. The factory forwards it with a `none` fallback (so a bare `--provider` override or a per-agent override also gets `none`), and the provider constructor fails closed on any unknown value.

### `tools:` behavior

| Agent `tools:` | `native_tools` | Built-in tools | Permission mode |
|---|---|---|---|
| omitted | `none` | none — `Skill` only when the agent has skills | `dontAsk` |
| omitted | `claude_code` | full `claude_code` preset | `bypassPermissions` |
| `[]` | either | none — `Skill` only when the agent has skills | `dontAsk` |
| non-empty | either | refused, as before (workflow tool names are not Claude tool IDs) | — |

`workflow_tools_passthrough` stays `False`; this is not an allowlist and there is no tool-name translation. An explicit `tools: []` always wins over the workflow-level opt-in. `dontAsk` is used rather than leaving the mode unset, because the CLI's `default` mode prompts, and a workflow run cannot answer a prompt. `dontAsk` is in the SDK's `PermissionMode` since 0.1.51, below the `>=0.2.82` floor.

### MCP-only agents under `none`

Declared MCP servers stay usable with no built-in tools at all:

- Once the workflow and plugin server maps are merged, each server is pre-approved by exactly one server-scoped rule, `mcp__<server>__*`. No `*`, `mcp__*` or other unanchored rule is ever generated.
- The SDK appends its own `Skill` grants to `allowed_tools`, so those are preserved.
- `strict_mcp_config=True`, the private `0600` config file, and its cleanup order are unchanged.
- **Server names are validated, not escaped.** Neither the schema nor the MCP translation constrains them, and a name like `a__*,Bash` would otherwise split into extra rules. `a__b` would also be read as server `a`. So names must match a strict shared pattern: letters, digits, `.`, `-` and single `_`, with no leading or trailing `_`. Anything else is refused at `conductor validate` and at run time, before the config file is written.
- Known limitation, documented rather than worked around: an MCP tool that requires interactive approval may be denied under `dontAsk`. There is deliberately no fallback to `bypassPermissions`.

### Why explicit `tools: []` plus MCP is rejected

`tools: []` means *no tools*. The SDK has no per-agent MCP switch, so without a guard any workflow or plugin MCP servers would still attach, and under `dontAsk` they would even be pre-approved. This combination was already rejected by `conductor validate` for workflow servers, but `conductor run` never calls the validator.

It is now refused at run time too, against the merged workflow + plugin map, under either `native_tools` value, before the MCP config is written or the SDK is invoked. The static check now also covers plugin-supplied servers. For an MCP-only agent, omit `tools:`.

### Plugin subagents

With no built-in tools there is no dispatch tool, and the SDK exposes no verified identifier to grant back. So a plugin that ships subagents is refused under `none` or `tools: []`, at validate and before the run starts, rather than registered unreachable. Subagents never enable the preset implicitly.

The remedy is accurate for this provider. `agents: false` on its own is refused while that plugin's skills are enabled, because registering the plugin root for its skills exposes its subagents again. The message therefore names `skills: false` as well where it applies; the validator names the exact switch per plugin.

### `settings_dir`

`add_dirs` widens only the built-in file tools, so it grants something only when the session has them. The existing "no `project` tier" warning used to always claim the directory was "still granted to the model's built-in file tools". It now describes the effective policy, at validate and at run time:

- under `claude_code` the grant remains;
- under `none` or `tools: []` the directory has no effect.

The warning's deduplication key now includes the tool policy, so each line stays accurate.

### Visible warning for the opt-in

`native_tools: claude_code` adds one fixed line to the existing experimental-provider banner. It states that filesystem, shell, web and editing tools are enabled with automatic approval and that `working_dir` is not a sandbox. It is driven by the provider metadata in `workflow_started`, so it shares the banner's existing once-per-run guard (a resume's second `workflow_started` does not reprint it) and its `--silent` suppression. The text is a fixed literal with no runtime values.

## Compatibility and migration

**This is a breaking change for the experimental `claude-agent-sdk` provider.** Workflows that relied on omitting `tools:` to read files, run commands or edit code must opt in:

```yaml
runtime:
  provider:
    name: claude-agent-sdk
    native_tools: claude_code
```

Without the opt-in such workflows still validate, but their agents run with no built-in tools. Configurations and checkpoints written before this field existed resolve to `none`. Authentication behavior (`auth_mode`, readiness probing) is untouched.

Examples: `repo-qa`, `session-key` and `setting-sources` opt in because they read or change files. In `setting-sources`, the prompt-only `audit` agent opts back out with `tools: []`. The MCP-only, Skill-only and prompt-only examples stay on `none`. A test now pins the expected policy for every agent in every shipped claude-agent-sdk example.

One changelog fragment is added (`changelog.d/+claude-agent-sdk-native-tools.changed.md`); `CHANGELOG.md` is not edited.

## Local validation

Run on the exact committed state, using an existing environment (`claude-agent-sdk` 0.2.87, Python 3.14) and without installing anything:

| Check | Result |
|---|---|
| Focused suite for this change (9 test files) | **553 passed, 1 skipped** (the skip is an unrelated missing example file in a parametrized sweep) |
| `ruff check --no-cache src tests` | **passed** |
| `ruff format --check --no-cache src tests` | **passed** (514 files) |
| `conductor validate` over all `examples/*.yaml` | **52 / 52 valid** |

**Not claimed:** I did not get a clean local run of the full test suite or of `ty check`. The local environment lacks `psutil`, which causes a `test_mcp_serve.py` collection error, four `TestGatherMcpServe` failures, and one `ty` unresolved-import diagnostic. There is also one macOS case-sensitivity failure, `test_mis_cased_skill_md_is_reported`. I reproduced every one of these identically against the unmodified base, and `ty` reports the same diagnostic set there. CI is the authoritative cross-platform result.


src/conductor/providers/claude_agent_sdk.py
```diff
@@ -1884,8 +1900,36 @@ async def _execute_session(
                 refuse_mcp_server_clashes(translated, session_servers)
                 session_servers.update(translated)
             if session_servers:
+                # Every refusal below happens BEFORE the secrets file exists
+                # and before the SDK is invoked: a configuration that is going
+                # to be refused must not first write resolved credentials to
+                # disk, even though the finally would reclaim them.
+                self._refuse_explicit_no_tools_with_mcp(agent, session_servers)
```
jrob5756, line 1922: **BLOCKING**: This guard also rejects OutputValidator's synthetic grading call. When a workflow declares `runtime.mcp_servers` and an agent has a `validator:` block, OutputValidator builds a synthetic agent with `tools=[]` and runs it through this same provider — but the provider still carries the workflow's MCP servers, so the refusal fires. OutputValidator catches the exception and reports `passed=True, errored=True`, so grading and retries quietly stop happening. This hits both `native_tools` modes, even when the primary agent correctly omits `tools`.

Give the grading call an MCP-free session rather than loosening the refusal for authored agents — weakening it would reopen the hole this PR is closing. Also add a regression through `OutputValidator` that proves grading still runs when the workflow has MCP servers configured; right now nothing catches this.

src/conductor/providers/claude_agent_sdk.py
```diff
@@ -2247,13 +2291,192 @@ def _session_transcript_exists(session_id: str, cwd: str) -> bool:
...
+                    f"the workflow never declared.",
+                    suggestion=(
+                        "Rename the server using only letters, digits, '.', '-' and "
+                        "single '_' characters, not starting or ending with '_'. It "
+                        "is declared in 'runtime.mcp_servers' or by a plugin that "
+                        "ships it."
+                    ),
+                    is_retryable=False,
+                )
+            rules.append(f"mcp__{name}__*")
```
jrob5756, line 2486: **BLOCKING**: `MCP_PERMISSION_SAFE_NAME` allows dots, so a name like `fs.tools-2` generates the permission rule `mcp__fs.tools-2__*`. Claude's CLI normalizes the server component to `fs_tools-2` in the actual tool name but compares permission rules literally, so `mcp__fs_tools-2__lookup` never matches that rule. Under the new default `dontAsk` mode, tools from an otherwise valid, documented server config get denied.

Drop `.` from the shared permission-safe name pattern and update the validation messages and docs to require CLI-stable names. Add a test that rejects dotted names outright, rather than one asserting an unnormalized rule is valid.

jrob5756 (commented):
This one isn't ready. Two of the blocking findings undercut the feature's own goal of securing native tools by default, and CI is red on the reviewed commit.

Blocking:
- `src/conductor/providers/claude_agent_sdk.py:1907` — the new MCP refusal also catches OutputValidator's synthetic grading calls, so validator blocks silently stop grading on any workflow with `mcp_servers` configured.
- `src/conductor/providers/claude_agent_sdk.py:2469` — dotted MCP server names produce permission rules the CLI's own normalization can never match, so valid server configs get denied under the new default-deny mode.
- CI run [35718818294](https://github.com/microsoft/conductor/actions/runs/35718818294), attempt 2, is `completed/failure`. The `Test (Python 3.13, windows-latest)` job failed: https://github.com/microsoft/conductor/actions/runs/35718818294/job/106851594182
- `CI / Test (Python 3.13, windows-latest)` check is reporting `FAILURE` on the same run/job as above.

Get the Windows test job green on this commit before asking for another look.

src/conductor/providers/claude_agent_sdk.py
ldavidgomez: You're right: this was causing validator failures to be treated as passes. I fixed it in 4d9b6aa without weakening the existing `tools: []` check.

`OutputValidator` now requests an MCP-free execution. Workflow and plugin MCP servers are removed before the refusal check, config write, or SDK call. Regular workflow agents cannot set this option, so an authored agent that combines `tools: []` with MCP servers is still rejected.

The regression test uses the real `OutputValidator` and verifies that the result is parsed instead of falling through the fail-open path. It also confirms that the grading session receives neither MCP servers nor built-in tools. The current head is 8be7db5; its workflows are still waiting for approval, so I don't have a CI result for it yet.

src/conductor/providers/claude_agent_sdk.py
ldavidgomez: I confirmed this against the CLI bundled with claude-agent-sdk 0.2.87. It replaces dots with underscores when building MCP tool names, so a permission rule containing the original dotted name will not match. This is fixed in 4d9b6aa.

Dotted server names are now rejected by both `conductor validate` and the runtime check, before the config file is written or the SDK starts. We reject them instead of silently renaming them; hyphens and single underscores remain valid.

The tests now cover dotted workflow and plugin server names as invalid cases while retaining positive coverage for valid names. The error messages and documentation also explain the normalization mismatch.

jrob5756 (approved):

PR #556, opened 2026-09-22
ci: run pytest suite in parallel

hertznsk:
## Why this matters

Conductor is developed extensively with AI coding agents, so test latency has a direct cost beyond ordinary developer waiting time. A serial full-suite run took about nine minutes while using only one CPU core. During that idle period, an agent cannot continue validating or iterating, and long pauses can allow the LLM provider's prompt/KV cache to expire. The next turn may then need to process the repository context again, increasing both turnaround time and inference cost.

Reducing the feedback loop to roughly two and a half minutes keeps agent sessions active, makes iterative validation practical, and lowers the chance of paying repeatedly to rebuild the same model context. The same improvement also benefits human contributors and CI capacity.

## Summary

- add `pytest-xdist` to the development toolchain
- run the normal local test suite with four bounded workers and CI/release suites with two workers using `--dist loadfile`
- keep real API, install-script, and performance tests outside the parallel default path
- propagate per-test temporary directories to child processes and document serial debugging

## Performance

On the same Linux development host, the default suite improved from **546 seconds** serially to **142 seconds** with four workers, a **74% wall-clock reduction**.

## Validation

- `make test`: 9550 passed, 20 skipped in 140.55s
- `make test-cov PYTEST_WORKERS=2`: 9550 passed, 20 skipped, combined 92% coverage
- risk-focused xdist run across background processes, web, Fleet TUI, and MCP launch integration: 1171 passed, 1 skipped
- marker/isolation regression set: 26 passed
- `make check`: passed; ty reported six pre-existing unused-ignore warnings
- Oracle review verdict: SHIP

## Notes

Install-script integration tests remain explicitly serial with `-n 0`. Set `PYTEST_WORKERS=0` for a serial local run.


jrob5756 (approved):

